PR #33, opened 2015-01-13
change cpu_linux.go: on some android, runtime.NumCPU() will not return r...

codeskyblue:
CPUCounts(true) use runtime.NumCPU() to get cpu numbers, But I tested this will get wrong number some times.


shirou:
Thanks! There are very different type of linux. If you notice something, please PR or make an issue anytime!

PR #34, opened 2015-02-07
+ process struct for freebsd 386

rekby:


shirou:
Thanks for the commit to the 386 ! cgo is not used for FreeBSD, but I am working to use it. After cgo introduced to FreeBSD, I will change this file. (of course, PR is always well come ;-) )

PR #35, opened 2015-02-07
+ host and process for linux 386

rekby:


shirou:
Thank you! Writing manual changed line in source code is good idea. I will add these to other files.

PR #36, opened 2015-02-10
Fix panics with docker stats

jwilder:
Fixes two panic in CgroupMem and CgroupCPU that can occur if you pass an invalid container ID.   The stat file may not exist and the error returned from reading it was being ignored.  This fix returns the error and does not try to continue reading the contents of the file.


shirou:
Thank you! I forgot to error check, my bad.

Issue #42, opened 2015-03-18
// TODO: how to get _SC_CLK_TCK ?  (in cpu code)


vrecan:
I have an example on how to get this

https://github.com/vrecan/goiostat/blob/master/systemCall/systemCall.go

I am currently trying to get cpu stats for ide iowait system and user and noticed it's not doing this correctly  right now.

vrecan:
I needed it to correctly calculate service time with iostats

shirou:
Thank you for the issue and the PR. But gopsutil aims to built with "pure" golang, not cgo. This TODO means I don't know how to get _SC_CLK_TCK with system call or anything. If you know, please reopen this issue.

Thank you.

vrecan:
Just curious why do you care so much that it's using cgo? It builds easily on linux and it's in a file that will only build on linux. 

I don't know of a way to get this without c, which means the implementation will always be wrong on some platforms.

shirou:
Because if using cgo, it can not be cross compiled (at least it's not easy).  Since gopsutil is cross-platform library, cross-compiling is very important. And this is why I am choosing golang.

PR #43, opened 2015-03-18
syscall for ticks per second linux

vrecan:


vrecan:
#42 fixes

shirou:
Thanks but I reject because it's using cgo.

PR #47, opened 2015-06-15
implement CPUTimes (and CPUPercent) on darwin. use ps command.

hhatto:
Hi.

implemented to process.CPUTimes & process.CPUPercent on MacOSX.
using `process.process_darwin.callPs()`.

Thanks for creating gopsutil.

Regards.


shirou:
Thank you for your contribution! Looks great!

PR #48, opened 2015-06-16
implement NumThreads on darwin. use ps command

hhatto:


shirou:
Wow! Thank you so much!

hhatto:
thanks :-)

PR #49, opened 2015-06-18
Add SuSe platform detection

databus23:
As the commit messages says this adds platform detection for SuSe Linux


shirou:
Great contribution!
I don't have a SuSe Linux but I checked the [ohai](https://github.com/chef/ohai/blob/31f6415c853f3070b0399ac2eb09094eb81939d2/lib/ohai/plugins/linux/platform.rb#L84) and I think it should work.

Thank you!

PR #53, opened 2015-07-15
Fix interpretation error about /proc/stat

gwind:
ref: 

- https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L128

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -150,10 +150,6 @@ func parseStatLine(line string) (*CPUTimesStat, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


shirou:
Thank you! I can not remember why change the indexes.

gwind:
You are welcome, :-)

btw, there is have a confused attr in https://github.com/shirou/gopsutil/blob/master/cpu/cpu.go#L22 .

shirou:
Ah, `stolen` is came from [gosigar](https://github.com/cloudfoundry/gosigar/blob/master/sigar_linux.go#L339) or else. I can change to `Stolen` but it breaks compatibility. So I want to not change it.

PR #58, opened 2015-08-05
Allow gopsutil docker to build on non-linux boxes by stubbing out non…

sparrc:
…-linux libraries

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -1,0 +1,36 @@
+package docker
+
+import "errors"
+
+var ErrNotAvailable = errors.New("docker not available")
+
+type CgroupMemStat struct {
+	ContainerID             string `json:"container_id"`
+	Cache                   uint64 `json:"cache"`
+	RSS                     uint64 `json:"rss"`
+	RSSHuge                 uint64 `json:"rss_huge"`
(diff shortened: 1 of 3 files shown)


sparrc:
This should allow gopsutil/docker to build on mac systems, but will currently return a NotImplementedError if the functions are actually tried to use. Mac machines will still have access the the datatypes of the docker package though.

shirou:
docker package is separated from other packages. So I was thinking just not import docker package on on non-linux system.
This PR means you want to use docker package with linux AND non-linux system from one source code, and also want to use datatypes in the docker package.

I think your PR is quite reasonable and will merge with thankfulness, but just want a clarification.

Thank you.

sparrc:
@shirou, the reason for wanting this was that we would like to not have to write separate build files for the docker monitoring system on our end for linux & mac. Instead, we would like to be able to build and unit test the docker system on both OSes.

One of the main benefits of this PR is that it allows a mac system to access all the datatypes of the docker package for the purpose of unit testing, so that we can build on a mac despite the functions not being implemented.

shirou:
I understand. Thank you!

I am interested in docker monitoring system, but the docker package is not currently well tested on various environment. I will appreciate if you find and report a bug.

PR #60, opened 2015-08-26
Bug fix and HostStatInfo.Platform* windows implementation

mayowa:
Bug Fixes
On some windows computers cpu.CPUInfo() failed because, Win32_OperatingSystem.L2CacheSize isn't available in the wmi database
Some computers didn't return a value for Win32_OperatingSystem.ProccessorId so the type was changed to *string to accommodate this

Additions
On windows builds HostStatInfo.Platform* fields now return proper values


shirou:
Thank you for your PR! Awesome windows specialist!

However, I can not verify your code since my Windows can return ProcesserId. But even if we can get a ProcesserId,  your code seems take away it. Cloud you change your code to check that ProcesserId is nil or not? And if not nil, use that.

mayowa:
I'm not sure I know why you no longer get a value for ProcessorID, I just changed the type from string, to *string, which should only be nil, if wmi doesn't return a value, or did you mean to talk about the L2CacheSize property instead? But, I'll take another look at ProcessorID just in case.

cpu/cpu_windows.go
```diff
@@ -63,15 +62,17 @@ func CPUInfo() ([]CPUInfoStat, error) {
 	if err != nil {
 		return ret, err
 	}
-	for i, l := range dst {
+	
+	procID := ""
```
shirou: I think this makes procID always empty. sorry if I am wrong.

cbednarski:
@mayowa Thanks for working on this! It turns out I need some of the same features you added. :)

cpu/cpu_windows.go
```diff
@@ -15,12 +15,11 @@ import (
 
 type Win32_Processor struct {
 	LoadPercentage            uint16
-	L2CacheSize               uint32
 	Family                    uint16
 	Manufacturer              string
 	Name                      string
 	NumberOfLogicalProcessors uint32
-	ProcessorId               string
+	ProcessorId               *string
```
cbednarski: I think it's better to use `string` instead `*string`. The reason is that `*string` may be `nil` and trying to access this value will cause a program to crash. If this is type `string` you can simply test `== ""` to see if it has been specified, but accessing the string normally will always work because it is initialized with the default value.

cpu/cpu_windows.go
mayowa: Oh that explains it, It is left over code that i forgot to remove. Initially i tried to do what you suggested but it didn't work. Ok i'll have a second go at it,  @cbednarski once i fix this it should address your concerns too

mayowa:
@shirou done! 
@cbednarski the Win32_Processor struct is only used to query wmi which sets a nil if a value isn't available. hence the reason why ProcessorId has to be '*string'. The CPUInfoStat struct which is returned by cpu.CPUInfo has ProcessorId as 'string' so in CPUInfo() I convert from *string to string before setting the value of CPUInfoStat.ProcessorId

shirou:
Thank you fixinig. It looks work (I can not confirm now).

Also, could you tell me what value in `ProcessorId` if ProcessorId is `string`, not `*string` on your environment?
If it's just an empty, string is more safe than a pointer. I know you want to accommodate but avoiding nil pointer exception is much more important I think. (From your code, nil pointer exception can not be occured, but when I forget later and edited, it can be caused.)

mayowa:
When Win32_Processor.ProcessorId is 'string', I get the following error:
 cpu_test.go:82: error wmi: cannot load field "ProcessorId" into a "string": unsupported type (<nil>)

(I'm testing this from a Windows 7 VM)

shirou:
Thanks for the confirmation. So in this case, `*string` is required.  All point is cleared. I will merge this.

@cbednarski thank you for joining this discussion!

cbednarski:
Thanks, both of you, for working on this!

PR #61, opened 2015-08-26
Read Darwin CPU frequency from sysctl hw.cpufrequency

cbednarski:


shirou:
Thank you so much! great!

PR #64, opened 2015-08-27
Fix memory reporting for linux systems

sparrc:
/proc/meminfo reports memory in KiloBytes and so needs a multiplier of
1024 instead of 1000.
The kernel reports in terms of pages and the proc filesystem is left
shifting by 2 for 4KB pages to get KB. Since this is a binary shift,
Bytes will need to shift by 10 and so get multiplied by 1024.

From the kernel code. PAGE_SHIFT = 12 for 4KB pages
"MemTotal:       %8lu kB\n", K(i.totalram)

Thanks to @subhachandrachandra!

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -26,10 +26,9 @@ var (
 
 const WaitMSec = 500
 
-func DiskUsage(path string) (DiskUsageStat, error) {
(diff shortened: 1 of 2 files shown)


shirou:
wow! Thank you so much!

Issue #68, opened 2015-09-02
Doesn't compile on windows


vrecan:
```
C:\Users\ben.aldrich\go\src\github.com\shirou\gopsutil\disk [master]> go build
# github.com/shirou/gopsutil/disk
.\disk_windows.go:43: no new variables on left side of :=
.\disk_windows.go:43: cannot use DiskUsageStat literal (type *DiskUsageStat) as type DiskUsageStat in assignment
.\disk_windows.go:54: cannot use ret (type DiskUsageStat) as type *DiskUsageStat in return argument

```

vrecan:
I will have a PR up soon to fix this

vrecan:
Also noticed a decent amount of the tests fail in windows
```
=== RUN TestReadlines
--- PASS: TestReadlines (0.00s)
=== RUN TestReadLinesOffsetN
import (
--- PASS: TestReadLinesOffsetN (0.00s)
=== RUN TestIntToString
--- PASS: TestIntToString (0.00s)
=== RUN TestByteToString
--- PASS: TestByteToString (0.00s)
=== RUN TestStringsContains
--- PASS: TestStringsContains (0.00s)
=== RUN TestPathExists
--- PASS: TestPathExists (0.00s)
PASS
ok      github.com/shirou/gopsutil/common       0.024s
# github.com/shirou/gopsutil/process
process\process_test.go:246: p.CreateTime undefined (type Process has no field or method CreateTime)
=== RUN TestCpu_times
--- PASS: TestCpu_times (0.00s)
=== RUN TestCpu_counts
--- PASS: TestCpu_counts (0.00s)
=== RUN TestCPUTimeStat_String
--- PASS: TestCPUTimeStat_String (0.00s)
=== RUN TestCpuInfo
--- PASS: TestCpuInfo (1.06s)
=== RUN TestCPUPercent
--- FAIL: TestCPUPercent (3.07s)
        cpu_test.go:82: error wmi: cannot load field "LoadPercentage" into a "uint16": unsupported type (<nil>)
        cpu_test.go:82: error wmi: cannot load field "LoadPercentage" into a "uint16": unsupported type (<nil>)
=== RUN TestCPUPercentPerCpu
--- FAIL: TestCPUPercentPerCpu (3.07s)
        cpu_test.go:82: error wmi: cannot load field "LoadPercentage" into a "uint16": unsupported type (<nil>)
FAIL
FAIL    github.com/shirou/gopsutil/cpu  7.235s
[...]

PR #69, opened 2015-09-02
Fix windows compile issues

vrecan:
This should fix compile fail. 

fixes: #68


shirou:
Thank you!
I will add build test to CI.

PR #70, opened 2015-09-04
DiskPartitions Fails if cd not in drive And used and usedPercent are missing in windows

vrecan:
Tested this on windows 7. Percent used comes out correctly on my drives.


shirou:
Thank you for CD-ROM device fix and confirmation of used percentage. It also works on my Windows7.
I can't remember why the code was commented out...

PR #71, opened 2015-09-05
Fix cpu percent on windows

vrecan:
Tested on windows 7:

When divided by 100 you get cpu percentage of  %0.14 when task manager shows %14. Once this is removed they match.


shirou:
I checked other os and original psutil implementation, and you are correct.
Thank you!

PR #74, opened 2015-09-11
Ensure that process.MemoryInfo returns a up-to-date value.

choplin:
process.MemoryInfo in process_linux.go returns directly process.memInfo, which might be outdated. This patch calls fillFromStatm before returning p.memInfo so that p.memInfo becomes up-to-date.


shirou:
Thank you. I will look at other methods.

PR #80, opened 2015-09-18
use the Used memory explicitly rather than the calculated Available v…

sparrc:
…alue

Qualitatively this has been a bit more accurate in my experience

### DIFF
--- a/mem/mem_darwin.go
+++ b/mem/mem_darwin.go
@@ -61,7 +61,7 @@ func VirtualMemory() (*VirtualMemoryStat, error) {
 	ret.Available = ret.Free + ret.Buffers + ret.Cached
 
 	ret.Used = ret.Total - ret.Free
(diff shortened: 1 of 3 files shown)


sparrc:
part of the reasoning behind this is that I don't exactly understand the difference between "available" and "free" memory? Free memory is clearly documented in `man proc`, but available is not, can you explain what that is?

shirou:
`Available` meaning depends on platform. See https://pythonhosted.org/psutil/#psutil.virtual_memory . 
- https://github.com/giampaolo/psutil/blob/master/psutil/_psbsd.py#L76
- https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L192

And I noticed, gopsutil implementation on FreeBSD is still `TODO`. I will fix it very soon.

sparrc:
Okay, now I understand, I'm going to close this PR

PR #82, opened 2015-09-18
Use vm_stat command on darwin to get memory

sparrc:
This has the benefit of allowing us to also grab the cached memory so
that we can get an "available" measurement


shirou:
Thank you!
I have a plan to mocking using interface and expected text file (see common/Invoker). But not completed yet. So forgive me when I will change your test code.

PR #83, opened 2015-09-21
Fix docker on Centos 7

neezgee:
Fixes this error:

    error getting docker info: open /sys/fs/cgroup/cpuacct/docker/05a087a0d8b88c706493d28de9c911a62a513dae5dd19bb76a50b6c7ed39e5ab/cpuacct.stat: no such file or directory

On Centos 7 docker control groups are /sys/fs/cgroup/cpuacct/system.slice/docker-....

Please pull.


shirou:
I confirmed CentOS7 use different path. Thank you!

PR #84, opened 2015-09-22
When MemAvailable is in /proc/meminfo, use it (kernel 3.14+)

sparrc:
MemAvailable is now officially tracked in the linux kernel! They did this to try to fix some of the confusion around used/actual_used memory. In my experience, the number the kernel is reporting is a little bit lower than what you get when you just do `Free + Cache + Buffered`.

### DIFF
--- a/mem/mem_linux.go
+++ b/mem/mem_linux.go
@@ -13,6 +13,8 @@ import (
 func VirtualMemory() (*VirtualMemoryStat, error) {
 	filename := "/proc/meminfo"
 	lines, _ := common.ReadLines(filename)
+	// flag if MemAvailable is in /proc/meminfo (kernel 3.14+)
+	memavail := false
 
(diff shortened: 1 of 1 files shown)


mem/mem_linux.go
```diff
@@ -43,7 +48,9 @@ func VirtualMemory() (*VirtualMemoryStat, error) {
 			ret.Inactive = t * 1024
 		}
 	}
-	ret.Available = ret.Free + ret.Buffers + ret.Cached
+	if !memavail {
```
shirou: Could you change condition to opposite such as `if memavail` ? I think it makes easier to understand. After that, I will merge this. Thanks!

shirou: oops. I made a mistake. Sorry, forget about it.  will merge it. Thank you!

shirou:
Your PR makes more accurate. Thanks a lot!

PR #85, opened 2015-09-28
CreateTime() on linux returning an incorrect timestamp

andyjones:
CreateTime() on linux seems a few hours out - looks like it is converting the bootTime to microseconds before adding the elapsed time in seconds instead of adding them and then converting the total to microseconds


shirou:
Ah. Indeed. Thank you!

andyjones:
thanks

PR #89, opened 2015-10-10
cpu[darwin]: separetes cpu_darwin into cgo and nocgo to get CPUTimes.

shirou:
Separetes cpu_darwin into these files.

- cpu_darwin.go (common)
- cpu_darwin_cgo.go
- cpu_darwin_nocgo.go
  - it retunres `not implemented yet` error

I hope it will work on those environemnt.

1. darwin self compile without cgo
2. darwin self compile with cgo (default)
3. linux cross compile to darwin without cgo
4. linux cross compile to darwin with cgo

I confirmed 1,2, and 3.

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -2,27 +2,10 @@
 
 package cpu
 
-/*
-#include <stdlib.h>
-#include <sys/sysctl.h>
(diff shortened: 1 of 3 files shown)


dadgar:
LGTM

Issue #90, opened 2015-10-13
Make host/linux.go continue if os.Hostname fails.


client9:
Hi,  thanks so much for this project!  We love it.

in host/linux.go we have this function, but   `os.Hostname` failure causes it to fail completely.

In other places, failure just means "zero value". 

My proposal is to

* on os.Hostname failure, just use "" for host.
* always return "nil" for error

Ideally I would like to "return by value" and not a `*HostInfoStat` but that might break compatibility.

Im happy to send in a pull request.

with kind regards,

n

```go
func HostInfo() (*HostInfoStat, error) {
	hostname, err := os.Hostname()
	if err != nil {
		return nil, err /// CHANGE -->  hostname = "" and keep going
	}

	ret := &HostInfoStat{
		Hostname: hostname,
		OS:       runtime.GOOS,
	}

	platform, family, version, err := GetPlatformInformation()
	if err == nil {
		ret.Platform = platform
		ret.PlatformFamily = family
		ret.PlatformVersion = version
	}
	system, role, err := GetVirtualization()
	if err == nil {
		ret.VirtualizationSystem = system
		ret.VirtualizationRole = role
	}
	uptime, err := BootTime()
	if err == nil {
		ret.Uptime = uptime
	}

	return ret, nil
}
```

shirou:
Thank you for the proposal. I agree about continueing even if `os.Hostname` fails. 

Could you send me a PR for  linux? I will create same thing for other platforms. Thanks!

client9:
great will do!

n

shirou:
I really appreciate your contributing. Thank you very much.

PR #91, opened 2015-10-13
shirou/gopsutil#90 make hostinfo more robust

client9:
I cleaned up linux, freebsd, and darwin so they all work the same.

windows needs more work.

thanks,

nickg


shirou:
Thank you!

PR #97, opened 2015-10-16
chore(*): Fixes #94

### LINKED ISSUE #94: Using gopsutil in a container
So there is a default /proc file system within a container that is used for container level pids etc... It would be nice to override where gopsutil looks so you could bind mount the host /proc file system into the container.

jchauncey:
Added the ability to fetch an alternative location for /proc via an
environment variable. If the env var is not set it will return /proc as
the default value.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ gopsutil: psutil for golang
         :target: https://coveralls.io/r/shirou/gopsutil?branch=master
 
 
-This is a port of psutil (http://pythonhosted.org/psutil/). The challenge is porting all 
+This is a port of psutil (http://pythonhosted.org/psutil/). The challenge is porting all
 psutil functions on some architectures...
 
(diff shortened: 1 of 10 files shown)


process/types_darwin.go
```diff
@@ -50,7 +50,7 @@ package process
 
 #include <sys/sysctl.h>
 #include <sys/ucred.h>
-#include <sys/proc.h>
+#include <sys/proc).h>
```
shirou: Is this something typo?

jchauncey: Yup I'll fix it. Search and replace error

jchauncey:
changed dfault in comment to default (but kept argument the same since default is a keyword). Also fixed the typo in `types_darwin.go`.

shirou:
Thank you!

PR #98, opened 2015-10-17
cpu: support ARM CPUs.

### LINKED ISSUE #88: cpu.CpuInfo() return empty data on Raspberry Pi
cpu.CPUInfo() returns empty data on Raspberry Pi 2 (quadcore, ARM v7) running ubuntu 15.04 (kernel 3.17)  . This is due to /proc/cpuinfo being differently formatted. Following is an example:
```
processor : 0
model name  : ARMv7 Processor rev 5 (v7l)
BogoMIPS  : 57.60
Features  : half thumb fastmult vfp edsp neon vfpv3 tls vfpv4 idiva idivt vfpd32 lpae evtstrm
CPU implementer : 0x41
CPU architecture: 7
CPU variant : 0x0
CPU part  : 0xc07
CPU revision  : 5
```
CPU frequency cant be obtained from /proc/cpuinfo on ARM. There are two ways to get this data:
a) Invoke `lscpu`   and parse the resulting data
b) read /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq (one entry for each core)

I dont know whats the preferred way to implement this, @shirou if you can give some feedback, i'd be happy to implement this :-)
thanks for this awesome library

kardianos:
ARM CPUs don't include the same fields as  x86 and amd64 CPUs in
the /proc/cpuinfo list. Pull information from the /sys/...
device tree as well as updating when a CPU is done in cpuinfo.

Tested on an ARMv7 A20.

Fixes #88

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -3,6 +3,7 @@
 package cpu
 
 import (
+	"fmt"
 	"errors"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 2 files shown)


shirou:
Thank you for this great contribution. It seems redeeming from `/sys` is more general than lscpu.

@ranjib @cbednarski Could you confirm this PR is working at your ARM and intel environment?

cpu/cpu_linux.go
```diff
@@ -55,19 +56,36 @@ func CPUTimes(percpu bool) ([]CPUTimesStat, error) {
...
+		lines, err := common.ReadLines(fmt.Sprintf("/sys/devices/system/cpu/cpu%d/topology/core_id", c.CPU))
+		if err == nil {	
+			c.CoreID = lines[0]
+		}
+	}
+	return nil
+}
+
 func CPUInfo() ([]CPUInfoStat, error) {
 	filename := "/proc/cpuinfo"
```
shirou: Sorry, I merged another PR after your PR is sent. 
Could you change to use `common.GetEnv`  like https://github.com/shirou/gopsutil/blob/master/cpu/cpu_linux.go#L59 ? (See #97 )

shirou:
Thank you so much.

PR #102, opened 2015-10-19
Common internal

kardianos:
Put the "common" package behind the internal folder so it is not part of an API promise, implicit or explicit.


shirou:
Thanks!

Can the `internal` work on go1.4? I think it works, but I want just a confirmation.

kardianos:
It compiles just fine in go1.4, just like any other folder; in go1.4 it is only enforced for the std lib. For go1.5 it is enforced for all.

So yes, it will still compile and function for go1.4.

shirou:
Thank you!

PR #105, opened 2015-10-22
cpu: CPUInfo returns one item per physical thread. 

### LINKED ISSUE #104: cpu: CPUInfo needs to clarify or correct topology.
nomad currently uses cpu.CPUInfo to count cores. It takes the total CPUInfo struct count and adds each core count in each cpu to get the total number of cores. For example my six core machine will result in reporting 36 total cores when counted this way.

Probably Processor / Cores need better topology reporting on all architectures. For example under "/sys/.../cpuX/topology/ it can report which cores are in each processor.

kardianos:
Set "Cores" to 1 on Linux.

Fixes #104

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -80,13 +80,20 @@ func finishCPUInfo(c *CPUInfoStat) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


shirou:
Thank you so much!

PR #106, opened 2015-11-12
docker: fix static /sys paths by using env variables

cgeoffroy:
This PR removes some hard-coded `'/sys'` path in the docker lib. They are replaced by the value of the `HOST_SYS` environment variable.

This PR allows a container to monitor the host or another container without privileges and with a clear separation of the monitored directories and the container filesystem. To do so, simply
* Add `HOST_SYS=/rootfs/sys` to the container's environment
* mount the targeted directory as a volume in the container: `-v '/sys:/rootfs/sys:ro'`


dustinblackman:
:+1:

shirou:
Thanks a lot!

Issue #107, opened 2015-11-16
mem Usage about CgroupMemStat 


upccup:
hi @shirou  Very glad to see you add a metric related to docker. But there has a little question. Why do not you return docker memory like docker API 
```
"memory_stats" : {
        "stats" : {
           "total_pgmajfault" : 0,
           "cache" : 0,
           "mapped_file" : 0,
           "total_inactive_file" : 0,
           "pgpgout" : 414,
           "rss" : 6537216,
           "total_mapped_file" : 0,
           "writeback" : 0,
           "unevictable" : 0,
           "pgpgin" : 477,
           "total_unevictable" : 0,
           "pgmajfault" : 0,
           "total_rss" : 6537216,
           "total_rss_huge" : 6291456,
           "total_writeback" : 0,
           "total_inactive_anon" : 0,
           "rss_huge" : 6291456,
           "hierarchical_memory_limit" : 67108864,
           "total_pgfault" : 964,
           "total_active_file" : 0,
           "active_anon" : 6537216,
           "total_active_anon" : 6537216,
           "total_pgpgout" : 414,
           "total_cache" : 0,
           "inactive_anon" : 0,
           "active_file" : 0,
           "pgfault" : 964,
           "inactive_file" : 0,
           "total_pgpgin" : 477
        },
        "max_usage" : 6651904,
        "usage" : 6537216,
        "failcnt" : 0,
        "limit" : 67108864
     }
``` 
there has the max_usage and usage and limit clear.

shirou:
Even the package name is `docker`, it does not limited to Docker, can get such as systemd or rocket. I aim to implement more general way, so currently I do not have an plan to use Docker specific API. (yah,  currently. your comments are always welcome!)

upccup:
hi @shirou  Thanks for your reply.  But you may have misunderstood what I meant  I do not mean for you to call the docker's API. You can get docker memory usage in the file */sys/fs/cgroup/memory/docker/containerId/memory.limit_in_bytes*. I just mean in memory You can provide more info like * max_usage*, * usage*, * limit*, * failcnt*.

shirou:
Oh, sorry, I got it now. Thank you! Actually, I was lazy to read multiple files :P 

It seems plenty of files under `/sys/fs/cgroup/memory/docker/containerId/`

- memory.failcnt
- memory.limit_in_bytes
- memory.max_usage_in_bytes
- memory.usage_in_bytes
- memory.kmem.failcnt
[...]

PR #110, opened 2015-11-21
docker[linux]: add more memory stats in cgroup.

shirou:
refs. #107 

add these cgroup metrics to `CgroupMemStat`.

- memory.usage_in_bytes
- memory.max_usage_in_bytes
- memory.limit_in_bytes
- memory.failcnt


upccup:
:+1:

PR #114, opened 2015-11-23
Fix HostInfo.Uptime, which has been returning boot time

sparrc:
Also adding a HostInfo.BootTime field for holding the boot time


shirou:
Thank you!

PR #115, opened 2015-11-25
[freebsd] fix SwapMemory for hosts without swap

def:


shirou:
Thank you! Could you fix this error by add import?

```
# github.com/shirou/gopsutil/mem
./mem_freebsd.go:127: undefined: errors in errors.New
```

def:
done

shirou:
Thank you for the quick response and contributing!

PR #117, opened 2015-11-25
[host] fix build for freebsd 386

def:


shirou:
for my amd64 environment, your modification on the `host_freebsd_amd64.go` caused `unexpected EOF` to parse `Utmpx`. When I created `Users()` at first time, utmpx size is little different. So I commented out `Host` and `X__ut_spare`.
I have not dig this problem deeper. If you have some idea about solve this,  I very appreciate.

def:
I've not understood utmpx size calculation. It seems like a magic:) 
I can revert changes for amd64 and make similar logic for 386. Or we can use cgo for this feature, but it can cause build problems on pre 9.0 FreeBSD.

shirou:
Sorry for very late response. I merged #191, it might be work on FreeBSD 386. I close this PR. But if you have any problem, please open a issue.

PR #118, opened 2015-11-25
[disk] fix build for freebsd 386

def:


shirou:
Unfortunately, I don't have a 386 FreeBSD environment. I can not confirm your PR but it could be OK. So I merge it. Thank you!

PR #121, opened 2015-12-07
Handle process filenames with spaces in the name

### LINKED ISSUE #120: Doesn't handle processes with spaces in name
As per http://man7.org/linux/man-pages/man5/proc.5.html, the second field is the filename of the executable in parentheses:

> 2) comm  %s
                        The filename of the executable, in parentheses.
                        This is visible whether or not the executable is
                        swapped out.

In the case of spaces in the filename (as I observed today) `/proc/[pid]/stat` looks like:

    1591 (Plex Media Serv) S 1 1591 1591 0 -1 1077952768 19063 0 306 0 257 95 0 0 20 0 17 0 1353 596594688 13330 18446744073709551615 1 1 0 0 0 0 0 4097 17646 18446744073709551615 0 0 17 4 0 0 69 0 0 0 0 0 0 0 0 0 0

Rather than hard coding the field numbers they should be offset from the field ending in `)`.

jimmidyson:
Fixes #120

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -582,26 +582,31 @@ func (p *Process) fillFromStat() (string, int32, *cpu.CPUTimesStat, int64, int32
 	}
 	fields := strings.Fields(string(contents))
 
+	i := 1
+	for !strings.HasSuffix(fields[i], ")") {
+		i++
+	}
+
 	termmap, err := getTerminalMap()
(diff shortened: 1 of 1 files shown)


shirou:
Oh! yes, you are right. and I confirm your PR fixes it. Thank you very much!

PR #125, opened 2015-12-11
[net] fix panic on freebsd (netstat error with exit code 0)

def:
From freebsd jail netstat returns error message, but exit code is 0:
```
# netstat -ibdn && echo $?
netstat: kvm not available: /dev/mem: No such file or directory
ifnet: symbol not defined
0
```


shirou:
Thank you! (I found I have not implemented detecting virtualization on FreeBSD.)

PR #126, opened 2015-12-11
Add a godoc badge

sparrc:
Change coveralls badge to svg and put in a godoc badge.

drone.io should be supporting svg badges soon (see https://github.com/badges/shields/issues/150 and https://github.com/badges/shields/issues/294), which look much better, especially on 4K screens :-)


shirou:
Thank you!

PR #127, opened 2015-12-14
process[darwin] Fix unit of RSS and VMS

olivielpeau:
Should be bytes, `ps` returns Kbytes.


shirou:
Thank you. I confirmed RSS and VSZ are in Kbytes from `man ps` .

```
    rss       the real memory (resident set) size of the process (in 1024 byte units).
    vsz        virtual size in Kbytes (alias vsize)
```

PR #128, opened 2015-12-14
using HostProc for protocounters in linux

ontsysmattholcomb:
Using ENV variables for host /proc location when parsing protocol counters.


shirou:
Yes! Thank you very much!!

PR #129, opened 2015-12-14
Adding some identification to the 'Not Implemented Yet' error.  (Help…

lcaballero:
…s to identify which code base is producing this error.  By itself it's very generic.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 #*
 _obj
 *.tmp
+.idea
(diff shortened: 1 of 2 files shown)


shirou:
Sorry, I can not agree this PR. In the golang, error handling such as logging should be done on the invoker side, I think. If you need to identification, add some word to your test code.

For example, one of the largest golang project Docker does not embed the project name.
https://github.com/docker/docker/search?p=2&q=not+implemented&utf8=%E2%9C%93

PR #130, opened 2015-12-15
Add -W flag to netstat on BSD systems

sparrc:
This prevents netstat from truncating the interface name to 5
characters. See https://github.com/influxdb/telegraf/issues/438


sparrc:
from the netstat man page:

```
-W    In certain displays, avoid truncating addresses even if this causes some fields
           to overflow.
```

So maybe I need to test this more, I'm not sure what would happen if some fields overflow

shirou:
I did not know that option. Thank you. I think some overflow happend only on some displays, not stdout.

PR #131, opened 2015-12-15
CPU race condition fix

ryanjoneil:
Using `gopsutil/cpu` inside a multithreaded application triggers the go race detector. This removes the global variables that cause this and restructures the `CPUPercent` function for `cpu_unix`.


shirou:
You are right. Thank you so much!

PR #134, opened 2016-01-11
Get Arch Linux Version from LSB

pquerna:


shirou:
When I use Arch linux on EC2, `platform_version` is not set after applying your PR.
Arch linux does not have `/etc/lsb-release` or `/usr/bin/lsb_release`, only way to get platform_version is using `uname -r`. (FYI: https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb#L135)

I am not familiar with Arch linux, please let me know if I am wrong.

Thank you!

pquerna:
On the Rackspace Cloud Server I booted to test this, `/etc/lsb-release` is present:

```
LSB_VERSION=1.4
DISTRIB_ID=Arch
DISTRIB_RELEASE=rolling
DISTRIB_DESCRIPTION="Arch Linux"
```

The file `/etc/arch-release` exists, but is zero bytes (empty).

pquerna:
Looks like the Rackspace Arch Linux Image has `xe-guest-utilities` installed by default:
https://www.archlinux.org/packages/community/i686/xe-guest-utilities/

And that pulls in the `lsb-release` package, and thats what populates the file.

shirou:
hmm. I used AMI from https://www.uplinklabs.net/projects/arch-linux-on-ec2/ on Amazon EC2. `/etc/arch-releas` exists but empty

And I have asked my friend to check his Arch, but he also said `/etc/lsb-release` and `/usr/bin/lsb_release` does not exists. 

Could you tell me what is the difference?

shirou:
Thanks. So we can add a condition such as if `xe-guest-utilities` is installed, use LSB. Is it easy to know `xe-guest-utilities` is installed or not?

pquerna:
I think we can just look for `/etc/lsb-release` existing?

shirou:
oh ya, indeed. So, your PR fits all. When hosts can get `lsb-release` (it means `xe-guest-utilities` installed), use that. If not, empty. Perfect.

Thank you for the conversation! I will merge it.

PR #135, opened 2016-01-20
Optimize per-process percentage calculation

sparrc:
Few things:

- don't define calculate function on the fly
- only execute time.Now() once, which should marginally improve accuracy too
- only execute p.CPUTimes() once
- In diff, subtract _all_ of the times fields. Even if they aren't filled, it's worth including them in case they do get filled in the future.

Possible improvements:

Make `thisCPUTime` and `thisCPUTimes` fields on the `Process` object, and fill them in when `CPUTimes()` gets called. That way, if users are getting cpu times and usage, we will only call the cpu times function once.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -106,24 +106,15 @@ func PidExists(pid int32) (bool, error) {
 // If interval is 0, return difference from last call(non-blocking).
 // If interval > 0, wait interval sec and return diffrence between start and end.
 func (p *Process) CPUPercent(interval time.Duration) (float64, error) {
-	numcpu := runtime.NumCPU()
-	calculate := func(t1, t2 *cpu.CPUTimesStat, delta float64) float64 {
-		if delta == 0 {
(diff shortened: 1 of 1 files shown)


shirou:
Thank you. It seems performance improvement is a little, but readability improvement is quite big.

> Possible improvements:

Could you tell me what about this possible improvements? Even if `thisCPUTime` field is added, users who want to get *latest* cpu times should call `CPUTimes()`, I think. Or, if make `lastCPUTimes` and `lastCPUTime` exported, users who want to get cpu times *even if not latest* become happy?

sparrc:
@shirou that's true, I didn't think about that, users who want to get the CPU percent at the very moment `CPUPercent` is called would not be happy with that optimization.

shirou:
I will merge this PR. If you have something, please open another PR or issue. 

Thank you very much!

PR #138, opened 2016-01-26
Add -a when calling ps on darwin

sparrc:
Adding a `-a` flag because this way `ps` will list _all_ processes (rather than just the user running the command)

This matches the linux behavior, which reads "numeric" files found in `/proc`, which includes _all_ processes as well.

for reference, here is what running ps -a vs. ps can look like on linux. The first executable is the result of running the gopsutil `process.Pids()` function.

```bash
vagrant@vagrant-ubuntu-trusty-64:~$ /vagrant/nprocs # running len(process.Pids()) on linux
73
vagrant@vagrant-ubuntu-trusty-64:~$ ps ax --no-headers | wc -l # running ps -a (includes the currently running ps process)
74
vagrant@vagrant-ubuntu-trusty-64:~$ ps x --no-headers | wc -l # running ps without -a (current darwin behavior)
4
```

Also added call to `exec.LookPath` for good measure

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -5,6 +5,7 @@ package process
 import (
 	"bytes"
 	"fmt"
+	"os/exec"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


shirou:
Totally agree. Thank you so much!

PR #139, opened 2016-02-02
host_linux: Skip everything that isn't a normal process.

rburchell:
host_darwin does the same filtering. Not doing this gives us some rather strange
entries that likely aren't what we want.

Before:
    {"user":"reboot","terminal":"~","host":"3.10.0-327.4.5.el7.x86_64","started":1454378260}
    {"user":"LOGIN","terminal":"ttyS0","host":"","started":1454378270}
    {"user":"LOGIN","terminal":"tty1","host":"","started":1454378270}
    {"user":"runlevel","terminal":"~","host":"3.10.0-327.4.5.el7.x86_64","started":1454378276}
    {"user":"root","terminal":"pts/0","host":"vpn","started":1454404513}

After:
    {"user":"root","terminal":"pts/0","host":"vpn","started":1454404513}

### DIFF
--- a/host/host_darwin.go
+++ b/host/host_darwin.go
@@ -17,6 +17,9 @@ import (
 	"github.com/shirou/gopsutil/internal/common"
 )
 
+// from utmpx.h
+const USER_PROCESS = 7
+
 func HostInfo() (*HostInfoStat, error) {
 	ret := &HostInfoStat{
 		OS:             runtime.GOOS,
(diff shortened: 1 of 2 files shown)


rburchell:
I haven't added tests for this because I'm uncertain of how to do so in a 'useful' way. Suggestions welcome.

shirou:
Thank you so much!

Could you avoid magic number by using const like this? After that, I will merge it.
```
// from utmp.h
const USER_PROCESS = 7
```

> I haven't added tests for this because I'm uncertain of how to do so in a 'useful' way. Suggestions welcome.

Unfortunately, gopsutil doesn't have enough tests because there are various environments. I have a plan about *mock* but not implemented yet.

rburchell:
> Could you avoid magic number by using const like this? After that, I will merge it.

Done. I also took the liberty to convert the host_darwin usage.

shirou:
merged. Thank you!

PR #141, opened 2016-02-02
fix callps command generate error

### LINKED ISSUE #140: The process.Cmdline will return incorrect result in mac os
if we specific pid of process, it will always return
/sbin/launchd <nil>

ovear:
fixed #140 
it caused by using ps incorrectly

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -415,9 +415,9 @@ func callPs(arg string, pid int32, threadOption bool) ([][]string, error) {
 	if pid == 0 { // will get from all processes.
 		cmd = []string{"-ax", "-o", arg}
 	} else if threadOption {
-		cmd = []string{"-ax", "-o", arg, "-M", "-p", strconv.Itoa(int(pid))}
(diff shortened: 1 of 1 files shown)


ovear:
if use -a option with -p ,the -p option will not work, it will return all process.

like this

> HereticdeMacBook-Pro:main hereric$ ps -ax -o command -p 2111
COMMAND
/sbin/launchd
/usr/libexec/UserEventAgent (System)
/usr/sbin/syslogd
/usr/libexec/kextd
/System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/FSEvent
/System/Library/CoreServices/appleeventsd --server
/usr/libexec/configd
/System/Library/CoreServices/powerd.bundle/powerd
/usr/libexec/airportd
/usr/libexec/warmd
/System/Library/Framework
balabla....

and the first of the result will always be '/sbin/launchd'

but the correct result should use it without -a option, like this

> HereticdeMacBook-Pro:main hereric$ ps -x -o command -p 2111
COMMAND
ping 192.168.200.155

shirou:
When I merge #138, I haven't noticed that. Thank you very much!

PR #144, opened 2016-02-04
There are two possible error scenarios for CallPgrep.  

weberr13:
One indicates a broken system (no pgrep command) and one is a normal error state of pgrep
meaning no processes found for the criteria given (in this case the parent pid does not exist or the process simply has no children).  The later case is quite useful to know about so I added a static error for this case.

### DIFF
--- a/internal/common/common_unix.go
+++ b/internal/common/common_unix.go
@@ -3,10 +3,13 @@
 package common
 
 import (
+	"errors"
 	"os/exec"
 	"strconv"
 	"strings"
 )
+
+var ErrorNoChildren = errors.New("Process does not have children or does not exist")
 
 func CallLsof(invoke Invoker, pid int32, args ...string) ([]string, error) {
 	var cmd []string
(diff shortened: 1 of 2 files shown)


vrecan:
:+1:

shirou:
Thank you for the PR!

I use `exec.LookPath`. So If no pgrep command exists, it returns `executable file not found in $PATH` error.

`invoke.Command(pgrep, cmd...)` uses `exec.Command`, and it may return various kind of error. So, changing every error to `ErrorNoChildren` seems little bit overdo.

I agree that if `pgrep` returns nothing, something wrong. So, how about to check a length of return value and if empty, returns `ErrorNoChildren` error? (please do not capitalize an error string. see [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) )

weberr13:
Updated the request to include your suggestions.

shirou:

PR #145, opened 2016-02-05
try to fix diskio stats on FreeBSD-10.x

cactus:
FreeBSD apparently changed the magic sysctl mib values for devstats. At any rate, the values
I get back from sysctlnametomib are not the same as the hardcoded ones. As a result:

    --- FAIL: TestDisk_io_counters (0.00s)
    disk_test.go:39: error no such file or directory
    disk_test.go:42: ret is empty, map[]

This code attempts to fetch the Mib values during init, and uses
those values for fetching stats. 

Likely needs better error handling during init though.

### DIFF
--- a/disk/disk_freebsd.go
+++ b/disk/disk_freebsd.go
@@ -2,6 +2,14 @@
 
 package disk
 
+/*
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+*/
+import "C"
+
 import (
 	"bytes"
 	"encoding/binary"
@@ -19,6 +27,19 @@ const (
 	KernDevstat    = 974
 	KernDevstatAll = 975
 )
+
+var DiskMib []int32
+
+func init() {
+	cs := C.CString("kern.devstat.all")
+	defer C.free(unsafe.Pointer(cs))
+	omib := [4]C.int{}
+	size := 4
+	C.sysctlnametomib(cs, (*C.int)(unsafe.Pointer(&omib[0])), (*C.size_t)(unsafe.Pointer(&size)))
+	for i := 0; i < size; i++ {
+		DiskMib = append(DiskMib, int32(omib[i]))
+	}
+}
 
 func DiskPartitions(all bool) ([]DiskPartitionStat, error) {
 	var ret []DiskPartitionStat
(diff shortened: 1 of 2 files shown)


cactus:
For reference, here are the values I get for `sysctlnametomib("kern.devstat.all")`:

     [1, 2147482904, 2147482903]


FreeBSD version:

~~~
$ freebsd-version 
10.2-RELEASE-p12
~~~

shirou:
Thank you for the PR. I am worry about MIB changing, so your PR is very correct.
But, gopsutil aims not to use cgo for easy cross-compiling. So I can not accept this PR.

How about implement `sysctlnametomib` by using `syscall.Syscall`? The go runtime [already implemented](https://github.com/golang/go/blob/f78a4c84ac8ed44aaf331989aa32e40081fd8f13/src/syscall/syscall_freebsd.go#L29), but not exported. So we need re-implemet but actually we can.

cactus:
I didn't realize you had a desire to avoid cgo. That is a perfectly valid thing to do, in my opinion. No worries.

I will indeed look at reimplementing the method you linked to -- thanks for pointing it out to me!

cactus:
@shirou I found an apparently undocumented, but exported, method in go stdlib for fetching sysctl by name. It uses nametomib internally. [link to it](https://github.com/golang/go/blob/f78a4c84ac8ed44aaf331989aa32e40081fd8f13/src/syscall/syscall_bsd.go#L452)

Here is a [diff using that](https://github.com/shirou/gopsutil/compare/master...cactus:sysctl-go)

Would you prefer to use the diff I just linked to, or would you prefer to have nametomib that you linked to earlier reimplemented internally, and use that instead?

shirou:
I'm a fool. 
When I read `Sysctl` signature, and it returns only string, so I think we can not get binary. But as you wrote, just cast to `[]byte`, we can use.

I confirmed `syscall.Sysctl("kern.devstat.all")` and my `common.CallSyscall` return equivalent value (just last Null is not exists). So, your diff works well and is more durable about MIB changing. Yes, I prefer this `syscall.Sysctl` approach.

Thank you for suggesting me. I am very pleased to know this, and I will change other codes which use my old `CallSyscall` function.

cactus:
Ok. I will close this one and open a new pull req with the other code.

cactus:
Opened pull request #148 to replace this one.

PR #148, opened 2016-02-06
try to fix diskio stats on FreeBSD-10.x

cactus:
FreeBSD apparently changed the magic sysctl mib values for devstats.

    --- FAIL: TestDisk_io_counters (0.00s)
    disk_test.go:39: error no such file or directory
    disk_test.go:42: ret is empty, map[]

This code uses an undocumented, but exported, go stdlib method to fetch
the sysctl by string instead of mib.


shirou:
Thank you! This works great on my environment.

PR #151, opened 2016-02-10
fix meminfo output on consecutive calls

### LINKED ISSUE #150: proc.MemoryInfo() rss sizes don't match top
proc.MemoryInfo() output says my process is 9mb while top says the same process RES is 35mb. It also never seems to grow, if I write a memory grow it stays between 8 - 9mb while top reports it's over 1-2gb.

I am going to dig more into this but I wanted to see if anyone else is seeing this behavior? When I originally implemented these calls I got very similar results to top.

This is on centos 7 on a quadcore xeon box.

vrecan:
this fixes #150


shirou:
oh, no. 
me, idiot.  Thank you so much!

PR #152, opened 2016-02-10
Make a public function for computing total CPU time

walles:
This function used to be a private part of `process.go`.

Since I needed that functionality however I think it's better to make it public
than for me to copy it into my own code.

As a side effect of this change, I also fixed a bug in the function where Stolen
was not part of the sum. Having the function close to the CPUTimesStat
declaration will make problems like this less likely to re-occur in the future.


shirou:
Thank you! I have no objection to make it public.

PR #153, opened 2016-02-12
Return correct user name from process.Username()

walles:
Before this change, `process.Username()` always returned `root` on Darwin.

With this change in place, it returns the user the process is running as.

This change also adds a regression test for the new behavior and some documentation links.


shirou:
Thank you. I checked on my Darwin environment.

PR #156, opened 2016-02-12
Add build check

shirou:
add build check.


circle.yml
```diff
@@ -1,3 +1,11 @@
 machine:
   timezone:
-    Asia/Tokyo
\ No newline at end of file
+    Asia/Tokyo
+test:
+  override:
+    - GOOS=linux GOARCH=amd64 go test -v ./...
+    - GOOS=linux GOARCH=386 go get -v ./...
```
walles: Can you run the tests on `linux-i386`?

shirou: No. Because `utmp` struct size is different from host amd64, `TestUsers` will fail.

PR #160, opened 2016-02-19
Document + add tests for for mem.VirtualMemory()

walles:
This change updates, documents and adds tests for the (previously undocumented) behavior of `Used`
to "RAM used by programs".

This matches how modern versions of `free` defines `Used` on Linux: https://gitlab.com/procps-ng/procps/commit/6cb75efef85f735b72e6c96f197f358f511f8ed9

We also remove the undocumented and unused `Shared` field of the `VirtualMemoryStat` struct.

So with this change in place, `VirtualMemoryStat` contains:
* three human-consumable fields for `Total`, `Used` and `Available` memory
* one human-consumable `UsedPercentage` field
* a number of kernel specific fields

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -4,18 +4,44 @@ import (
 	"encoding/json"
 )
 
+// Memory usage statistics. Total, Available and Used contain numbers of bytes
+// for human consumption.
+//
+// The other fields in this struct contain kernel specific values.
 type VirtualMemoryStat struct {
-	Total       uint64  `json:"total"`
-	Available   uint64  `json:"available"`
-	Used        uint64  `json:"used"`
+	// Total amount of RAM on this system
+	Total uint64 `json:"total"`
+
+	// RAM available for programs to allocate
+	//
+	// This value is computed from the kernel specific values.
+	Available uint64 `json:"available"`
+
+	// RAM used by programs
+	//
+	// This value is computed from the kernel specific values.
+	Used uint64 `json:"used"`
+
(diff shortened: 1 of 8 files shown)


shirou:
Due to the #161 merge, this PR becomes conflict. I am sorry to annoying you, but could you resolve conflicts? 
Thank you very much!

walles:
Conflicts fixed.

shirou:
merged! Thank you so mu!!

walles:
One thing to consider for this API is if all the kernel specific values should be in a `map[string]uint64`.

I just did `wc -l /proc/meminfo` on a random Linux machine and it said `44`. Keeping up with that won't be possible in the long run, and I also think it's impossible to know which values people will be interested in.

This is something I most likely won't look into, but I thought I'd document it somewhere for posterity.

shirou:
In Non-Linux system, there are no `proc`, no common method to get memory information. It is not easy to get same information of Linux.
And also, `map[string]` is easy to write, but I think it narrows a compile advantage.

walles:
On OS X, if you call [`host_statistics`](http://web.mit.edu/darwin/src/modules/xnu/osfmk/man/host_statistics.html) you'll get a `vm_statistics` struct back:
http://www.opensource.apple.com/source/xnu/xnu-1456.1.26/osfmk/mach/vm_statistics.h

Simply adding the names / values to the map should be fine for anybody wanting to trawl the depths of kernel memory usage on OS X.

PR #163, opened 2016-02-24
Add test for process.CPUTimes()

walles:
Test validated on an OS X laptop from 2011, both on OS X (native) and Linux (in a VirtualBox).

Measured CPU times vary from 0.19-0.21 over ten runs on both platforms, which is well within the accepted marigins of 5x too large (1s) to 5x too small (0.04s).

I realize this could potentially intermittently fail on a system that is totally overloaded, but given my trial runs I think it's safe.

### DIFF
--- a/process/process_test.go
+++ b/process/process_test.go
@@ -1,6 +1,7 @@
 package process
 
 import (
+	"fmt"
 	"os"
 	"os/user"
 	"runtime"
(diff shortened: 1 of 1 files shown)


shirou:
Thank you. I confirmed my FreeBSD environment can also pass.

Issue #165, opened 2016-02-27
Process.Cmdline() makes it hard to parse individual arguments


cgilling:
Hi, first off, thanks for writing this library, I've been looking for something like this for a while and finally stumbled upon this one.

I was wondering if there is some way that the info from `Cmdline` could be exposed as a `[]string` rather than just a `string`, the problem I have is that it makes it hard to get the real list of command line parameters, for example if I called a binary with something like:

```bash
./test 2342 "this is a single arg" hi there
```

I don't think I would be able to find out that `this is a single arg` was really just one argument

Perhaps a new function could be added to expose the it in slice form rather than just string form?

cgilling:
From what I can tell from the `linux`, `darwin` and `freebsd` implementations this would be pretty easy to implements, because all the args are segmented by `byte(0)`, but the windows implementations perhaps not as easy and it just seems to be a strings. Would that be a non-starter?

shirou:
Since #167 has been merged, I close this issue. Thank you so much!

PR #167, opened 2016-02-28
process: add CmdlineSlice function for linux + freebsd

cgilling:
This is the implementation for #165

This allows for getting more exact information about each argument
especially if there are arguments that have spaces in them.

This was not implemented for darwin or for windows because they
both currently have not way of properly parsing the cmdline string.
Darwin parses the output of 'ps' which is already whitespace
segmented, and windows just has the cmdline string.

NOTE: I have only tested this on `linux` as I don't have a `freebsd` install. I did run `GOOS=freebsd go test -c` to ensure that the `freebsd` version compiles.

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -88,6 +88,9 @@ func (p *Process) Cmdline() (string, error) {
 	}
 	return strings.Join(r[0], " "), err
 }
+func (p *Process) CmdlineSlice() ([]string, error) {
+	return nil, common.NotImplementedError
+}
 func (p *Process) CreateTime() (int64, error) {
 	return 0, common.NotImplementedError
 }
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -81,6 +81,9 @@ func (p *Process) Name() (string, error) {
 func (p *Process) Exe() (string, error) {
 	return "", common.NotImplementedError
 }
+
+// Cmdline returns the command line arguments of the process as a string with
+// each argument separated by 0x20 ascii character.
 func (p *Process) Cmdline() (string, error) {
 	r, err := callPs("command", p.Pid, false)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


shirou:
I am very appreciate to your contribution. To get all cmd line is good for some users, and separate functions to maintain compatibility is exactly what I wanted. (But, I want to replace `Cmdline` by your `CmdlineSlice` near future.)

I confirmed it worked on FreeBSD 10 environment like a charm.

```
% cat - blah "AB CD"
[]string{"cat", "-", "blah", "AB CD"}
```

And, how about to use [go-shlex](https://github.com/flynn-archive/go-shlex) to split a cmdline on Darwin and Windows?

Thank you!

cgilling:
Thanks for the pointer to that `go-shlex` library. I'll work on integrating that. I've been looking for something exactly like that but had trouble finding it, the same project I plan of using `gopsutil` for `go-shlex` will be quite useful as well.

cgilling:
Hmmm just tried this out and don't think it will work, I ran:
```sh
go run waitForSigint.go "this is a test"
```
and then this is the output from `ps`:

```sh
> ps -x -o command -p 86968
COMMAND
/var/folders/yl/dgf4ylw56wl1bng7wlx2ypbm0000gn/T/go-build504897322/command-line-arguments/_obj/exe/waitForSigint this is a test
```

Seems like it doesn't work because the ps output doesn't keep the quote characters.

Any other suggestions?

shirou:
hmm. Sorry, I haven't try that.

Now we know we can not split applicable. But I think the difference between platform is acceptable. So could you change freebsd and windows to split just a space, and add some comments to note users? If user really want to get applicable arguments by themselves, they can just `strings.Join(a, " ")`.

cgilling:
@shirou, I made the requested changes. Unfortunately I am unable to test the windows changes as I don't have a windows setup.

shirou:
I confirmed it works perfect at least Windows7.
Thank you very much about joining the discussion and awesome PR!

Issue #166, opened 2016-02-27
linux: process.NewProcess() slow


cgilling:
I was testing this library and found that (at least on linux) `process.NewProcess` ends up being pretty slow because it calls [p.fillFromStatus](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L70) to check and see if its a valid process. It seems like simply stating the `/proc/[pid]` file would be enough to check and see if the pid is valid

It seems like the only function that depends on it having been called is [Name()](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L70) which I would think could be managed differently (for example one could check to see if `p.name == ""` and if so then call `p.fillFromStatus`.

I'm not sure if this would be an acceptable change but would be helpful for me as I would like to use this library to scan running processes for a certain process with specific command line arguments, and having `p.fillFromStatus` every time slows this down quite a bit.

If this is acceptable I could write up a pull request.

shirou:
Since #168 is merged, I close this issue. If you want to let me know something more, please feel free to re-open.

Thank you for your suggestion and great PR!

cgilling:
thanks for the quick responses, makes working on the project even across timezones quite nice.

PR #168, opened 2016-02-28
process: change linux NewProcess to only stat /proc/[pid]

cgilling:
Implements #166

Before it was doing a fillFromStatus() call which was much slower
and none of the information was needed, except by the Name() func
which now will call fillFromStatus() if p.name is not set.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -61,13 +61,15 @@ func (m MemoryMapsStat) String() string {
 	return string(s)
 }
 
-// Create new Process instance
-// This only stores Pid
+// NewProcess creates a new Process instance, it only stores the pid and
+// checks that the process exists. Other method on Process can be used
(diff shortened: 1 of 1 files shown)


shirou:
Indeed. When I write at first, `status` is a important information, status is read whenever `NewProcess` called. But after implement, it changed not a `process` struct create time, method invoked time. So, your suggestion is correct. 
I want to merge it, but, could you change some little?

1. use `common.HostProc`. It allowed to change other proc filessystem mount point such as inside a Docker.
2. Add a comment, such as `// check only the process id is valid.` or else.

Thank you so much!

cgilling:
@shirou, I just made the requested changes and rebased them back into a single commit. Let me know if the way I wrote the comment is good. Thanks.

shirou:
Great! Thank you for your contribution!!

PR #171, opened 2016-03-03
Provide correct length for slice creation.

raviparimi:
- Creating a slice of length numFDs and later appending elements to it
  results in a slice whose first numFDs elements are all nil. It is
  sufficient to create a slice of zero length since we are appending
  elements to it.
- The current allocation will make ret a slice of length 0. What's
  needed is a slice of length len(ofs).


shirou:
Thank you! A test for the `OpenFiles()` is not provided. I will add later.

PR #185, opened 2016-04-18
Remove the requirement to use lost for pid …

freman:
…by using the information provided in status to get the parent pid

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -20,6 +20,7 @@ type Process struct {
 	Pid            int32 `json:"pid"`
 	name           string
 	status         string
+	parent         int32
 	numCtxSwitches *NumCtxSwitchesStat
 	uids           []int32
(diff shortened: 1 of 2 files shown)


shirou:
Thank you! This is very important but I forget.

But on my Ubuntu environment, `Ppid` is `PPid`. So, could you change to parse `PPid` also like this?

```
+               case "Ppid", "PPid":
```

And, the test failed by
```
# github.com/shirou/gopsutil/process
process/process_linux.go:124: undefined: fmt in fmt.Errorf
```

Could you add "fmt" to imports?

freman:
Yeh, sure, that was slack copy/pasting from a ubuntu server back to my macbook, was going to fix it but got distracted.

shirou:
It works now. Thank you for your contributing!

PR #193, opened 2016-04-27
Fix definitions in type_linux and regenerate files

### LINKED ISSUE #179: FTBFS on "ppc64el" and "arm64"
As reported in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=820466

>It looks like the existing
host_linux_arm.go is actually meant for arm64; please recharacterize it
appropriately and add a version suitable for 32-bit ARM (armel/armhf).

_gopsutil_ FTBFS on _ppc64el_ and _arm64_:

~~~~
src/github.com/shirou/gopsutil/host/host_linux.go:108: undefined: utmp
src/github.com/shirou/gopsutil/host/host_linux.go:109: invalid expression unsafe.Sizeof(u)
src/github.com/shirou/gopsutil/host/host_linux.go:117: undefined: utmp
~~~~

Please investigate.

discordianfish:
I've change the types_linux.go so the generated timeval struct name matches the reference in utmp. Then I generated the files. But I've cheated: Since I don't have the toolchain for 386 and arm, I manually edited them. This should fix #179


discordianfish:
Ok looks like this broke utmp parsing. Looking at it right now.

discordianfish:
I have to admit, I don't really know what I'm doing here.. So I'm just gonna add a amd64 file and keep the rest as it is.

PR #200, opened 2016-05-10
Fix compile errors for iOS/arm64

azinman:
- Added conditional preprocessor guard on cpu_darwin_cgo.go

- Duplicated disk_darwin_amd64 for arm64 (after confirming
that sys/mount.h are the same between the two platforms, hence the
constants should be valid).


shirou:
I can not confirm but this looks great! Thank you so much!

Issue #201, opened 2016-05-18
Implement timeouts for exec command runners


sparrc:
Currently gopsutil has a few places where OS commands are run using the `exec` library.

Any one of these commands _could_ hang, causing the gopsutil call to hang. While the hang is an OS problem, gopsutil should be able to see it and return an error.

Unfortunately the exec library doesn't have any built-in support for timeouts, but these can be implemented relatively easily using timers (see https://github.com/influxdata/telegraf/blob/master/internal/internal.go#L153).

I likely won't have time to work on this for a while, but I wanted to put the feature request out there, and I will try to find some time for this soon if I can.

shirou:
#203 is merged. Thank you!

PR #203, opened 2016-05-20
add Timeout to invoke command and use common.Invoke

shirou:
refs: #201

Thanks  influxdata telefgaf!


sparrc:
great! thanks @shirou! 👍

PR #205, opened 2016-06-01
Calc cpu percent using last variables if interval of 0 is given

### LINKED ISSUE #202: cpu percent(0, true) fails to return expected results
the output is always either this:
```
[]float64) (len=16 cap=16) {
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0
}
```
or this
```
[]float64) (len=16 cap=16) {
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 100,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0,
 (float64) 0
}
```

previously if I would make multiple calls it would give me the percentage used between the two calls. Now it seems like it's giving me whatever it see's right when I call it.

What I am after here is a simple call I can make multiple times and each time it gives me the difference between the two calls.

vrecan:
This fixes #202

### DIFF
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -5,6 +5,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 )
 
 type TimesStat struct {
@@ -37,8 +38,21 @@ type InfoStat struct {
 	Flags      []string `json:"flags"`
 }
 
-var lastCPUTimes []TimesStat
-var lastPerCPUTimes []TimesStat
+//CPUPercent
+type cpuPercent struct {
+	sync.Mutex
+	lastCPUTimes    []TimesStat
+	lastPerCPUTimes []TimesStat
+}
+
+var lastCPUPercent cpuPercent
+
+func init() {
+	lastCPUPercent.Lock()
+	lastCPUPercent.lastCPUTimes, _ = Times(false)
+	lastCPUPercent.lastPerCPUTimes, _ = Times(true)
+	lastCPUPercent.Unlock()
+}
 
 func Counts(logical bool) (int, error) {
 	return runtime.NumCPU(), nil
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -39,14 +39,13 @@ type InfoStat struct {
 	Flags      []string `json:"flags"`
 }
 
-//CPUPercent
-type cpuPercent struct {
+type lastPercent struct {
 	sync.Mutex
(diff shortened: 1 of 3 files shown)


vrecan:
I have this running in our internal app that collects our node stats and everything looks good. I also ran the tests with the race detector and all the proc tests pass. 

Side note: 
I do see a one test failing on my linux box:
```
--- FAIL: TestGetProcInodesAll (0.01s)
        Error Trace:    net_linux_test.go:20
	Error:  	Should NOT be empty, but was map[]		
```

Not related to what I changed but I thought you might want to know about it.

cpu/cpu.go
```diff
@@ -2,11 +2,11 @@ package cpu
 
 import (
 	"encoding/json"
+	"github.com/shirou/gopsutil/internal/common"
```
KenjiTakahashi: Why mix these? Hint: Use `goimports` for the greater good :-).

cpu/cpu_unix.go
```diff
@@ -7,24 +7,46 @@ import (
...
+	if t2Busy <= t1Busy {
+		return 0
+	}
+	if t2All <= t1All {
+		return 1
+	}
+	return (t2Busy - t1Busy) / (t2All - t1All) * 100
+}
+
+func calcualteALLBusy(t1, t2 []TimesStat) ([]float64, error) {
```
KenjiTakahashi: Function name typo.

vrecan: thanks, fixed

cpu/cpu.go
vrecan: goimports never plays nice for me with remote mounting filesystems in sublime.

shirou:
Great work! Thank you !!

I run tests several times but I could not get this error. So, I merged this PR. If someone meets this, please send information.
```
--- FAIL: TestGetProcInodesAll (0.01s)
        Error Trace:    net_linux_test.go:20
    Error:      Should NOT be empty, but was map[]  
``

PR #207, opened 2016-06-03
Fix potential panic in linux disk IO counters

sparrc:
Old kernels have a bug in diskstats where lines can have less than 14
fields. This applies to the kernel present in RHEL 5.2 and earlier.

It's a bit of a niche but probably best to patch to be safe from future
bugs too.

RHEL bug case:
https://bugzilla.redhat.com/show_bug.cgi?id=583285

Encountered in Telegraf:
https://github.com/influxdata/telegraf/issues/1322


shirou:
Looks great. Thank you!

PR #217, opened 2016-06-13
[net]linux Add support for linux net interface fifo errors

sandlbn:
Some of the fields like fifo from /proc/net/dev where not exposed in the net package.


shirou:
Thank you! I will try to implement on other platform.

PR #222, opened 2016-06-22
Casting the values of working set and page file to uint64

diptanu:
This fixes the build on `windows/386` platforms.


shirou:
Thank you!

PR #226, opened 2016-07-03
Return a valid JSON `String()` output for network lists

moul:
Hi, what do you think about JSON marshalling lists as you are already doing with standalone items ?

If you like the idea, I can continue on the other resource types

---

Before

``` go
fmt.Println(myList)
// output: [{"a": "hello", "c": 42} {"a": "world", "c": 1337}] -> invalid json
```

After

``` go
fmt.Println(myList)
// output: [{"a": "hello", "c": 42},{"a": "world", "c": 1337}] -> valid json
```

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -54,18 +54,25 @@ type ProtoCountersStat struct {
 	Stats    map[string]int64 `json:"stats"`
 }
 
-// NetInterfaceAddr is designed for represent interface addresses
+// InterfaceAddr is designed for represent interface addresses
 type InterfaceAddr struct {
 	Addr string `json:"addr"`
 }
 
+// InterfaceAddrList is a list of InterfaceAddr
+type InterfaceAddrList []InterfaceAddr
+
+// InterfaceStat represents the basic information of an interface
 type InterfaceStat struct {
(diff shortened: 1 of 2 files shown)


QuentinPerez:
LGTM

shirou:
Sorry, could you tell me more detail about what is a problem?

```
	V := InterfaceStat{
		MTU:  1500,
		Name: "eth0",
		Addrs: []InterfaceAddr{
			{Addr: "1.2.3.4"}, {Addr: "5.6.7.8"},
		},
	}
        fmt.Println(V)
```

returns `{"mtu":1500,"name":"eth0","hardwareaddr":"","flags":null,"addrs":[{"addr":"1.2.3.4"},{"addr":"5.6.7.8"}]}` and it is valid.

moul:
Your example is correct, it works perfectly for standalone items, but JSON is invalid as soon as you print a list (space separator instead of comma)

shirou:
I got it. You want to use `InterfaceStat.Addrs` individually. Thank you for teach me.

If you make types for other items, I will appreciate and merge it.

moul:
Perfect :)

shirou:

Issue #227, opened 2016-07-11
host.Info() -> Procs?


prologic:
What is `host.Info()` and the `InfoStat->Proc` field meant to represent? The no. of processes running on the system? If so; this consistently shows `0` for me on Darwin / OS X

prologic:
Thanks!


James Mills / prologic

E: prologic@shortcircuit.net.au
W: prologic.shortcircuit.net.au

PR #228, opened 2016-07-11
Correctly detect host Procs on all platforms.

sean-:
Previously this only worked on Windows.  Expand to all supported platforms.

Fixes: #227


shirou:
Thanks!

But unfortunately, import cycle is happened on linux. We need only number of procs, so, could you add a new function on `host/host_linux.go` such as,

```
func procs() (uint64, error) {
	f, err := os.Open(common.HostProc())
	if err != nil {
		return 0, err
	}
	list, err := f.Readdir(-1)
	f.Close()
	if err != nil {
		return 0, err
	}
	return uint64(len(list)), error
}
```

sean-:
@shirou There you go.  LMK!

shirou:
You implemented  `NumProcs` in the common. Really good! Thank you for your contribution!

prologic:
Thanks!

PR #232, opened 2016-07-21
add disk IoTime on freebsd and fix read & write time calculation

phemmer:
This PR does 2 things.

1. It adds the `IoTime` metric to FreeBSD
2. It fixes the the disk time calculations.

In regards to number 2, the `Compute()` function was effectively doing nothing. It was converting `BINTIME_SCALE` to a `uint64`. But because the value is a float < 0, when converted to `int64`, it becomes `0`. So the function effectively just returned the field `b.Sec`. This fixes the calculation by preserving the value as a float.
Also relatedly, the times were being returned in seconds, when the raw values offer much higher precision than that. Since the linux values are returned in milliseconds, I opted to keep the type (a `uint64`), but am now returning the values measured in milliseconds instead of seconds. *(It would probably be a good idea to use a `time.Duration` instead of `uint64`)*

### DIFF
--- a/disk/disk_freebsd.go
+++ b/disk/disk_freebsd.go
@@ -124,8 +124,9 @@ func IOCounters() (map[string]IOCountersStat, error) {
 			WriteCount: d.Operations[DEVSTAT_WRITE],
 			ReadBytes:  d.Bytes[DEVSTAT_READ],
 			WriteBytes: d.Bytes[DEVSTAT_WRITE],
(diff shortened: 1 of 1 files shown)


shirou:
Confirmed on my FreeBSD 10, great contribution. Thanks a lot! 

> (It would probably be a good idea to use a time.Duration instead of uint64)

Hmm, you're partly right. However, I want to avoid API change, and convert `time.Duration` from milliseconds is easy. so, I prefer it keeps.

PR #233, opened 2016-07-22
Add some dummy implementation for nacl

razvanm:
This patch is not exhaustive. I only patched the functions that didn't
work in the code I was trying to compile.

### DIFF
--- a/cpu/cpu_nacl.go
+++ b/cpu/cpu_nacl.go
@@ -1,0 +1,15 @@
+// +build nacl
+
+package cpu
+
+import (
+	"time"
+)
+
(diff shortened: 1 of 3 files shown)


shirou:
Thanks, but is this a complete PR, or just a WorkInProgress?

gopsutil on nacl is very interesting.

razvanm:
This fixes the code I was interested so this is all from my perspective. I'm open to concrete suggestion on what else to add. :-)

shirou:
Then, please implement on your forked repository. If you want to discuss with other contributors, could you open an issue and re-open PR after discussion?

Thank you.

Issue #235, opened 2016-07-24
IO counters for bytes read and write all zero


ncabatoff:
At least on my Ubuntu 14.04 system, proc(5) specifies that /proc/[pid]/io contains these fields:

              rchar: characters read
              wchar: characters written
              syscr: read syscalls
              syscw: write syscalls
              read_bytes: bytes read
              write_bytes: bytes written
              cancelled_write_bytes:

However Process.fillFromIO() looks for the fields syscr, syscw, readBytes, and writeBytes.  Googling suggests that this is a simple error/typo and not something they changed in some version of the kernel.

shirou:
Thank you for your contrition! Your great PR #236 fixes it. So I close this issue.

PR #236, opened 2016-07-24
fix incorrect names for fields in proc/pid/io

ncabatoff:
Fix issue #235


shirou:
Thank you!! it was introduced replacing json type name on gopsutil 2.0...

PR #237, opened 2016-08-11
Add HostID to Host

sean-:
Add `host.Info().HostID` which returns a "durable-enough" UUID to distinguish a host.


shirou:
Very interesting contribution. Thanks a lot!

PR #238, opened 2016-08-16
Added fields to VirtualMemoryStat

kthommandra:
Added the writeback, dirty and writebacktmp fields from /proc/meminfo… on Linux

Tests done:
- unit test using mem_test.go
- used in Telegraf system plugin (with appropriate changes in Telegraf) and confirmed that new fields were being properly inserted into InfluxDB measurement


shirou:
Thank you!

PR #239, opened 2016-08-16
Ensure "make check" can succeeed if it doesn't find any golint errors

lfittl:
In the Makefile, when running `golint` in `check` we're using `egrep -v` to check for any errors except for those that match the whitelist.

Unfortunately `egrep` will return an errorcode if it doesn't find anything and the `make check` task will return an error.

This commit flips the logic, so `make check` is successful if there are no linting errors.


shirou:
Exactly! Honestly, I forget about check task is in the Makefile...

PR #241, opened 2016-08-16
Add separate kernelVersion field to host information

lfittl:
This is mostly intended for Linux, where we are returning the OS version
in the PlatformVersion field, which seems reasonable. Often it is still
useful to know which Linux kernel is running (uname -r).

For FreeBSD and Darwin the kernel version matches the platform version,
since they previously used the kernel version for the platform version.

For Windows the kernel version is empty, since there is no clear way
to determine it.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -18,11 +18,12 @@ type InfoStat struct {
 	Hostname             string `json:"hostname"`
 	Uptime               uint64 `json:"uptime"`
 	BootTime             uint64 `json:"bootTime"`
-	Procs                uint64 `json:"procs"`          // number of processes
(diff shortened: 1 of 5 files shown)


shirou:
Cool. But how about to read `/proc/version` and parse it? Reading filesystem is more stable than invoke command.

lfittl:
@shirou Good catch - I've updated this to use `/proc/sys/kernel/osrelease` (which is the same place that `uname -r` gets the data from)

shirou:
`/proc/sys/kernel/osrelease` is best choice. Thanks a lot!

PR #242, opened 2016-08-16
Add missing disk IO statistics for merged reads/writes, IOPS in progress

lfittl:
This is the missing piece so I can replace my previous platform-specific logic with `gopsutil`. It adds:

* Number of merged write operations
* Number of merged read operations
* Number of I/O operations in progress

It seems reasonable to add those, even though its just Linux for now, since other platforms likely provide similar data, and the alternative (having to read `/proc/diskstats` yourself in addition to using `gopsutil`) seems worse.

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -33,15 +33,18 @@ type PartitionStat struct {
 }
 
 type IOCountersStat struct {
-	ReadCount    uint64 `json:"readCount"`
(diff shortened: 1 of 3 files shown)


shirou:
Thank you a lot! You are great contributor. I am very appreciate.

lfittl:
@shirou Thank you for the great library - I'll send follow-ups for the other PRs shortly :)

PR #243, opened 2016-08-18
Add CGO implementation of Process.Exe on OS X

ppanyukov:
A the moment we have `Process.Exe` "not implemented" for OS X.

If CGO is an option for darwin, it can be done like this.

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -79,7 +79,7 @@ func (p *Process) Name() (string, error) {
 	return common.IntToString(k.Proc.P_comm[:]), nil
 }
 func (p *Process) Exe() (string, error) {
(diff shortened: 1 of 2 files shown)


ppanyukov:
Not sure why circleci failed. Is it failing to get all the source code?

shirou:
Thank you so much. However, gopsutil aim to be *pure golang* library. If you want to include CGO function, could you separate `_cgo.go` and `_nocgo.go` like  [cpu](https://github.com/shirou/gopsutil/tree/master/cpu) ?

ppanyukov:
yes, sure will do @shirou

Lomanic:
Superseded by #757.

PR #246, opened 2016-08-21
Implement CreateTime for darwin.

theothertomelliott:
Parses elapsed time and converts this to epoch time.
Passes CreateTime unit test on OSX and verified using a separate app.


shirou:
sorry for a little bit late response. It works for me on darwin. Thank you for your contribution!

PR #248, opened 2016-08-24
Addressing frequency unit discrepancies

dvusboy:
* for Darwin, it is a minor tweak for readability: the value
  returned is in Hz, so using a variable named 'hz' makes more
  sense than 'mhz'
* for Linux, the unit is in kHz so we need to divide the value
  from `cpuinfo_max_freq` by 10^3 to get MHz (see
  cpu-freq/user-guide.txt of the kernel documentation)


shirou:
Yes, it is! Thank you!

PR #251, opened 2016-08-26
Correctly handle long running processes on OSX

theothertomelliott:
The full etime format is dd-hh:mm:ss. Replacing '-' with ':' fixes parsing of this format.


shirou:
Sorry for late response. And thank you for your contirbution!

PR #253, opened 2016-08-29
Standardize `Mhz` to mean maximum CPU frequency on Linux platform

### LINKED ISSUE #249: [FEATURE] Handling CPU frequency scaling
With modern CPUs that support variable clock speed for the purpose of power efficiency, it would be nice for `cpu.InfoStat` to return minimum, maximum and current frequencies if they are available. AFAIK, these are available on Linux as well as Darwin.

On Linux, these are available under `sysfs` at `/sys/devices/system/cpu/cpuX/cpufreq/cpuinfo_{min,max,cur}_freq`. On Darwin, besides the `hw.cpufrequency` key already in use, we also have `hw.cpufrequency_{min,max}`. As for FreeBSD, the picture seems to be more complicated and I don't have the resource to investigate.

Specifically, the importance of having access to the maximum CPU frequency is made apparent in resource management frameworks. For example, in this [Nomad issue](https://github.com/hashicorp/nomad/issues/1392), because the reported MHz in `cpu.InfoStat` is the value from `/proc/cpuinfo`, it is actually the current CPU frequency, typically when the machine is idle. It means Nomad CPU fingerprinting will not reflect the true CPU resource available, because the idle frequency is often substantially lower than the one at full blast.

dvusboy:
* resolve #249
* in `cpu_windows.go`, `Mhz` is the value of `MaxClockSpeed`
* on Linux platform, the `Mhz` value is extracted from `/proc/cpuinfo`
  which reflects the current clock speed; treat this as the fallback
  value instead
* read from `cpufreq/cpuinfo_max_freq` under sysfs to get the
  maximum clock speed for `Mhz`, just like for Windows platform
* also fix the path to `cpu.CoreID` value; the filename is `core_id`

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -65,22 +65,29 @@ func sysCPUPath(cpu int32, relPath string) string {
 }
 
 func finishCPUInfo(c *InfoStat) error {
-	if c.Mhz == 0 {
-		lines, err := common.ReadLines(sysCPUPath(c.CPU, "cpufreq/cpuinfo_max_freq"))
-		if err == nil {
-			value, err := strconv.ParseFloat(lines[0], 64)
-			if err != nil {
-				return err
-			}
-			c.Mhz = value/1000.0  // value is in kHz
-		}
-	}
+	var lines []string
+	var err error
+	var value float64
+
 	if len(c.CoreID) == 0 {
-		lines, err := common.ReadLines(sysCPUPath(c.CPU, "topology/coreId"))
+		lines, err = common.ReadLines(sysCPUPath(c.CPU, "topology/core_id"))
 		if err == nil {
 			c.CoreID = lines[0]
 		}
 	}
+
+	// override the value of c.Mhz with cpufreq/cpuinfo_max_freq regardless
+	// of the value from /proc/cpuinfo because we want to report the maximum
+	// clock-speed of the CPU for c.Mhz, matching the behaviour of Windows
+	lines, err = common.ReadLines(sysCPUPath(c.CPU, "cpufreq/cpuinfo_max_freq"))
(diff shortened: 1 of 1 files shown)


shirou:
CircleCI and my Ubuntu 14.04.4 LTS(3.13.0-85-generic) don't  have `/sys/devices/system/cpu/cpu0/cpufreq` directory. Do you know when is introduced or which distribution has?

lfittl:
@shirou I think this depends on the system - the `cpufreq` directory will only be there if the system is actually capable of changing the CPU frequency.

If it is not, falling back to `/proc/cpuinfo` should be correct I think.

dvusboy:
AFAIK, this is introduced as far back as the first commit of the kernel in git. However, I've found that under virtualization, the `cpufreq` may not be exposed at all. I'll update the code to make it not to return error if we managed to extract some non-zero value out of `/proc/cpuinfo`.

lfittl:
@dvusboy @shirou If it helps, here is one reference I found that says it only exists if the system is actually enabled to change CPU frequency:

```
If the OS is Linux, you can manually control the CPU speed by reading and writing some virtual files in the "/proc" and "/sys" virtual file systems.  You have to be root to set these.  Here are some details; I hope this is helpful...

1.) Is the system capable of software CPU speed control?
If the "directory" /sys/devices/system/cpu/cpu0/cpufreq exists, speed is controllable.
-- If it does not exist, you may need to go to the BIOS and turn on EIST and any other C and P state control and visibility, then reboot
```
https://software.intel.com/sites/default/files/comment/1716807/how-to-change-frequency-on-linux-pub.txt

shirou:
Thanks @dvusboy and @lfittl ! It works for my environment.

PR #254, opened 2016-08-30
OSX Workaround: support interface with multiple digits

bclermont:
On OSX, `netstat -ibdnW` strip every digits out of interface name, if it's more than a single digit. example:

```
$ ifconfig -a
lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> mtu 16384
       	options=3<RXCSUM,TXCSUM>
       	inet6 ::1 prefixlen 128
       	inet 127.0.0.1 netmask 0xff000000
       	inet6 fe80::1%lo0 prefixlen 64 scopeid 0x1
       	nd6 options=1<PERFORMNUD>
gif0: flags=8010<POINTOPOINT,MULTICAST> mtu 1280
stf0: flags=0<> mtu 1280
utun81: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1500
utun83: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1500
utun85: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1500
utun86: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1500
utun88: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1500
``` 

and the output of `netstat`:

```
$ netstat -ibdnW
Name  Mtu   Network       Address            Ipkts Ierrs     Ibytes    Opkts Oerrs     Obytes  Coll Drop
lo0   16384 <Link#1>                         31241     0    3769823    31241     0    3769823     0   0
lo0   16384 ::1/128     ::1                  31241     -    3769823    31241     -    3769823     -   -
lo0   16384 127           127.0.0.1          31241     -    3769823    31241     -    3769823     -   -
lo0   16384 fe80::1%lo0 fe80:1::1            31241     -    3769823    31241     -    3769823     -   -
gif0* 1280  <Link#2>                             0     0          0        0     0          0     0   0
stf0* 1280  <Link#3>                             0     0          0        0     0          0     0   0
utun8 1500  <Link#88>                          286     0      27175        0     0          0     0   0
utun8 1500  <Link#90>                          286     0      29554        0     0          0     0   0
utun8 1500  <Link#92>                          286     0      29244        0     0          0     0   0
utun8 1500  <Link#93>                          286     0      28267        0     0          0     0   0
utun8 1500  <Link#95>                          286     0      28593        0     0          0     0   0
```

All those 5 interfaces `utun8*` are considered as only `utun8` and only one is really processed.

This change use `ifconfig -l` first to list all interfaces, and then invoke `netstat` for each of them.

Slower, but it works.

### DIFF
--- a/net/net_darwin.go
+++ b/net/net_darwin.go
@@ -7,40 +7,53 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
-
-	"github.com/shirou/gopsutil/internal/common"
 )
 
-// example of netstat -idbn output on yosemite
+// example of `netstat -ibdnWI lo0` output on yosemite
 // Name  Mtu   Network       Address            Ipkts Ierrs     Ibytes    Opkts Oerrs     Obytes  Coll Drop
 // lo0   16384 <Link#1>                        869107     0  169411755   869107     0  169411755     0   0
 // lo0   16384 ::1/128     ::1                 869107     -  169411755   869107     -  169411755     -   -
 // lo0   16384 127           127.0.0.1         869107     -  169411755   869107     -  169411755     -   -
 func IOCounters(pernic bool) ([]IOCountersStat, error) {
+	const endOfLine = "\n"
+	// example of `ifconfig -l` output on yosemite:
+	// lo0 gif0 stf0 en0 p2p0 awdl0
+	ifconfig, err := exec.LookPath("/sbin/ifconfig")
+	if err != nil {
+		return nil, err
+	}
+
 	netstat, err := exec.LookPath("/usr/sbin/netstat")
 	if err != nil {
 		return nil, err
 	}
-	out, err := invoke.Command(netstat, "-ibdnW")
+
+	// list all interfaces
+	out, err := invoke.Command(ifconfig, "-l")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


net/net_darwin.go
```diff
@@ -7,44 +7,57 @@ import (
...
 		}
-		exists = append(exists, values[0])
+
+		// only the first output is fine
+		values := strings.Fields(lines[1])
 
 		base := 1
 		// sometimes Address is ommitted
-		if len(values) < 11 {
+		if len(values) < 12 {
```
bclermont: this is actually an other bug fix

shirou:
Sorry for late response. I have not noticed about netstat truncates interface name. Use `ifconfig` to get all interface is very good idea, but it becomes slow.

If the problem is only truncated interface name, how about checking duplicate by netstat for all interface, and if there are duplicated, then run `ifconfig` and individual `netstat` ?
By using this fallback method, speed is same as current implementation in best case.

bclermont:
I changed the logic to make itself figure if interface name had been truncated or not. and it will run ifconfig and netstat to get the stats of only the interfaces that was truncated.

I couldn't figure a more simple implementation

shirou:
Thank you so much!

PR #255, opened 2016-09-04
Fix: wrong swap values on i686 / 4GB ram.

lberruti:
According to sysinfo manpages: swap fields need to be interpreted along
with the mem_unit (Unit) field.

See also http://stackoverflow.com/a/4229727


shirou:
Indeed. On my environment, sysinfo.Unit is `1`, there are no problem at this moment. But your patch is necessary for other environment. Thank you for your contribution.

PR #259, opened 2016-09-16
process: determine page sizes via function

### LINKED ISSUE #258: process: determine page sizes via function
The current implementation for linux uses constants for the page sizes which is not ideal, as some architectures support different sizes.
A better solution would be to use Go's [`Getpagesize`](https://golang.org/pkg/os/#Getpagesize) function to determine the used page size. This is already being used in `mem_darwin_*.go`.

If there are no objections, I'll gladly write a patch for this.

monstermunchkin:
Instead of hard-coding the page size for linux systems, use Go's
`Getpagesize` function.

This resolves #258.

Signed-off-by: Thomas Hipp <thipp@suse.de>

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -20,7 +20,10 @@ import (
 	"github.com/shirou/gopsutil/net"
 )
 
-var ErrorNoChildren = errors.New("process does not have children")
+var (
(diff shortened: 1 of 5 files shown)


shirou:
Thank you for your great contrition! I will add on FreeBSD/Darwin/Windows.

Issue #230, opened 2016-07-17
FTBFS on "ppc64el"


onlyjob:
Version 2.1 FTBFS on _ppc64el_ as follows:

~~~~
# github.com/shirou/gopsutil/process
src/github.com/shirou/gopsutil/process/process_linux.go:559: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:560: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:581: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:582: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:583: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:584: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:585: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:586: undefined: PageSize
src/github.com/shirou/gopsutil/process/process_linux.go:723: undefined: ClockTicks
src/github.com/shirou/gopsutil/process/process_linux.go:724: undefined: ClockTicks
src/github.com/shirou/gopsutil/process/process_linux.go:724: too many errors
~~~~

See full build log: https://buildd.debian.org/status/fetch.php?pkg=nomad&arch=ppc64el&ver=0.4.0%2Bdfsg-1&stamp=1468713257

shirou:
perhaps, #225 is related. But as I said, I can not work because I don't have that arch.

andhe:
hello @shirou 

I've tried building latest git master on ppc64el and not quite there yet it seems (unless I've made a mistake which is very possible). It  fails on a testcase, here's a snippet from the build log:

=== RUN   TestCPUPercentIntervalZeroPerCPU
--- PASS: TestCPUPercentIntervalZeroPerCPU (0.01s)
FAIL
exit status 1
FAIL	github.com/shirou/gopsutil/cpu	0.291s
=== RUN   TestDisk_usage
--- PASS: TestDisk_usage (0.00s)



Which later ends with:

[...]

Issue #225, opened 2016-06-30
compile failed under ppc64le


xiayuhai:
go version go1.7beta2 linux/ppc64le
go build github.com/shirou/gopsutil/process
master: commit 2728d81

> # github.com/shirou/gopsutil/process
> 
> ../../shirou/gopsutil/process/process_linux.go:559: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:560: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:581: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:582: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:583: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:584: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:585: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:586: undefined: PageSize
> ../../shirou/gopsutil/process/process_linux.go:723: undefined: ClockTicks
> ../../shirou/gopsutil/process/process_linux.go:724: undefined: ClockTicks
> ../../shirou/gopsutil/process/process_linux.go:724: too many errors

shirou:
related #194. Something is not worked. But I can not try because I don't have a ppc64le machine.

PR #261, opened 2016-09-19
process: make `ClockTicks` arch-independent

### LINKED ISSUE #260: process: move `ClockTicks` to `process_linux.go`
From what I can see, the Linux kernel uses a fixed value (100) for the clock ticks on all architectures currently supported by Go, and so do all the `process_linux_*.go` files here.

Therefore, I propose we move the constant `ClockTicks` to `process_linux.go`. This should then also fix the remaining compile errors on ppc64le (#230 and #225).

I'll gladly write a patch for this, provided there are no objections.

monstermunchkin:
The value for `ClockTicks` is defined as `100` by the Linux kernel for
all currently supported architectures in Go. Therefore, there is no need
to define this constant for each architecture separately.

This fixes #260.

Signed-off-by: Thomas Hipp <thomashipp@gmail.com>

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -26,7 +26,8 @@ var (
 )
 
 const (
-	PrioProcess = 0 // linux/resource.h
+	PrioProcess = 0   // linux/resource.h
+	ClockTicks  = 100 // C.sysconf(C._SC_CLK_TCK)
 )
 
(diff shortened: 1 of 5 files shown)


shirou:
Thank you very much!

This potentially solves #230 and #225.

PR #263, opened 2016-09-21
added weightedIO field to disk IOCountersStat (last field of /etc/mtab)

DigitalMarc:
The field is described as "weighted time spent doing I/Os (ms)"
It enables mimic iostat avgqu-sz


lfittl:
Just as a supporting note, I'd also find it useful to have this field available :)

shirou:
Thank you!

DigitalMarc:
You are welcome!

PR #264, opened 2016-09-26
cpu.Percent() on Windows is very slow.

hillu:
Each call to `cpu.Percent()` on Windows takes about one second (measured on Windows XP running inside a VirtualBox VM), because the WMI queries just take this long. The call otherwise does not work as documented: `percpu` is not supported, setting  `interval` change anything.

These commits throw away the Windows-specific implementation in favor of calculating values based upon what the lower-level `Times()` call provides.

Please consider this PR as a request for comments rather than production quality. I have only quickly tested on the same XP VM mentioned above that beyond better latency, the measurements still seem to make sense.


shirou:
Excellent! Tests get faster from 11sec to 0.01sec, very cool. I can not confirm the value is correct since the lack of my experience about Windows, but definitely it works.

However, could you do me a favor? Your PR returns `ErrNotImplementedError ` if `percpu` is true. But the original code is returns not a error, just one result. This change may break someones code. So, could you change your PR to returns one result even if percpu is true?

Thank you!

hillu:
Sure, will do.

hillu:
@shirou Do you think that the PR is in mergeable shape now.?

shirou:
Sorry, I have not noticed your update. It is very good to merge. Thank you for your contribution!

PR #265, opened 2016-10-06
add Uids field to net_linux

jkahrs:


jkahrs:
while looking at it i'm not sure if i'm happy about cluttering this up so much :/

shirou:
Thank you so much!
But I feel copy everything from process_linux is a little bit messy. Could you copy (or write)  only the uid related things?

jkahrs:
Thank you for your feedback @shirou.
I'll look into this a little bit later this week. I think I'll be able to clean out some unnessary code.

shirou:

PR #267, opened 2016-10-11
Added Terminate() for Windows

botherder:
Since it appears to not have been implemented yet, I added support for `Terminate()` in *process_windows.go*.

As a side note, perhaps it would be worth adding PROCESS related constants (like `PROCESS_TERMINATE = 0x0001` to [shirou/w32](https://github.com/shirou/w32).

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -308,9 +308,20 @@ func (p *Process) Suspend() error {
 func (p *Process) Resume() error {
 	return common.ErrNotImplementedError
 }
+
 func (p *Process) Terminate() error {
-	return common.ErrNotImplementedError
+	// PROCESS_TERMINATE = 0x0001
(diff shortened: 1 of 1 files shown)


shirou:
Wow, great! Thank you so much!

> As a side note, perhaps it would be worth adding PROCESS related constants (like PROCESS_TERMINATE = 0x0001 to shirou/w32.

Indeed. I will add constants to w32 and edit that part of your code.

botherder:
Thank you.
If you don't mind, I might have some other expansions to make to Windows support.
For example, now I'm sorting it out ok with looping through Pids, but I would like to add process_iter support.

shirou:
Thank you for your contribution. PR is always welcome.  

`process_iter` is good for python. But for golang, I think it should use goroutine and a channel. I worry about this becomes complex. If you will implement something, could you show me the code? If my mind is a pointless concern, I will merge it. Thank you.

PR #268, opened 2016-10-13
Add ConnectionsMax function that limits connections per pid.

conorbranagan:
The motivation is to improve performance of fetching connections across
all processes when some processes can have several hundred or thousands of file
descriptors. Right now when you have many thousands of fds the process spends
lots of time inside the syscalls from Readdir and Readlink.

The public API works as before with two new functions:
- `ConnectionsMax`
- `ConnectionsPidMax`

Each function takes an additional int argument that sets the max number of fds
read per process.

This does add some duplication between `ConnectionsMax` and `Connections`
so I'm open to suggestions on a better approach.


shirou:
Oh, Great. Perhaps, it will help for other platforms. Could you add empty (this means just returns `common.ErrNotImplementedError`) to all other platforms?

shirou:
Thank you so much!

PR #269, opened 2016-10-14
Allow clock field in /proc/cpuinfo as cpu MHz fallback value

andhe:
Needed on ppc64le debian porter boxes atleast.

See #230 

This avoids getting MHz from both sysfs cpufreq and /proc/cpuinfo failing. Previously an error was thrown. With this patch we have MHz value in fallback.

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -152,9 +152,9 @@ func Info() ([]InfoStat, error) {
 				return ret, err
 			}
 			c.Stepping = int32(t)
-		case "cpu MHz":
+		case "cpu MHz", "clock":
 			// treat this as the fallback value, thus we ignore error
-			if t, err := strconv.ParseFloat(value, 64); err == nil {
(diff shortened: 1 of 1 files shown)


cpu/cpu_linux.go
```diff
@@ -152,9 +152,9 @@ func Info() ([]InfoStat, error) {
 				return ret, err
 			}
 			c.Stepping = int32(t)
-		case "cpu MHz":
+		case "cpu MHz", "clock":
 			// treat this as the fallback value, thus we ignore error
-			if t, err := strconv.ParseFloat(value, 64); err == nil {
+			if t, err := strconv.ParseFloat(strings.Replace(value, "MHz", "", 1), 64); err == nil {
```
dvusboy: What happens if the unit is not "MHz", but, say, "kHz" or "GHz"? Since I don't have any experience with PPC/Power architecture, I don't know. Of course, "cpu kHz" can be a problem too, but I have not seen that.

shirou:
wonderful!  Thank you for joining discussion and contribution!

PR #271, opened 2016-10-21
Obtain executable path from macOS process (fix #270)

### LINKED ISSUE #270: Executable path for macOS processes
I find myself currently in need of retrieving the executable path for processes on macOS.
As it currently stands, that is not implemented in gopsutil.

```
func (p *Process) Exe() (string, error) {
        return "", common.ErrNotImplementedError
}
```

Has anybody any directions as to how would be the best way to implement support for it?
One of the references I could find is libproc:
https://astojanov.wordpress.com/2011/11/16/mac-os-x-resolve-absolute-path-using-process-pid/

Thoughts?

**Update**: I'm attempting a work around through `Cmdline()`, but it is tricky, given that the command line returned by ps doesn't escape spaces and the slices can't really distinguish between words that are part of the executable path or of the following arguments.

One not very decent way is to use something like this command:

`lsof -p PID -Fn | awk 'NR==3{print}' | sed "s/n\//\//"`

botherder:
This is in relation to the ticket #270.
It might appear a bit hacky, but it is the most reliable way I have found to obtain the executable path for macOS processes. It finds the appropriate line from open handles.

Additionally, I had to add a function in common to introduce piping of shell commands.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -343,7 +343,7 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 }
 
 // https://gist.github.com/kylelemons/1525278
-func Pipeline(cmds ...*exec.Cmd) (pipeLineOutput, collectedStandardError []byte, pipeLineError os.Error) {
+func Pipeline(cmds ...*exec.Cmd) ([]byte, []byte, error) {
         // Require at least one command
         if len(cmds) < 1 { 
                 return nil, nil, nil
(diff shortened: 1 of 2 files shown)


shirou:
Thank you so much!

I think invoking external processes is a little bit heavy. So, could you change `sed` and `awk` in the PR to filter by go code? I haven't make a benchmark, but I think it faster and more readable.

lucab:
darwin seems to have a C native function for this, `_NSGetExecutablePath()`, see https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/dyld.3.html. You still need to follow through symlinks (hard&soft) in case of multicall binaries. We are also interested in having this directly in gopsutil, instead of wrapping it ourself in rkt.

shirou:
We already introduce C native function to `cpu` and `mem` with `cpu_darwin_nocgo.go` fallback mechanism. So as @lucab said, use `_NSGetExecutablePath()` for cgo and use ps and others for nocgo is reasonable, I think.

shirou:
Sorry for too late response. I confirm it works on my environment. Thank you for your contribution!

PR #274, opened 2016-10-24
Improve CPU identification for POWER processors

leitao:
Currently gopsutils fails to indentify the POWER processors family,
returning an almost empty Info() structure.

This patch improves the POWER identification without changing what is
available for x86.

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -144,10 +144,22 @@ func Info() ([]InfoStat, error) {
 			c.Family = value
 		case "model":
 			c.Model = value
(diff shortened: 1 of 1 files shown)


shirou:
Thank you so much!
I can not confirm it works, but it must be great!

lucab:
As a downstream consumer of gopsutil (in rkt), I'd like to have a tagged version that we can vendor and point distributions to, getting all recent non-x86 fixes (without the need of backporting or pinning to arbitrary commits). Can you kindly consider pushing a tag, if current master is stable enough?

PR #287, opened 2016-12-19
Add Slab, Shared and PageTables metric for VirtualMemory under linux

hush-hush:


shirou:
Thank you. I will merge this PR.

But it seems linux memory info becomes too big, I think. Some of metrics might become deprecated in the future.

PR #289, opened 2016-12-20
Cache the boot time after first query.

conorbranagan:
When fetching stats on all processes at once there's a non-trivial amount of
time spent in the `BootTime` call that we run for every process. But since this
value should never change once a host is up, we can cache it after the first call.

I think this assumption is correct but let me know otherwise.


shirou:
Indeed. Thank you!

PR #290, opened 2016-12-20
Fast duplication check in inodes processing.

conorbranagan:
Instead of encoding a JSON string of each connection (non-trivial at high
connection volumes) we can use the connTmp struct for map look-ups if we
eliminate the unused `uids` field.

Also switches to using the empty struct instead of bool for zero memory
overhead.


shirou:
Oh, exactly. Thank you for your contribution!

PR #291, opened 2016-12-21
Add 'SwapCached' to the VirtualMemoryStat informations on linux

hush-hush:


shirou:
Thanks! However as I mention at #287, I think metrics are too much. These values might become deprecated in the future release. (If so, major version will be changed.)

PR #293, opened 2016-12-23
fix: division by zero possible

HeChuanXUPT:
some filesystem have zero Inode


shirou:
Wow, I haven't hear such a file system.  Thank you for your contribution!

PR #296, opened 2017-01-10
Implemented resource limits function

smith3v:
Added limit constants. Implemented function to parse process limits from /proc.
Please let me know if you have any notes.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -60,6 +60,25 @@ type NumCtxSwitchesStat struct {
 	Involuntary int64 `json:"involuntary"`
 }
 
+const (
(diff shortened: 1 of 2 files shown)


process/process.go
```diff
@@ -60,6 +60,25 @@ type NumCtxSwitchesStat struct {
 	Involuntary int64 `json:"involuntary"`
 }
```
shirou: Could you add a comment about where these constants comes from? for example  `sys/resource.h in Linux`  or appropriate url.

smith3v: Done.

shirou:

Issue #300, opened 2017-01-16
process.Name() returns truncated process name


jcaberio:
Python version of psutil had this  [bug before](https://github.com/giampaolo/psutil/issues/628) but it's already fixed. Below are the python and golang versions of the test code. Notice that 'netcast-dashboard' was displayed correctly on python but truncated on golang psutil.

```
# pypsutil.py
import psutil

for proc in psutil.process_iter():
    try:
        pinfo = proc.as_dict(attrs=['pid', 'name'])
    except psutil.NoSuchProcess:
        pass
    else:
        print(pinfo)
```

```
$ python pypsutil.py
{'pid': 1, 'name': 'init'}
{'pid': 2, 'name': 'kthreadd'}
{'pid': 3, 'name': 'migration/0'}
{'pid': 4, 'name': 'ksoftirqd/0'}
{'pid': 5, 'name': 'migration/0'}
{'pid': 6, 'name': 'watchdog/0'}
{'pid': 7, 'name': 'migration/1'}
{'pid': 8, 'name': 'migration/1'}
{'pid': 9, 'name': 'ksoftirqd/1'}
{'pid': 10, 'name': 'watchdog/1'}
{'pid': 11, 'name': 'events/0'}
{'pid': 12, 'name': 'events/1'}
{'pid': 13, 'name': 'cgroup'}
{'pid': 14, 'name': 'khelper'}
{'pid': 15, 'name': 'netns'}
{'pid': 16, 'name': 'async/mgr'}
{'pid': 17, 'name': 'pm'}
{'pid': 18, 'name': 'sync_supers'}
{'pid': 19, 'name': 'bdi-default'}
{'pid': 20, 'name': 'kintegrityd/0'}
{'pid': 21, 'name': 'kintegrityd/1'}
{'pid': 22, 'name': 'kblockd/0'}
{'pid': 23, 'name': 'kblockd/1'}
{'pid': 24, 'name': 'kacpid'}
{'pid': 25, 'name': 'kacpi_notify'}
{'pid': 26, 'name': 'kacpi_hotplug'}
{'pid': 27, 'name': 'ata_aux'}
{'pid': 28, 'name': 'ata_sff/0'}
{'pid': 29, 'name': 'ata_sff/1'}
{'pid': 30, 'name': 'ksuspend_usbd'}
{'pid': 31, 'name': 'khubd'}
{'pid': 32, 'name': 'kseriod'}
{'pid': 33, 'name': 'md/0'}
[...]

PR #303, opened 2017-01-17
fix truncated process name of linux

jcaberio:
fix for issue #300


shirou:
I confirm `upstart-udev-bridge` and `upstart-socket-bridge` was truncated and fixed with your PR. Thank you for your awesome contribution!

PR #306, opened 2017-01-27
add bson struct tags

tychoish:


shirou:
Could you tell me why bson tag is required for this library?

tychoish:
It makes it easier for me to serialize this data into bson documents to cache the data in mongodb, which is a thing I want to do, it's not required, certainly, but it would be helpful.

lfittl:
As there is tags for other databases as well (and they are not present), I find it a bit odd that this library would have tags for BSON/MongoDB specifically.

I wonder if you could achieve the same benefit by asserting to a struct type you define that has the BSON tags, with the new type conversion rules re: tags in Go 1.8? (https://tip.golang.org/doc/go1.8#language)

tychoish:
I believe that it'll be a while before we're able to migrate to Go1.8, entirely, but I'm currently working around this by applying this change during our vendoring operation. 

I think requiring users to maintain parallel sets structs for all gopsutil stucts is a bit onerous, and would support adding tags for more packages to simplify use of these types.

shirou:
I agree with @lfittl. `json` is widely used but `bson` is not and too application specific. gopsutil aims to become an library used by various kind of application.

shirou:
I close this issue. Thank you for your suggestion.

PR #308, opened 2017-02-01
Add Go-native structs for Darwin/386 and FreeBSD/arm

sean-:


shirou:
Cool! Thank you!!
Unfortunately, I can not confirm it works but I will merge it.

PR #309, opened 2017-02-01
Add native struct support for FreeBSD/arm.

sean-:
This was ported by hand and not machine generated, but appears to get
software to build.  Material was sourced from the following and a pile
of `grep(1)`:

[1] https://wiki.freebsd.org/EdMaste/ArchitectureSpecifics
[2] https://svnweb.freebsd.org/base/head/sys/sys/user.h?revision=309676&view=markup#l119

### DIFF
--- a/process/process_freebsd_arm.go
+++ b/process/process_freebsd_arm.go
@@ -1,0 +1,192 @@
+// Created by cgo -godefs - DO NOT EDIT
+// cgo -godefs types_freebsd.go
+
+package process
+
+const (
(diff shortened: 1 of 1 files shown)


sean-:
With this PR, we can now build Consul on FreeBSD/arm without cgo.

shirou:
I am very thank you for this kind of contribution. I feel regret to confirm that I can not confirm it.

sean-:
Likewise.  Regardless, thank you!

PR #311, opened 2017-02-01
fix panic when /proc/stat is empty

sparrc:
I don't really know why this would be the case, but I suppose there are
always edge-cases.

see https://github.com/influxdata/telegraf/issues/2356


shirou:
hmm, it's strange. But it could be occurred from the code. Thank you!

PR #312, opened 2017-02-03
Add HostID support to Windows.

sean-:
Use `HKLM/SOFTWARE\Microsoft\Cryptography`'s `MachineGuid` registry key to enable `HostID` support on Windows.

### DIFF
--- a/host/host_windows.go
+++ b/host/host_windows.go
@@ -7,7 +7,9 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/StackExchange/wmi"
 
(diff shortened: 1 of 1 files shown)


shirou:
Thank you so much! It works perfect on my Windows.

sean-:
Excellent, thank you for telling me that!  FWIW, there are reports that `x/sys/windows/registry` works conditionally based on Win32 vs Win64 (https://groups.google.com/forum/#!topic/golang-nuts/ZmWaTdv_z7U), but the above works 100% of the time on both platforms, which is why I used `syscall.RegOpenKeyEx()` vs [`registry.OpenKey()`](https://github.com/golang/sys/blob/master/windows/registry/key.go#L11-L21).  FYI in case someone is curious in the future why `registry.OpenKey()` wasn't used.

shirou:
Wow, great information! Your comment helps someone definitely.

PR #313, opened 2017-02-07
Unconditionally lowercase the `HostID` from all supported platforms.

sean-:
This was an oversight from earlier.  All of the platforms I use toss out lowercase UUIDs, but Linux does not.

### DIFF
--- a/host/host_darwin.go
+++ b/host/host_darwin.go
@@ -59,7 +59,7 @@ func Info() (*InfoStat, error) {
 
 	values, err := common.DoSysctrl("kern.uuid")
 	if err == nil && len(values) == 1 && values[0] != "" {
-		ret.HostID = values[0]
+		ret.HostID = strings.ToLower(values[0])
 	}
 
 	return ret, nil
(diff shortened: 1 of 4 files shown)


shirou:
`HostId` is platform dependent, therefore I left it. But I thought about it with your PR, it should be consistent. So, I merge this PR. Thank you for your contribution!

sean-:
Agreed with your points and thinking.

`HostID` is platform dependent, but nearly all platforms emit a hex-encoded UUID (at least, all of the platforms that `gopsutil` supports emit a hex-encoded UUID).  Because it is a hex-encoded UUID, it should be case insensitive.  We ran into a situation where we were doing a `string` comparison of an existing, unmodified UUID that happened to be already stored as uppercase, and a lowercase version of the UUID.  Both UUIDs values were from the same server, however one of them had received a `strings.ToLower()` in the application that was using `gopsutil`, the other had not.  This led to the thinking that the correct thing to happen should be that `gopsutil` unconditionally return all of its `HostID`s as the same case (i.e. `strings.ToLower()`.

Thank you for merging!

PR #314, opened 2017-02-14
error earlier on darwin for get parent pid

tychoish:
I've been seeing a panic caused by this issue.


shirou:
Indeed. Thank you!

PR #316, opened 2017-02-15
fix panic in starvation situations

tychoish:
line 208 was panic'ing because fields could be 0 len, if we had trouble reading from proc, this just makes this safe.


shirou:
Exactly! Thank you!

PR #319, opened 2017-02-22
Fixes memory leak

toshipiazza:
In net_linux.go we open all files in /proc but we never close them. This causes a psutil golang app to crash on a busy system after 3 or 4 calls to `net.Connections()`.


shirou:
Wow!! Really thank you!

PR #320, opened 2017-02-22
close open files

tychoish:
after seeing #319, I did a little audit, and found these changes which I think will be generally useful.


shirou:
Oh my... I really thank you for this PR.

Issue #322, opened 2017-03-03
getMachineGuid() function return a empty value in x86 windows builds


molecul:
getMachineGuid() function return a empty value in x86 windows builds.

Test file: https://gist.github.com/molecul/550c9655f5006a37c06227b44b421f9a

Results
x86: http://pastebin.com/qYiUNJcC
amd64: http://pastebin.com/RcRF7TBW

shirou:
close via #323 and #324. Thank a lot!

PR #323, opened 2017-03-03
Fix for getMachineGuid() return value

molecul:
getMachineGuid() function return a empty value in x86 windows builds.
Issue: https://github.com/shirou/gopsutil/issues/322

### DIFF
--- a/host/host_windows.go
+++ b/host/host_windows.go
@@ -80,7 +80,7 @@ func Info() (*InfoStat, error) {
 
 func getMachineGuid() (string, error) {
 	var h syscall.Handle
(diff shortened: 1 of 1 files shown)


shirou:
Thank you. I am not familiar with Windows, but I am wander about `KEY_WOW64_32KEY` should be used on 32 bit OS, or not?
https://msdn.microsoft.com/library/aa384129(v=vs.85).aspx

molecul:
According with msdn, syscall.KEY_WOW64_64KEY open a "window" to 32-bit registry on x86 systems.

molecul:
It's mean that this patch should be work correctly on x86/amd64 platforms

shirou:
It works at least on my Windows7 64bit environment, and it may also work 32bit according to your comment and MSDN. 

PR #325, opened 2017-03-08
cpu: Expose CPU microcode information

simar7:
This commit exposes the CPU microcode information.

Signed-off-by: Simarpreet Singh <simar@opendns.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -133,6 +133,7 @@ Several methods have been added which are not present in psutil, but will provid
   - Mhz
   - CacheSize
   - Flags        (ex: "fpu vme de pse tsc msr pae mce cx8 ...")
+  - Microcode
 
 - load/LoadAvg()  (linux, freebsd)
 
@@ -262,6 +263,7 @@ hostname              x     x      x       x       x
   CoreID              x
   Cores               x                            x
   ModelName           x     x      x       x       x
+  Microcode           x
 **LoadAvg**
   Load1               x     x      x       x
   Load5               x     x      x       x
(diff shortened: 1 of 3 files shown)


shirou:
Thank you for your PR!

This patch can get microcode as hex such as `0x1` , because `/proc/cpuinfo` includes this line.
```
microcode       : 0x1
```

However, on my darwin, run `/usr/sbin/sysctl machdep.cpu` and get microcode as decimal.
```
machdep.cpu.microcode_version: 23
```

Do you know hex or decimal, which is prefer? I want to have consistent between platforms.

simar7:
Hey @shirou good question. To the best of my knowledge, I have run into various possible values for microcodes, they're usually hex but I've also seen on linux systems where microcode values were plain strings (no hex, e.g. "0001..")

That's why I've kept it in a string format to keep it compatible across platforms.

shirou:
So, how about store as int32? `strconv.ParseInt` with `base==0` can handle `0x01`, `0001`, and `1`.

simar7:
@shirou Yeah I agree that's possible but it then wouldn't make much sense for an hex value to be exposed as an int when reading. For an example see this go playground: https://play.golang.org/p/RC_oIkzC_N

There will be cases that someone will use the value of the microcode "as-is", which means they will expect it in either hex (linux) or an int32 (darwin) but it wouldn't make sense at that point if it was exposed using the gopsutil lib as an int always.

Does that make sense? Happy to answer follow up questions if you have any!

shirou:
Hmm. I make sense microcode should be *as-is*. However, I am thinking about if a end-user who uses some tool written with gopsutil check microcode revisions between various kind of OS/Distributions.
If microcode is *as-is*, that user has to switch that distributions and OSes, such as

```
switch host.InfoStat.OS {
  case "linux":
        (cast from hex)
  case "darwin":
        (cast from int)
}
```

... let me see ...

OK. I have changed my mind at the point where I wrote so far.  This `switch` is not difficult and it is very important to pass a raw information to the end-user.

So, I merge this PR. Thank you very much about your cool contribution!

PR #327, opened 2017-03-08
disk: linux: add label based on devicemapper name of disk

### LINKED ISSUE #326: disk: device mapper name in linux
I would like to support device mapper naming in linux for the disk module.

It's not as exotic as it sounds, eg LVM devices get a mapper name; but my main interest is is multipath devices.

to get the name, there's following
https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-block-dm

so given device name `abc` it is as easy as trying to get the content of `/sys/block/abc/dm/name`.
it will be empty for non-dm devices and for older kernels or system without /sys mounted

my main question is: would such feature get merged? and if so, what attributename can i use for that?

i would implement it similar to `getSerialNumber` and maybe add it in `IOCounter` as `d.DMName = getDeviceMapperName()`, similar to the way the serial number is added.

stdweird:
Fixes #326

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -46,6 +46,7 @@ type IOCountersStat struct {
 	WeightedIO       uint64 `json:"weightedIO"`
 	Name             string `json:"name"`
 	SerialNumber     string `json:"serialNumber"`
+	Label            string `json:"label"`
 }
 
 func (d UsageStat) String() string {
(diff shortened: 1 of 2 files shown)


ofek:
@stdweird Can you please rebase this?

ofek:
Yes it can

shirou:
Thank you!

PR #328, opened 2017-03-12
cpu: Implement Solaris support for Info()

jen20:
This commit adds Solaris<sup>1</sup> support for the Info() function of package cpu, with the exception of the L2 cache size which is not trivially available<sup>2</sup>.

Support is implemented by parsing the output of `isainfo -b -v` (for the instruction set flags), and `psr-info -p -v` for other information. Example outputs from a range of different size VMs and Joyent containers are included as part of the tests.

`Times()` is still unimplemented, and will follow in another pull request.

1: This has only been tested with Illumos in the form of SmartOS. I believe it to be portable to other Illumos distributions but have not tested on Oracle Solaris.

2: Enough support is added here to be usable for my port of HashiCorp's Nomad to SmartOS.


sean-:
Excellent @jen20, thank you!!!!

@shirou , I gave this a few pokes offline and it looks good to me.

shirou:
awsome! It can get CPU information on my OpenIndiana environment.

However, the `TestParseProcessorInfo` is failed.

I think `tc.expected` is `[]*cpu.InfoStat` and `cpus` is `[]cpu.InfoStat`, so DeepEqual is failed. Here is what I run `fmt.Printf("%#v\n", tc.expected)`

```
=== RUN   TestParseProcessorInfo
[]*cpu.InfoStat{(*cpu.InfoStat)(0xc42001c320)}
[]cpu.InfoStat{cpu.InfoStat{CPU:0, VendorID:"GenuineIntel", Family:"6", Model:"78", Stepping:3, PhysicalID:"0", CoreID:"0", Cores:1, ModelName:"Intel(r) Core(tm) i7-6567U CPU @ 3.30GHz", Mhz:3312, CacheSize:0, Flags:[]string(nil)}}
```

So, could you fix this test? 
1. return `[]cpu.InfoStat`
2. `json.Marshal` format to compare
3. use [`assert.JSON.eq`](https://godoc.org/github.com/stretchr/testify/assert#JSONEq)
4. or anything else.

Again, thank you great contribution!

jen20:
@shirou Yup, good catch on the test. It's passing now.

shirou:
Thank you so much!

PR #331, opened 2017-03-14
Host: Add Solaris support for Info(), BootTime(), Uptime()

jen20:
This commit adds support for Info(), BootTime() and Uptime() in package Host. It uses no cgo, preferring to parse the output of `kstat -p` instead.

Thanks go to @gfrey for the parsing logic for `/etc/release` and `uname`. I have only tested this on SmartOS however.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -248,12 +248,12 @@ Original Metrics
 ================== ===== ======= ======= ====== ======= =======
 item               Linux FreeBSD OpenBSD MacOSX Windows Solaris
 **HostInfo**
-hostname              x     x      x       x       x
-  uptime              x     x      x       x
-  proces              x     x      x
-  os                  x     x      x       x       x
-  platform            x     x      x       x
-  platformfamily      x     x      x       x
+hostname              x     x      x       x       x       x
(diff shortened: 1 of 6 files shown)


sean-:
Also 👍 , @shirou .

jen20:
I have extended this PR to also add `Total` support for `VirtualMemory`. This works correctly in the global zone as well as in non-global zones with a memory cap.

shirou:
Awesome!!

`cpu.Info()` works like a charm on my OpenIndiana environment. However, `cpu.VirtualMemory` test failed because it only contain Total Memory. Do you know how to get other value such as `Used` ?

or if there are nothing to get, this test can be disabled when `runtime.GOOS == 'soralis'`

jen20:
@shirou Most of the values can be obtained through `kstat(1M)` (without cgo) or the `kstat` library functions (with cgo), but I don't have time to do so right now - I've just implemented what Nomad needs. I'll disable all the failing tests for now and push another commit to this branch.

shirou:
Very quick action. Thank you !!

PR #333, opened 2017-03-16
Added skeleton for a package to retrieve process info from files

botherder:
Similarly to what the net package does.
This clearly needs to be extended and integrated appropriately as I'm not entirely sure what's the procedure for the right cross-platform implementation.


shirou:
I'm sorry, but could you explain this PR purpose?

botherder:
Have a function to retrieve pids of processes that access a given file.

Issue #329, opened 2017-03-13
Adding temperature measures


dotpy3:
This is a proposition for an additional feature for gopsutil.

I've noticed **CPU temperature measures** were not available in gopsutil. I need such measures in one of my projects for macOS and Linux where I use gopsutil, so I started looking into a possible implementation of this.

## Linux

The python psutil exposes a [`sensors_temperatures()`](http://pythonhosted.org/psutil/#sensors) function, that returns the current temperatures of multiple sensors. It is, however, only available for Linux for the moment. The way it is implemented is detailed [in the source code](https://github.com/giampaolo/psutil/blob/1922a7674011852b64f3405e47113ba540a86c77/psutil/_pslinux.py): they use the `/sys/class/hwmon` interface. The drawback with this implementation is that it isn't explicitly indicated which sensors measure CPU temperature - which means it is up to the user to determinate it.

## macOS

Even though their implementation doesn't support macOS, I've started to look if there were pieces of software that allowed the same thing. There is a Ruby program, called [iStats](https://github.com/Chris911/iStats), that shows system stats, including CPU temperature:

```
$ iStats
--- CPU Stats ---
CPU temp: 55.0°C ▁▂▃▅▆▇

--- Fan Stats ---
Total fans in system: 1
Fan 0 speed: 0.0 RPM  ▁▂▃▅▆▇

[...]
```

The way it is implemented in the package is through a C library, [`smc.c`](https://github.com/Chris911/iStats/blob/master/ext/osx_stats/smc.c), that acts as an interface to the Apple System Management Control tool (that has such information).

## Implementation proposition

Given that we would like a platform-agnostic implementation, what we could do is provide a global `get_sensors_temperatures()` function, that would:
+ In the case of a Linux machine, return all the sensors' temperatures as a list
+ In the case of a macOS, return only the CPU's temperature, but as a list, to match the types of the Linux implementation (unless there are other temperature sensors that could be interesting to expose?).

Information retrieval would be trivial on Linux, considering the source code of the python version of `psutil`, but the macOS implementation would either require to import `smc.c` through Cgo, or using a full-Go version. It seems there is a [Go port of smc](https://github.com/hnw/go-smc), that is built with Cgo.

What do active contributors think? Do you think it could be worth it to implement this? If so, what approach should we take into implementing this?

shirou:
Thank you for suggestion. I agree CPU temperature is important information and since psutil supports it, gopsutil either.

## Mac OS

[...]

PR #334, opened 2017-03-19
Added temperatures sensors for macOS and Linux

dotpy3:
This pull request, following #329, introduces in the `host` package : 

+ A new structure, `TemperatureStat`, that stores a temperature key and its sensor value. This is to ensure a standard portable structure for temperature sensors data for the package:

```go
type TemperatureStat struct {
	SensorKey   string  `json:"sensorKey"`
	Temperature float64 `json:"sensorTemperature"`
}
```

+ A new function, `SensorsTemperatures()`, available for the moment on macOS and Linux builds. This function returns a list of `TemperatureStat` objects - each of these objects represent the value of a temperature sensor.

  + On Linux, the data is obtained the same way as psutil, through the `hwmon` interface. It is syntaxically correct and follows the psutil's procedure, but hasn't been tested properly yet - if a Linux-using contributor could test this is on their machine, I would be grateful for it.

  + On macOS, the data is obtained through the System Management Controller, with a C library acting as a middleman (`host/include/smc.{c,h}`). This C library is GNU-licensed, and the license has been kept as-is. Thus, the function is only available for Cgo-enabled builds. Match between a sensor key and what it represents is available in the `host/include/smc.h` file. The function returns 0 for a sensor value if there was an error on the SMC side or that the sensor is not available, which is the normal behaviour of the SMC function - but if we only want values that could have been accessed, we could ignored sensor values that are equal to 0.

This is my first contribution to the project, so if any contributor has any comment on this, please feel free to speak up!

### DIFF
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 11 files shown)


host/host_linux.go
```diff
@@ -521,3 +522,36 @@ func Virtualization() (string, string, error) {
 	}
 	return system, role, nil
 }
+
+func SensorsTemperatures() ([]TemperatureStat, error) {
+	var temperatures []TemperatureStat
+	files, err := filepath.Glob("/sys/class/hwmon/hwmon*/temp*_*")
```
shirou: Could you use `common.HostSys` to change `/sys` path via environment variable?

shirou:
Thank you! It works like a charm at least on my Darwin environment. awesome!

dotpy3:
Sure! I didn't add any unit tests for `SensorsTemperatures()` yet - only for the struct -, since otherwise the tests would fail on the platforms where temperatures are not yet available, but tell me if you want me to add some!

shirou:
I'm very sorry. I have not noticed your update. It seems good for me. Thank you for your contribution!

PR #335, opened 2017-03-20
cpu[linux]: ignore cpufreq/cpuinfo_max_freq error

### LINKED ISSUE #282: cpu.Info() returns error when clock speed not available
I'm getting the following error from `cpu.Info()`:

```
open /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq: no such file or directory
```

on an ARM device that has no clock speed entry in`/proc/cpuinfo`:

```
$ cat /proc/cpuinfo
Processor	: Marvell PJ4Bv7 Processor rev 2 (v7l)
processor	: 0
BogoMIPS	: 1332.01

processor	: 1
BogoMIPS	: 1332.01

Features	: swp half thumb fastmult vfp edsp vfpv3 tls
CPU implementer	: 0x56
CPU architecture: 7
CPU variant	: 0x2
CPU part	: 0x584
CPU revision	: 2

Hardware	: Marvell Armada XP Development Board
Revision	: 0000
Serial	: 0000000000000000
```

and no `cpufreq` node available under `/sys/.../cpu`:

```
$ ls /sys/devices/system/cpu/cpu0/
cpuidle online topology
```

In this case, I think it would be useful to return the known information about the CPUs, even if the Mhz field is set to 0.

shirou:
This fixes #282, just a ignore error and leave MHz 0.

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -83,22 +83,14 @@ func finishCPUInfo(c *InfoStat) error {
 	// of the value from /proc/cpuinfo because we want to report the maximum
 	// clock-speed of the CPU for c.Mhz, matching the behaviour of Windows
 	lines, err = common.ReadLines(sysCPUPath(c.CPU, "cpufreq/cpuinfo_max_freq"))
-	// if we encounter errors below but has a value from parsing /proc/cpuinfo
-	// then we ignore the error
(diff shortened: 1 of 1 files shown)


tmm1:
:+1:

PR #338, opened 2017-03-24
Add hostinfo Virtualization vmware

itnihao:


shirou:
Sorry for late response. Thank you!

PR #342, opened 2017-03-31
clarify return value of cpu.Percent()

jgillich:
I was quite confused by what it would return, so I added a note.


shirou:
Indeed. Thanks a lot!

PR #351, opened 2017-04-14
Make IOCounters a variadic function

danielnelson:
Remove IOCountersForNames that was recently added in favor of variadic
function with same capabilities.


shirou:
Excellent! Thank you!

PR #354, opened 2017-04-27
Improve Command() handling and signal handling for launched processes.

sean-:
Use Go's `context` package for command timeouts.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -9,10 +9,10 @@ package common
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


shirou:
Thank you! It is exactly what I want to do. But `context` is a golang 1.8 feature so break some compatibility, could you wait merge until next patch release? gopsutil has monthly release policy, this means April 30.

sean-:
... yes, I can wait until Monday to have this merged.  :)  LMK if you have any other concerns.  Because this compatibility would be broken along API boundaries due to the removed symbols (I couldn't find any references to these exported, but unused symbols), it didn't seem problematic to merge.  The dependency on `context` is a new dependency, but seems improbable that someone would suddenly need to compile current `gopsutils` without `go1.8` available to them (static compilation of binaries ftw).

shirou:
Thank you. `WaitTimeout` and `CombinedOutputTimeout` are inside `internal`, so other applications can not use these. And I will write README to note new golang 1.8 dependency after merge.

PR #356, opened 2017-04-28
Optimize memory usage for net.Connections on Linux

ablagoev:
Hey,

In servers with higher tcp connection counts the /proc/net/tcp file and the likes get big. For example, one one of our servers /proc/net/tcp is 9 MBs. Parsing it becomes expensive.

As such the impact of keeping another copy of the connTemp structure just for indexing and filtering out duplicates is considerable. My PR proposes indexing on the socket type, src ip, src port, dst ip and dst port. The combination of these should be enough to ID the connection uniquely. This saves some memory because the other connTemp struct members are not used.

Regarding CPU utilization, on my local tests the modified version results in a little below 1% more CPU time.

I've run both mine and the original version through some test files from our production machines and both return the same results.

On average in our setup this change saves 30% of memory. We are using the telegraf input plugin, which runs netstat in 1 minute intervals.

### DIFF
--- a/net/net_linux.go
+++ b/net/net_linux.go
@@ -346,12 +346,13 @@ func ConnectionsPidMax(kind string, pid int32, max int) ([]ConnectionStat, error
 }
 
 func statsFromInodes(root string, pid int32, tmap []netConnectionKindType, inodes map[string][]inodeMap) ([]ConnectionStat, error) {
-	dupCheckMap := make(map[connTmp]struct{})
+	dupCheckMap := make(map[string]struct{})
 	var ret []ConnectionStat
 
 	var err error
 	for _, t := range tmap {
 		var path string
+		var connKey string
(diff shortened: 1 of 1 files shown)


shirou:
Wow, great contribution!

I will merge your PR, but can I change to use fmt.Sprintf() in [here](https://github.com/shirou/gopsutil/pull/356/files#diff-43bc4667c124b4ebde89cb672ed882fcR373)? or, have you already benchmarked between Sprintf? I understand Sprintf is a high cost but easier to read the code.

shirou:
I appreciate your contribution very much!

ablagoev:
Hey,

You correct, it is much better using Sptrintf. In terms of performance I don't think there is a measurable impact. On my local benchmarks there is barely a difference between the two.

Thanks for all of your work in this project too!

PR #358, opened 2017-04-30
Fromat socketType properly in tcp duplication check key

ablagoev:
Hey,

I am sorry for the spam and the second PR. Upon further testing I noticed that I wasn't formatting the c.sockType properly, it is a uint32 and I formatted it as a string.


shirou:
Ooops. Sorry, I have not noticed. This is my fault. Thank you.

PR #365, opened 2017-05-08
Include "DRIVE_REMOTE" volumes on windows

tmm1:
Unix systems show remote mounts as there is no distinction on those platforms, so it makes sense to include on windows as well.


shirou:
Hmm, it is reasonable. Thank you for your PR!

PR #366, opened 2017-05-08
Ignore GetVolumeInformation errors on DRIVE_REMOVABLE volumes

tmm1:
Prior to this, a missing removable drive at say `E:` would cause the function to return early and prevent it from including additional drives that appeared later (like say `J:`)


shirou:
Thank you!

Issue #78, opened 2015-09-17
Request for bogomips as part of cpuinfo.


selvik:
Thanks for the useful tool!  It would be great if bogomips could be added as part of the CPU stats.

shirou:
Bogomips is defined for only Linux. Since gopsutil is for the multi-platform, I can not add platform-specific field to the `CPUInfoStat`, sorry.

selvik:
Ahh...valid point, that makes sense! Thanks for replying.

shirou:
Thanks for your understanding. I close this issue.

PR #367, opened 2017-05-09
Extract bogomips for cpuinfo on linux

tmm1:
useful for some scenarios, seems harmless to add..


shirou:
Could you read #78 ?

If you have any idea about on other platforms, please reopen. Thank you.

PR #372, opened 2017-05-19
Read disk partitions on linux from /proc/mounts instead of /etc/mtab

### LINKED ISSUE #371: disk.Partitions fails on Android
`/etc/mtab` is not available on Android, however `/proc/mounts` is available.

tmm1:
fixes #371

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -214,8 +214,6 @@ var fsTypeMap = map[int64]string{
 // Partitions returns disk partitions. If all is false, returns
 // physical devices only (e.g. hard disks, cd-rom drives, USB keys)
 // and ignore all others (e.g. memory partitions such as /dev/shm)
-//
-// should use setmntent(3) but this implement use /etc/mtab file
 func Partitions(all bool) ([]PartitionStat, error) {
 	filename := common.HostProc("mounts")
 	lines, err := common.ReadLines(filename)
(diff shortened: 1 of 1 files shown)


tmm1:
I tested this on GOOS=android and it works now.

shirou:
After reading some articles, I understand you are correct. Thank you.
- https://linux.die.net/man/8/mount
- http://fibrevillage.com/sysadmin/278-understanding-the-difference-between-etc-mtab-and-proc-mounts-on-linux
- http://www.linuxquestions.org/questions/slackware-14/etc-mtab-vs-proc-self-mounts-371224/

But, `/proc/mounts` is an alias of `/proc/self/mounts` on my linux.

```
lrwxrwxrwx 1 root root 11  5月 19 12:06 /proc/mounts -> self/mounts
```

How about to change `common.HostProc("self/mounts")` ? 

And, if you can run on `GOOS=android`, could we add build tag to `disk_linux.go`? 

`// +build linux, android`

(I think this should not be included in this PR)

tmm1:
Sure, self/mounts works too.

Extra build flag is not required, because GOOS=android automatically sets Linux build flag already.

shirou:
So, could you change to `self/mounts` ? Then, I will merge this PR.

> Extra build flag is not required, because GOOS=android automatically sets Linux build flag already.

Great!

tmm1:
Done!

shirou:
merged. Thank you for your contribution!!

PR #375, opened 2017-05-24
Reduce unnecessary syscalls on common.NumProcs() in Linux

astj:
(in *nix, ) `os.File.Readdir()` calls `lstat` on files under the directory, but `os.File.Readdirnames()` does not call and simply returns its names.
In `common.NumProcs()`, only `len(list)` will be used as return value, so this `Readdir()` can be replaced by `Readdirnames()`, maybe.

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -35,7 +35,7 @@ func NumProcs() (uint64, error) {
 	}
 	defer f.Close()
 
-	list, err := f.Readdir(-1)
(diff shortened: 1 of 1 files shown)


shirou:
That’s a good point! Thank you!

(might be refs: https://github.com/mackerelio/mackerel-agent/issues/390)

astj:
Thank you!

(Yes, I noticed this while investigating the issue.)

Issue #373, opened 2017-05-23
Windows OS - Plan to include Performance Metrics for each core and processor


RajGupta-Continuum:
gopsutil provides overall cpu performance metrics for Windows OS using GetSystemTimes Win32 API.
In case of multiprocessor (and multicore) system, we want performance metrics for each core and processor. 
Can we have this capability?
I can add this capability and propose to use [Win32_PerfFormattedData_Counters_ProcessorInformation](http://library.wmifun.net/cimv2/win32_perfformatteddata_counters_processorinformation.html) WMI class.

shirou:
SInce I am not good at Windows, if you can make a PR for enhance, I will very appreciate.

(#266 might be related)

RajGupta-Continuum:
Thanks for prompt response, I will make a PR for the same.

shirou:
#376 is merged. Thank you so much!

PR #376, opened 2017-05-25
Windows: Read processor performance metrics overall, per cpu and per core

RajGupta-Continuum:
Pull request for the enhancement ([#373](https://github.com/shirou/gopsutil/issues/373)) to get the processor performance metric cpu and core wise specific to windows platform.

In addition to this added a function which provide number of processes and processor queue length.

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -23,6 +23,26 @@ type Win32_Processor struct {
 	MaxClockSpeed             uint32
 }
 
+// Win32_PerfFormattedData_Counters_ProcessorInformation stores instance value of the perf counters
+type Win32_PerfFormattedData_Counters_ProcessorInformation struct {
+	Name                  string
+	PercentDPCTime        uint64
+	PercentIdleTime       uint64
+	PercentUserTime       uint64
+	PercentProcessorTime  uint64
+	PercentInterruptTime  uint64
+	PercentPriorityTime   uint64
+	PercentPrivilegedTime uint64
+	InterruptsPerSec      uint32
+	ProcessorFrequency    uint32
+	DPCRate               uint32
+}
+
+type Win32_PerfFormattedData_PerfOS_System struct {
+	Processes            uint32
+	ProcessorQueueLength uint32
+}
+
 // TODO: Get percpu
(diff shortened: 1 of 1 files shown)


shirou:
Thank you for the PR. However, it seems this PR not complete yet. I think some change is needed to adapt current `cpu.Times` API which is about invoke `PerfInfo()` from `Times()` when `percpu` is true. If my thought is correct, I will change the code after merge this PR.

RajGupta-Continuum:
Yes, you are correct but we will not have one to one mapping for all TimesStat attributes collected from linux to windows. Only some of them are mapped. I found msdn blog on this, refer for more details.
https://blogs.msdn.microsoft.com/ericgolpe/2015/03/18/comparing-linuxunix-and-windows-performance-counters-on-microsoft-azure/

Please suggest and let me know if you want me to complete this PR.

shirou:
I understand one-to-one mapping is impossible. So, could you make some of the value? 

At my first glance, these values can be mapped.
```
Windows      cpu.go
User       -> User
Privileged -> system
Idle       -> Idle
Interrupt  -> Irq
```

And, I think there are not necessary to become strict same meaning between platforms. It is impossible. Important thing is become consistent in a platform, I think.

RajGupta-Continuum:
Thanks Shirou. You have mapped common attributes correctly.

Windows          cpu.go
User             -> User
Privileged    -> system
Idle              -> Idle
Interrupt      -> Irq
Name	    -> CPU [Possible values are "_Total", "iCPU,_Total", "iCPU,iCore" where iCPU is CPU index with values 0, 1.. so on and iCore is Core index with values 0, 1.. so on]

I agree with your thought that to be consistent with in a platform. Let me know if any work I can do on this.

shirou:
So, could you add `func perCPUTimes()([]TimesStat, error)` which returns a `[]TimesStat` according to that attribute map? Then, invoke from `Times()` like

```
func Times(percpu bool) ([]TimesStat, error) {
     if percpu {
          return perCPUTimes()
     }
   ...
```

RajGupta-Continuum:
I have added suggested changes, please review and do the needful. 
Thank you.

shirou:
It works on my Windows Environment. Great work! Thank you!!

Issue #30, opened 2014-12-19
use golang.org/x/sys instead of syscall


shirou:
From go 1.4, syscall is frozen and requested to use golang.org/x/sys
https://golang.org/doc/go1.4

Should move to sys package.

stevenh:
Be careful about that as its not maintained very well, with lots of missing fixes as I found out today :(

shirou:
Thank you for the information. syscall is just frozen, not deprecated. I will wait until it become more mature.

kardianos:
@shirou syscall is still actively used today and will continue. /x/sys is just where any exposed new functionality will be added. /x/sys augments, but will never replace, syscall. I recommend closing this issue as invalid.

shirou:
Thank you for the comment!
Hmm. yah, it seems `syscall` and `/x/sys` are both active as you said.

Currently, `/x/sys` new functionality is not required, so I close this issue temporally.
When `/x/sys` can replace syscall entirely or new function is required, I will reopen this issue.

shirou:
refs #380.

PR #378, opened 2017-06-01
Add CPU and Disk stats support to Illumos

sean-:
Add support for various Illumos stats (via `kstat(1)` atm).

CC @jen20


disk/disk_unix.go
```diff
@@ -2,11 +2,11 @@
 
 package disk
 
-import "syscall"
+import "golang.org/x/sys/unix"
 
 func Usage(path string) (*UsageStat, error) {
-	stat := syscall.Statfs_t{}
-	err := syscall.Statfs(path, &stat)
+	stat := unix.Statfs_t{}
```
shirou: I'm sorry, but could you tell me why use `unix` instead of `syscall`?
If use `unix`, at least darwin platform this error occured.

```
./disk_unix.go:17: cannot use stat (type unix.Statfs_t) as type syscall.Statfs_t in argument to getFsType
```

`getFsType` is this function 

```
func getFsType(stat syscall.Statfs_t) string {
        return common.IntToString(stat.Fstypename[:])
}
```

sean-: Drat, let me look into that.  I'm mid-way through cherrypicking some patches onto this branch that will add memory support.  I don't know why I didn't catch this during testing, however.

IMNSHO the `syscall` package should _not_ be used anywhere where an `x/sys/unix` package variant is available _because_ the `syscall` package is frozen, whereas the `x/sys/unix` package will track the current API.

In this particular case, there wasn't support for `syscall.Statfs` on the target platform, but was under the `unix.` package.  It has been lamented a few times by the Go authors that the `syscall` package should be removed.  IMO `syscall` would have been "fine" (on some level) if it weren't frozen.  But since `unix` is there and not frozen, `syscall` should be outright deprecated and eschewed from all code bases so that programs can converge on a single maintained package (read: `unix` > `syscall`).

For instance, Illumos grew support for `unix. Statvfs` which didn't exist previously and couldn't have existed under the `syscall` package. 
 https://github.com/golang/sys/commit/1e99a4f9d247b28c670884b9a8d6801f39a47b77

shirou: Sorry for my laziness since #30, we should move to `x/sys/unix`. I agree. However, I think this migration is not a part of this PR. So could you remove from this PR? Or, if  `x/sys/unix` is required to work on illumos, I will make a new PR to migrate to `x/sys/unix` (but it takes a little time).

sean-:
This PR is now based on top of #380.  Once #380 is merged then I'll repush this branch.

sean-:
@shirou this should be G2G to review and merge.  The JSON parsing uses `easyjson` to generate the JSON parser so values can be decoded directly into their respective `uint64` values.  It's also faster and takes less memory according to the [`easyjson`](https://github.com/mailru/easyjson#benchmarks) README.

shirou:
Could you revert some commits about rename `*_solaris.go` to `*_illumos.go` ? Filename is used to choose target by golang compiler, and `solaris` is reserved for `GOOS` (see [here](https://golang.org/doc/install/source#environment)).

And, I support using `easyjson` in order to generate JSON faster, but I think it may be not included this PR. Could you separate?

sean-:
@shirou All of the `_illumos.go` files have the required `// +build solaris` tags in order for Go to build the package correctly.  These `kstat(1M)` values are not present on Solaris and are only available on Illumos.  For example:

https://github.com/illumos/illumos-gate/commit/9468939ef8704ee9aba7596c1e9ff9b059109cac

As for `easyjson`, the only thing that's required to be checked in is the `easyjson` generated files, which have been added:

* https://github.com/shirou/gopsutil/pull/378/files#diff-66dc619370689636d58049ee8ad71a32
* https://github.com/shirou/gopsutil/pull/378/files#diff-dfb15f1ddb55a38f514623e138d75179
* https://github.com/shirou/gopsutil/pull/378/files#diff-705a30ef1a2bbc3afa04ed161cd20a0d

The tool is available via the [`tools`](https://github.com/shirou/gopsutil/pull/378/files#diff-b67911656ef5d18c4ae36cb6741b7965R15) target and regeneration of the generated files can be recreated via the [`generate`](https://github.com/shirou/gopsutil/pull/378/files#diff-b67911656ef5d18c4ae36cb6741b7965R19) target.  The `generate` target will only be usable on platforms where `GOOS=="solaris"`.  I moved the generated files behind the necessary build tags so that non-Illumos variants won't carry around the extra code necessary to parse and won't be required to install `easyjson`.

Downstream consumers of `shirou/gopsutil` who are building on Illumos will now be required to vendor the following three packages:

* `github.com/mailru/easyjson`
* `github.com/mailru/easyjson/jlexer`
* `github.com/mailru/easyjson/jwriter`

However that exercise is left to the consumer since `shirou/gopsutil` is a library and not an application.

Does that work/make sense?

shirou:
Well, I am not familiar with Solaris, but `kstat` is only on illumos, only one of a Solaris distribution? Since gopsutil aim to become widely used for multi platform and multi distribution, I think it is hard to add distribution (not platform) dependent functionality.

sean-:
@shirou `kstat` is available on multiple Solaris distributions so this should work for most of the `solaris` build targets.  Some of the counters are platform specific.  I view this PR in the same vein as `*_bsd.go` using `sysctl(1)` to generate stats: the expectation is that there will be a need to add a `freebsd.go` or `openbsd.go` to address their platform-specific needs.

At present the worst that will happen is some of the values will be `0`, however, the best that will happen is values will be non-zero and correct.  Which is to say, this PR is strictly an improvement over the current state.

As issues arise, I'm happy to continue to try and address.

shirou:
`openbsd`, `freebsd`, `netbsd` and also `dragonfly` are called BSD family but have different GOOS target. not distribution.

sean-:
@shirou right, but there isn't a `GOOS` yet for Illumos vs Solaris vs SmartOS.  When that happens, I'm happy to revisit.

sean-:
@shirou Until golang/go#20603 is accepted or finished, could you please review and merge this PR?  Once the PR is merged I will back out the `solaris` build tags and will leave it up to a qualified member of the Solaris community to add the necessary build tags once they verify the functionality works.

shirou:
Understood. I leave this PR open. If there is progress, could you write comment here? Thanks.

Smithx10:
What is the status of this?  I've recently started diving into Nomad and SmartOS and require this functionality.

shirou:
#20603 is still not merged, so no one is working I think.

Smithx10:
I'll open up a new PR for this when the new illumos tag is added to go 1.13 at the end of July.

Lomanic:
[Go 1.13 has landed with support for the illumos GOOS build tag](https://golang.org/doc/go1.13#illumos). Waiting for the corresponding PR @Smithx10 😉

Though I would prefer to not add a new external dependency (easyjson) just for illumos.

Smithx10:
@Lomanic,

The latest news is that @jperkin was working on a PR.   I backed off and am letting him do his thing.   

@jperkin if you are short on cycles,   maybe one of us could pick up the slack and get a PR in.   Do you by chance have a branch out there?

Thank You,
Bruce Smith

PR #379, opened 2017-06-01
Fix data race in host.BootTime

magiconair:
When `host.BootTime` is accessed concurrently there is a data race on `cachedBootTime`. Since `host.BootTime` is the only method accessing this variable I've moved it in front of the function using it to make it less likely to just use it by accident.


shirou:
Indeed. Thank you for your great contribution!

PR #380, opened 2017-06-02
Move from the `syscall` package to the `golang.org/x/sys/{unix,windows}`

sean-:
This is a very mechanical change: `syscall.` -> `unix.` or `windows.`

Assuming this passes compile time checks, this should be g2g.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -425,7 +425,7 @@ func getMemoryInfo(pid int32) (PROCESS_MEMORY_COUNTERS, error) {
 }
 
 func getProcessMemoryInfo(h windows.Handle, mem *PROCESS_MEMORY_COUNTERS) (err error) {
-	r1, _, e1 := windows.Syscall(procGetProcessMemoryInfo.Addr(), 3, uintptr(h), uintptr(unsafe.Pointer(mem)), uintptr(unsafe.Sizeof(*mem)))
+	r1, _, e1 := syscall.Syscall(procGetProcessMemoryInfo.Addr(), 3, uintptr(h), uintptr(unsafe.Pointer(mem)), uintptr(unsafe.Sizeof(*mem)))
 	if r1 == 0 {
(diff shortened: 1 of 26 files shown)


process/process_windows.go
```diff
@@ -412,25 +412,25 @@ func getRusage(pid int32) (*syscall.Rusage, error) {
...
 		return mem, err
 	}
 
 	return mem, err
 }
 
-func getProcessMemoryInfo(h syscall.Handle, mem *PROCESS_MEMORY_COUNTERS) (err error) {
-	r1, _, e1 := syscall.Syscall(procGetProcessMemoryInfo.Addr(), 3, uintptr(h), uintptr(unsafe.Pointer(mem)), uintptr(unsafe.Sizeof(*mem)))
+func getProcessMemoryInfo(h windows.Handle, mem *PROCESS_MEMORY_COUNTERS) (err error) {
+	r1, _, e1 := windows.Syscall(procGetProcessMemoryInfo.Addr(), 3, uintptr(h), uintptr(unsafe.Pointer(mem)), uintptr(unsafe.Sizeof(*mem)))
```
shirou: on Windows, `windows.Syscall` and `windows.EINVAL` is not exists.

shirou:
Thank you for this awesome contribution!
I confirmed it works on

- Linux amd64
- Darwin amd64
- FreeBSD amd64

However,  `process/process_windows.go` could not compiled on Windows. It seems more study required to invoke dll proc using `x/sys/windows` packages. So, could you revert changes only ` process/process_windows.go`? other part of this PR is great.

Thank you again!

sean-:
@shirou Can you give the latest version a twirl on Windows?  I backed out the changes for identifiers and functions that I couldn't find in the `x/sys/windows` package.  I _think_ this should be correct for Windows now.

shirou:
It can be build on Windows now! Thank you for your cool contribution!

PR #382, opened 2017-06-09
Only read /proc/stat once when cpu.Times(true) is called on Linux

acd:
This commit eliminates unnecessarily rereading /proc/stat when loading the CPU usage for each processor on Linux.

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -33,18 +33,15 @@ func Times(percpu bool) ([]TimesStat, error) {
 	filename := common.HostProc("stat")
 	var lines = []string{}
 	if percpu {
-		var startIdx uint = 1
-		for {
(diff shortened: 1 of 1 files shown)


shirou:
Thank you for this awesome PR, yes, I agree.

However, it seems this PR breaks a behavior when `stat` is empty (related to #311).
Current behavior is return `[], nil` when empty, but this PR returns [error](https://github.com/shirou/gopsutil/pull/382/files#diff-fe74a9d71b6cc5d2ac79830c395cc590R41). I think return error is better, but I don't want to change behavior. 

Could you change to return nil? After change, I will merge soon.
Thank you.

acd:
Sure thing.  I updated the commit.

shirou:
I'm very sorry to late response. Thank you for your contribution!

PR #387, opened 2017-06-24
collect more data on windows

tychoish:


shirou:
Thank you! It works! However, this PR includes change of `cpu/cpu_windows.go`. But it seems not necessary. Could you remove this change or let me know if it is required. Thank you for your contribution!

tychoish:
I removed the file in question! Sorry for the noise!

tychoish:
It also strikes me that with this change there are some potential updates that we could make to the readme, as I think we have pretty good feature parity (or could easily obtain) on windows.

shirou:
Thank you for updating! Now ready to merge.
If you not mind, I will change `x` to  `process.parent` and `process.children` on the README after merge.

shirou:
I will change README about your great contribution. Thank you.

PR #392, opened 2017-06-28
[windows] add implementation for process.Username()

derekwbrown:
(Windows) Added implementation of Username()
Changed the calls to OpenProcess() to use  PROCESS_QUERY_LIMITED_INFORMATION rather than  PROCESS_QUERY_INFORMATION, as it's more likely to be authorized.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -181,8 +181,26 @@ func (p *Process) Status() (string, error) {
 	return "", common.ErrNotImplementedError
 }
 func (p *Process) Username() (string, error) {
-	return "", common.ErrNotImplementedError
+	pid := p.Pid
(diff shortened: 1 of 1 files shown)


shirou:
Thank you so much! Your PR works on my Windows environment.

However, the test uses `os/user`.
```
        currentUser, _ := user.Current()
        myUsername := currentUser.Username
```
And, `user.Current()` returns `<machine-name>\\<username>` so test failed. Do you have any idea to pass this test? 

Thank you !

derekwbrown:
Function now returns the proper domain\\username syntax, so should pass the test.

shirou:
I have confirmed your update can pass the test on my Windows environment. Thank you for your awesome contribution!

PR #395, opened 2017-07-10
err omit in getProcInodes

better88:


shirou:
Good catch!  Thanks a lot!

PR #402, opened 2017-07-17
Fix process.Exe() on macOS 10.12

chrisroberts:
The `lsof` command on macOS < 10.12 always selects the `p` field
when using the `-F` option. On macOS 10.12 the `f` field is also
always selected causing an incorrect result. This modification adds
both options which are always selected to maintain consistency.


shirou:
I don't have < 10.12 MacOS, so I can not confirm. But it looks like not break compatibility . Thank you .

PR #403, opened 2017-07-20
set win cpu total name to "cpu-total"

HeChuanXUPT:
set the name of windows cpu total percent to "cpu-total"
keep the same with Linux, Freebsd system


shirou:
Good point. Thank you!

PR #405, opened 2017-07-22
added function for calculation of CPU usage percentage of a process

Aditya267:


shirou:
Thank you. This is a short and basic code so I feel it might be not required for many users, but yes, some user like you want this. Thus I merge this. Thank you for your contribution!

PR #408, opened 2017-08-01
Add HostID support to the `solaris` build target.

sean-:
This falls back to `hostid(1)` on non-SmartOS platforms and satisfies the definition of host identity well enough.


shirou:
I haven't checked it is work on solaris, but I merged. Thank you for contributing for wide platform support!

PR #412, opened 2017-08-02
Just look at filenames for linux NumFDs call

conorbranagan:
Speed up `NumFDs` on Linux by just using a list of fd filenames and finding the length vs reading every file contents when we don't need it.


shirou:
Good catch! Thank you!

PR #417, opened 2017-08-16
Alter subprocess's environment instead of the hosts

### LINKED ISSUE #415: LC_ALL is set on calling process, not on subprocess only
The fix for #159 causes an issue for any processes and their children which do not play nicely with the C encoding.

Instead of using `os.Setenv` to change the locale, the correct behavior would be to clone the process's Context and adjust the LC_ALL key to have the value C and then using exec.CommandContext instead of exec.Command

Please see [hashicorp/nomad#3036](https://github.com/hashicorp/nomad/issues/3036) for an example of this causing some trouble in the wild.

schmichael:
Fixes #415

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -358,3 +358,19 @@ func Pipeline(cmds ...*exec.Cmd) ([]byte, []byte, error) {
 	// Return the pipeline output and the collected standard error
 	return output.Bytes(), stderr.Bytes(), nil
 }
+
+// getSysctrlEnv sets LC_ALL=C in a list of env vars for use when running
+// sysctl commands (see DoSysctrl).
+func getSysctrlEnv(env []string) []string {
+	foundLC := false
+	for i, line := range env {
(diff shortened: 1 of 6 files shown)


nugend:
There are a number of additional uses of LC_ALL in the other common_* platform files. *Also*, since this was setting that for the entire process space, the usage of exec.Command in [process_darwin.go](https://github.com/shirou/gopsutil/blob/5329f8ba79fc14ad7238a3c2e89f776621a35b85/process/process_darwin.go) and [common.go](https://github.com/shirou/gopsutil/blob/c251591dc712722a0db836b8497ff3b00031ae9a/internal/common/common.go) might also need to be replaced

schmichael:
@nugend Ha, indeed. Fixed them all and added a little test for the env munging.

nugend:
@schmichael Thanks. I think I just need to bite the bullet and learn Go. I feel like every utility is written in it now :P

shirou:
Good catch! Thank you for fixing.

PR #419, opened 2017-08-26
Add type definitions for mips and mipsle

hnw:
Support mips and mipsle. I ran `cgo` on Linux/MIPS machine (on QEMU) and got those files.


shirou:
I can not confirm this works, but I merge this. Thank you for your contribution!

PR #422, opened 2017-09-16
add support for oom-killer score (oom_score)

dastergon:
The [oom_score](https://github.com/torvalds/linux/blob/v4.12/Documentation/filesystems/proc.txt#L38)(`/proc/(pid)/oom_score`) is a per-process parameter that is used to determine the likelihood of a process getting terminated by the Out-of-Memory (OOM) killer of Linux Kernel. The higher the value, the more likely the process will be killed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -235,6 +235,7 @@ parent              x            x       x       x
 children            x     x      x       x       x
 connections         x            x       x
 is_running
+oom_score           x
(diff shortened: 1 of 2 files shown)


shirou:
OOM Killer is an Linux specific feature and I don't want to add such a platform dependent function, sorry.

dastergon:
No problem. Thanks for letting me know 👍

PR #426, opened 2017-09-28
Fix NaN percentage if process was created too soon

disq:
If the process was created within the second or so of a `process.CPUPercent()` call, it would return `NaN`: the crt_time calculation is made with second-precision so it causes a division-by-zero, but since we are in the float64 space, the result is `NaN`.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -214,6 +214,9 @@ func (p *Process) CPUPercent() (float64, error) {
 
         created := time.Unix(0, crt_time * int64(time.Millisecond))
         totalTime := time.Since(created).Seconds()
+        if totalTime <= 0 {
+                return 0, nil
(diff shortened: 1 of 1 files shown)


shirou:
good catch. But, how about using original variable calculation ? such as
```
totalTime := float64(time.Now().Unix()-(crt_time/1000)))
if totalTime <= 0 {
    return 0, nil
}
 return 100 * cput.Total() / totalTime
```

disq:
`time.Since` is idiomatic Go (whereas subtracting two unix seconds is not) and also since it's calculated at the millisecond level, this way it's more precise.

shirou:
Indeed, Understood. Thank you for your contribution!

PR #428, opened 2017-09-29
Implement Threads() in Linux

Leonid99:


Leonid99:
Currently, `Threads()` returns threads in the form of `map[string]string`. I wasn't sure what this data structure is supposed to represent, so I changed it to a data stracture that made sense for the case of Linux.

shirou:
Thank you!
But, since original [psutil returns map[thread_id]thread CPU times (user/system)](https://pythonhosted.org/psutil/#psutil.Process.threads), I don't want to change that signature.

Leonid99:
Thank you for your response. Python returns a list of named tuples: ```In [13]: p=psutil.Process(28302)

In [14]: p.threads()
Out[14]: 
[pthread(id=28302, user_time=4.21, system_time=0.31),
 pthread(id=28303, user_time=0.0, system_time=0.0),
....```

Leonid99:
@shirou Fixed it to be a `map`, as you originally suggested. Please tell me what you think.

shirou:
Thank you for changing your code. And sorry for merge other PR, could you fix conflicts about `process/proces.go` ? It seems only fmt change, just revert will be worked.

Basically, I don't want to change API signature. but after digging my code, I found no platform implement `process.Thread()`. So a impact of changing API is minimal, I feel. Thus, I will merge soon if you fix the conflicts.

Leonid99:
Thank you. I have removed the offending commit (it was a whitespace change on the file that I ended up not touching, so it was irrelevant for the change in question).
Thank you for this library!

shirou:
Thank you for your cool contribution!

PR #434, opened 2017-10-05
Ignore errors with single inode when getting all proc inodes

### LINKED ISSUE #433: net.Connections("all") returns error if permission denied opening pid directory
When upgrading from v2.17.04 to v2.17.08 we found that it is no longer possible to call `net.Connections("all")` on Linux unless you have permission to read all pid directories in /proc.

```go
func main() {
        res, err := net.Connections("all")
        if err != nil {
                fmt.Println(err)
                return
        }

        for _, item := range res {
                fmt.Println(item)
        }
}
```

With this code the output is `cound not get pid(s), 0`.  It looks like the error is originally from here: https://github.com/shirou/gopsutil/blob/master/net/net_linux.go#L419 and the error string `open /proc/1/fd: permission denied`.  When using the v2.17.04 the `[]net.ConnectionStat` are returned for directories that can be read.

As expected, if ran with root permissions items are returned correctly without erro.

danielnelson:
To fix this issue I think we should just ignore any error reading the fd.  I considered ignoring only permission errors but I think due to race conditions with processes we should ignore all errors.

fixes #433

### DIFF
--- a/net/net_linux.go
+++ b/net/net_linux.go
@@ -541,7 +541,7 @@ func getProcInodesAll(root string, max int) (map[string][]inodeMap, error) {
 	for _, pid := range pids {
 		t, err := getProcInodes(root, pid, max)
 		if err != nil {
-			return ret, err
+			continue
 		}
 		if len(t) == 0 {
 			continue


shirou:
I haven't noticed this PR. I'm so sorry,   perhaps it is almost same as e01a14e318951de4782908fbdb2b66a652a22065

PR #438, opened 2017-10-13
Implement Process.Times for Windows

mlkm:


grrrrrrrrr:
Hey @shirou , is there a chance we can get an update on this? We'd like to find out if gopsutil is an option for us to use, any answer would help. Thanks!!

PR #439, opened 2017-10-16
Add HOST_VAR envvar and use to locate utmp on Linux.

danielnelson:
Add `HOST_VAR` envvar and use to locate utmp on Linux.  

As discussed in https://github.com/influxdata/telegraf/issues/3339


shirou:
Thank you!!

PR #440, opened 2017-10-18
Export some useful funcs

chenchun:


shirou:
Why don't you implement these funcs by yourself?

chenchun:
Yes, I did copy these funcs to my project. But isn't it more convenient to make them exported for users of this lib？

shirou:
i don't think so. We already exported a calculated value. Most users don't care about how to calculate. And if exported once, we can not change that function's signature.

PR #444, opened 2017-11-06
make process.processes function public

tychoish:
We currently provide a 'list all pids' function which provides this
(or related) functionality, so if you wanted to collect data on all
running processes, we would end up doing redundant work, when we could
make the function public. 

I also ironed out a few cross-platform inconsistencies.


shirou:
Because we already has `Pids()` and easy to make `[]Process` from pids, it is not entirely convinced for me to introduce new API, but it can be understand what your PR want. So I merge it. 

Anyway, thank you for your suggestion and contribution!

PR #448, opened 2017-11-08
Fix IOCounters for OpenBSD

mpfz0r:
Use SysctlRaw instead of Sysctl.
The latter assumes NUL terminated strings
which returns the lenght off by one.
Therefore, only n-1 disks where reported.


shirou:
Good catch! Thank you!

PR #449, opened 2017-11-08
Implement Connection support for OpenBSD

mpfz0r:
This retrieves open TCP/UDP connections by using netstat(1)
File descriptors and pids are not supported.

### DIFF
--- a/net/net_openbsd.go
+++ b/net/net_openbsd.go
@@ -97,7 +97,7 @@ func ParseNetstat(output string, mode string,
 }
 
 func IOCounters(pernic bool) ([]IOCountersStat, error) {
-	netstat, err := exec.LookPath("/usr/bin/netstat")
+	netstat, err := exec.LookPath("netstat")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


net/net_openbsd.go
```diff
@@ -146,8 +151,138 @@ func ProtoCounters(protocols []string) ([]ProtoCountersStat, error) {
...
+func parseNetstatAddr(local string, remote string, family uint32) (laddr Addr, raddr Addr, err error) {
+	parse := func(l string) (Addr, error) {
+		matches := portMatch.FindStringSubmatch(l)
+		if matches == nil {
+			return Addr{}, fmt.Errorf("wrong addr, %s", l)
+		}
+		host := matches[1]
+		port := matches[2]
+		if host == "*" {
+			if family == syscall.AF_INET {
```
shirou: Could you use `switch-case` ? `AF_INET`, `AF_INET6` and default that returns 'unknown family' error.

net/net_openbsd.go
```diff
@@ -146,8 +151,138 @@ func ProtoCounters(protocols []string) ([]ProtoCountersStat, error) {
...
+	args := []string{"-na"}
+	switch strings.ToLower(kind) {
+	default:
+		fallthrough
+	case "":
+		fallthrough
+	case "all":
+		fallthrough
+	case "inet":
+
```
shirou: if nothing to add, could you add a comment about `nothign to add` or something?

net/net_openbsd.go
```diff
@@ -146,8 +151,138 @@ func ProtoCounters(protocols []string) ([]ProtoCountersStat, error) {
...
+		args = append(args, "-pudp")
+	case "udp4":
+		args = append(args, "-pudp", "-finet")
+	case "udp6":
+		args = append(args, "-pudp", "-finet6")
+	case "unix":
+		return ret, common.ErrNotImplementedError
+	}
+
+	netstat, err := exec.LookPath("/usr/bin/netstat")
```
shirou: gopsutil already have many of `LookPath` which has absolute path, but it is wrong. `LookPath` searchs from PATH so just use `netstat`.

shirou:
Thank you so much!

Issue #451, opened 2017-11-09
hang on package import on older-than-2016 windows platforms


leafnunes:
We have a golang service uses gopsutil to gather metrics and cpu usage. The first time the binary runs, it hangs forever on a WMI query that never returns (called from the init in `cpu/cpu.go`, Times).

Most of the queries in gopsutil should probably return pretty quickly or timeout.

I have a working change that wraps just the hanging call, but could extend it to every wmi.Query call if that seems reasonable.

shirou:
I close this issue. But the fundamental problem is not solved yet, if you find something, please re-open or create a new issue. Thank you!

PR #452, opened 2017-11-10
prevent hang on pkg import if wmi.Query hangs

leafnunes:
This is to address #451

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -3,6 +3,7 @@
 package cpu
 
 import (
+	"context"
 	"fmt"
 	"unsafe"
 
@@ -81,8 +82,9 @@ func Info() ([]InfoStat, error) {
 	var ret []InfoStat
 	var dst []Win32_Processor
 	q := wmi.CreateQuery(&dst, "")
-	err := wmi.Query(q, &dst)
-	if err != nil {
+	ctx, cancel := context.WithTimeout(context.Background(), common.Timeout)
+	defer cancel()
+	if err := common.WMIQueryWithContext(ctx, q, &dst); err != nil {
 		return ret, err
 	}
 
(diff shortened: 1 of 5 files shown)


shirou:
Could you dig more deeper why it hangs? It happens on other Windows environment?

And, I agree to add a timeout feature. But recent golang has `context` library which has [timeout](https://golang.org/pkg/context/#WithTimeout). Could you change to use `context` with some small wrap function?

leafnunes:
My access to windows instances is relatively limited, especially older platforms. Downstream users have observed this on windows server 2008 and 2012, but not 2016. If i can get ahold of decent debugging tools and regular access to a susceptible image, i can do more debugging.

Before i do the change, let me make sure i understand what you want: would you prefer i wrap every wmi.Query call with a context with timeouts, or just the one i'm wrapping now? Is using `common.Timeout` as the default for all of them ok with you?

shirou:
Could you create a wrapper function in `internal/common/common_windows.go`.
The wrapper function is like this.
```
func WMIQueryWithContext(ctx context.Context, query string, dst interface{}, connectServerArgs ...interface{}) error {
	errChan := make(chan error, 1)
	go func() {
		errChan <- wmi.Query(query, dst, connectServerArgs)
	}()

	select {
	case <-ctx.Done():
		return ctx.Err()
	case err := <-errChan:
		return err
	}
}
```

Then, we can use like this
```
	ctx, cancel := context.WithTimeout(context.Background(), common.Timeout)
	defer cancel()
	if err := WMIQueryWithContext(ctx, q, ret); err != nil{
		return err
	}
```

using `common.Timeout` is fine. and I will appreciate if you change all functions which use `wmi.Query`, but only `cpu.PerfInfo()` is too enough.

leafnunes:
How's this look?

shirou:
perfect!! Thank you for your contribution!!

PR #453, opened 2017-11-12
Use w32.EnumProcesses to get pids on Windows in process.Pids()

Lomanic:
Simple benchmark:

```go
package main

import (
	"github.com/shirou/gopsutil/process"
	"log"
	"time"
)

func main() {
	start := time.Now()
	pids, err := process.Pids()
	if err != nil {
		log.Fatalln("Could not read processes,", err)
	}
	log.Println("It took", time.Now().Sub(start), "to get", len(pids), "results")
}
```

Before PR:

    It took 388.0222ms to get 123 results
    
After PR:
```
It took 2.0001ms to get 123 results
```

PR #454, opened 2017-11-12
Handle case in Windows process.Pids() where buffer is too small to handle that many processes returned by w32.EnumProcesses

Lomanic:
See https://github.com/giampaolo/psutil/blob/1c3a15f637521ba5c0031283da39c733fda53e4c/psutil/arch/windows/process_info.c#L315-L329

You can see it in action by changing psSize size and increment to much lower values (dividable by 4, for example `var psSize uint32 = 128` and `psSize += 128` where we would retrieve 32 more processes at each iteration).

The previous PR #453 would not retrieve every processes if there were more than 2048/4 -> 512 processes running, this is now handled by this commit.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -94,18 +94,28 @@ func init() {
 
 func Pids() ([]int32, error) {
 	// inspired by https://gist.github.com/henkman/3083408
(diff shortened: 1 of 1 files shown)


shirou:
wow, Thank you!

PR #455, opened 2017-11-13
Use w32.EnumerateProcesses instead of slower wmi in windows process.Processes()

Lomanic:
As for #453


shirou:
awesome! Thank you!

PR #458, opened 2017-11-17
[net] linux: skip if not exist error on getProcInodesAll

danielnelson:
In https://github.com/shirou/gopsutil/commit/e01a14e318951de4782908fbdb2b66a652a22065 we added skipping of permission errors, but I am getting reports of errors like the following:
```
cound not get pid(s), 0: open /rootfs/proc/32665/fd: no such file or directory
```

I think this is caused by `Pids()` returning a pid which exits before `getProcInodes()` can process it.


shirou:
Good catch! Thanks!

Issue #362, opened 2017-05-04
Plan to v3


shirou:
v2 is released May, 2016. After one year, I think it is time to release v3. This issue files breaking changes which we wanted to include v3.

Note: v3 release schedule is not fixed.

- [x] RlimitStat change to int64 from uint64 (#346)
- [x] remove `process.NetIoCounter()` (#429)
- [x] fix typo: "memoryLimitInBbytes" -> memoryLimitInBytes (#464)
- [x] add `WithContext` methods (#469)
- [x] `process.GetWin32Proc` should not be exposed
- [x] more broadly, have the same public API/types on all platforms (no more public API/types/constants defined only for some platforms)
- [x] mem.VirtualMemoryStat JSON fields proper capitalization (#545)
- [ ] return detailed state in `process.State` (as `[]string, err`), instead of just one state (currently, R: Running S: Sleep T: Stop I: Idle Z: Zombie W: Wait and L: Lock). Linked to #596 (determining background state, see https://github.com/shirou/gopsutil/issues/596#issuecomment-434207724)
- [x] rename `load.MiscStat` for a better name #680

phemmer:
> RlimitStat change to int63 from int32

Needs to be `uint64`. While I'm assuming "int63" is a typo for "int64", signed still doesn't support all possible values.

This is the definition of `struct rlimit`:
```
struct rlimit {
  unsigned long rlim_cur;
  unsigned long rlim_max;
};
```

shirou:
Thanks. I've updated.

shirou:
Finally v3 released #938.

PR #464, opened 2017-12-03
CgroupMem : fix typo and wrong file names

barkbay:
Hi, just a PR to fix a typo.


shirou:
Thank you! However, since JSON name change breaks compatibility, so we can not accepted this even if a typo. I added to #362. So, could you remove name change ? cgroup file path fix is a very good catch!

shirou:

PR #467, opened 2017-12-08
Fix build with bazel on OSX

tanyatik:
Following guide here: 

https://github.com/bazelbuild/rules_go/issues/646#issuecomment-349676747

Checked local build and seems to work.


shirou:
Since I have not MacOSX now, I can't confirm. But it should work. Thank you for your contribution!

shirou:
Sorry, This PR is reverted due to build error on linux. https://circleci.com/gh/shirou/gopsutil/357

```
go get -t -d -v ./...
can't load package: package github.com/shirou/gopsutil/host: C source files not allowed when not using cgo or SWIG: smc.c
package github.com/shirou/gopsutil/host: C source files not allowed when not using cgo or SWIG: smc.c
```

tanyatik:
I created a new PR that should pass build on linux.
I added an empty file host_cgo.go , so that cgo is always used.

I could not find another workaround. 
I tried to build smc.c and smc.h into a library as suggested here  https://github.com/bazelbuild/rules_go/issues/646#issuecomment-349710509   but it requires changes in BUILD.bazel which cannot be generated by gazelle.
I realise this is quite hacky, so let me know if you have better idea on how to fix it.

PR #470, opened 2017-12-14
Fix build with bazel on OSX attempt 2 (add stub for host_cgo.go)

tanyatik:
I believe this should pass build on linux.
Added an empty file host_cgo.go , so that cgo is always used.

I could not find another workaround (I tried to build smc.c and smc.h into a library as suggested here  https://github.com/bazelbuild/rules_go/issues/646#issuecomment-349710509   but corresponding BUILD.bazel can't be generated by gazelle at the moment, it seems).
This is quite hacky, so let me know if you have a better idea on how to fix it.


tanyatik:
I have checked that it builds on linux

shirou:
Same error with #467 is occurred.
```
can't load package: package github.com/shirou/gopsutil/host: C source files not allowed when not using cgo or SWIG: smc.c
```

`go version go1.9 linux/amd64`

And, at least [here](https://github.com/shirou/gopsutil/pull/470/files#diff-d0628db12f51a73c116a76fd1b4f598cR3) should `package host`.

tanyatik:
@shirou   sorry it took me a while to get back to this.  Could you please run CircleCI on this branch to see if it works now?  Thanks!

shirou:
```
In file included from smc.c:39:0:
smc.h:37:28: fatal error: IOKit/IOKitLib.h: No such file or directory
 #include <IOKit/IOKitLib.h>
                            ^
compilation terminated.
```
I will revert it.

PR #471, opened 2017-12-14
Add support for parsing the tgid (thread group id) field.

nkirsch:
Parses the tgid field, which is the thread group id (aka process id) on Linux.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -30,6 +30,8 @@ type Process struct {
 
 	lastCPUTimes *cpu.TimesStat
 	lastCPUTime  time.Time
+
+	tgid int32
 }
 
 type OpenFilesStat struct {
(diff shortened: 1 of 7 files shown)


shirou:
gopsutils is aim to be multi platform library. So adding API only Linux is not welcome. At least windows, freebsd, and darwin should supported. But `NotImplementedYet` is ok.

shirou:
@nkirsch Could you add Windows ?

Below is just a comment for some others.

From  [5.2.2. Linux® processes and thread identifiers](https://www.freebsd.org/doc/en_US.ISO8859-1/articles/linux-emulation/mi.html) , there is not TGID at least FreeBSD. (perhaps, Linux only). So we should decide fake it or leave `NotImplemented` (or close this PR).

According to `5.2.3. PID mangling`, 
- in kernel (in Linuxulator) PID = PID and TGID = shared -> group pid
- to userland we present PID = shared -> group_pid and TID = proc -> p_pid
But we can not change PID meaning in this API version. So I am thinking about return `KInfoProc -> Pgid` as TGID.

shirou:

nkirsch:
@shirou My pleasure! Thanks for the wonderful library.

PR #478, opened 2017-12-26
[mem] Add commit-related stats to VirtualMemoryStat on Linux

olivielpeau:
Adds 2 memory statistics: `CommitLimit` and `CommittedAs`

### Motivation

Highlighting the docs at https://www.kernel.org/doc/Documentation/filesystems/proc.txt:

> CommitLimit: Based on the overcommit ratio ('vm.overcommit_ratio'), this is the total amount of  memory currently available to be allocated on the system.

> Committed_AS: The amount of memory presently allocated on the system. [...] This is useful if one needs to guarantee that processes will not fail due to lack of memory once that memory has been successfully allocated.

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -58,6 +58,8 @@ type VirtualMemoryStat struct {
 	Slab         uint64 `json:"slab"`
 	PageTables   uint64 `json:"pagetables"`
 	SwapCached   uint64 `json:"swapcached"`
+	CommitLimit  uint64 `json:"commitlimit"`
+	CommittedAS  uint64 `json:"committedas"`
 }
 
 type SwapMemoryStat struct {
--- a/mem/mem_linux.go
+++ b/mem/mem_linux.go
@@ -60,6 +60,10 @@ func VirtualMemory() (*VirtualMemoryStat, error) {
 			ret.PageTables = t * 1024
 		case "SwapCached":
 			ret.SwapCached = t * 1024
+		case "CommitLimit":
+			ret.CommitLimit = t * 1024
+		case "Committed_AS":
+			ret.CommittedAS = t * 1024
 		}
 	}
 	if !memavail {
--- a/mem/mem_test.go
+++ b/mem/mem_test.go
@@ -57,7 +57,7 @@ func TestVirtualMemoryStat_String(t *testing.T) {
 		UsedPercent: 30.1,
 		Free:        40,
 	}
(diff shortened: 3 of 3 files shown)


Lomanic:
Hi, thanks for your contribution @olivielpeau!

As gopsutil aims to be platform-agnostic, do you know if there is something equivalent for the other OSes supported (freebsd, openbsd, darwin, windows) instead of returning 0 if this gets merged? 

For example on Windows, which property of [memoryStatusEx](https://github.com/shirou/gopsutil/blob/6e221c482653ef05c9f6a7bf71ddceea0e40bac5/mem/mem_windows.go#L17) (or [performanceInformation](https://github.com/shirou/gopsutil/blob/6e221c482653ef05c9f6a7bf71ddceea0e40bac5/mem/mem_windows.go#L47)?) would it be?

olivielpeau:
Hi @Lomanic, thanks for your response! (and sorry I'm only answering now)

I had a look around and didn't seem to find similar stats on other platforms. The overcommit behavior of Linux is quite specific to the platform, so I don't think these 2 stats make much sense for other platforms.

I still think having these 2 stats is helpful on Linux (especially when `vm.overcommit_memory` is set to "don't overcommit", but also with "heuristic overcommit", see https://www.kernel.org/doc/Documentation/vm/overcommit-accounting).

Here's what I found on other OSes:
* darwin doesn't provide anything on that in [vm_statistics](http://web.mit.edu/darwin/src/modules/xnu/osfmk/man/vm_statistics.html) or the `vm_stat` command.
* freebsd doesn't have an equivalent to Linux's `overcommit_ratio`, and its definition of total allocatable memory depends on the [`vm.overcommit`](https://wiki.freebsd.org/SystemTuning#SYSCTL_TUNING) sysctl. As such I'm not sure it makes sense to add a `CommitLimit` equivalent. Also, I couldn't find exactly what a "currently allocated memory" equivalent (`Commited_AS`) would be on FreeBSD.
* not sure about Windows, I don't see much in these 2 structs that could be equivalents.

What do you think?

shirou:
Thank you @olivielpeau and sorry for late response. I was very think and consider carefully about introducing platform dependent value. But in this case, 
- other platform not affected, just return 0, 
- other platform doesn't have same meanings.
So, I merge this PR. Thank you for your contribution.

But, I will add a `Linux Only` comment to these variables.

Issue #469, opened 2017-12-11
[Windows] cpu.Info() timing out


FlorinAsavoaie:
In `github.com/shirou/gopsutil/internal/common`, there is a Timeout hardcoded for the WMI queries. When the CPU is busy, `cpu.Info()` (and maybe other stuff as well) is failing as the WMI is taking too long to respond to the query. I would expect to be able to pass this Timeout from my code.

vrecan:
I've run into this a few times as well. It would be nice to be able to pass in the context timeouts or set them in some way.

shirou:
I have added plan to v3 (#362) about `WithContext` method.

sadag:
Please review PR #488

Lomanic:
Closing as a duplicate of #394 (as always, WMI calls should be avoided)

PR #480, opened 2017-12-31
Add WithContext functions.

shirou:
This PR addes `WithContext` functions to all functions, all platform, by using [this script](https://gist.github.com/shirou/edd80a2c13d889b178e0929691d6d3e7).

related to #469. I had planed to v3 but I think just adding context does not change API.

@Lomanic I'd very appreciate if you could check this PR. I have confirmed go test works on Linux, FreeBSD.


Lomanic:
Sorry for late response, I was not able to run the tests (`go test ./...`) without errors on Windows 7 64bit.

I notice that unfortunately the `process.ProcessesWithContext()` function is missing btw.

shirou:
@Lomanic Could you show me what test is failing on Windows 7 (plz ignore `not implemented yet`)?

And add `process.ProcessesWithContext` is added. Thank you!

PR #487, opened 2018-02-12
[host] fix race condition to ret

truthbk:
Because a go routine is spawned when calling [this](https://github.com/shirou/gopsutil/blob/master/internal/common/common_windows.go#L119-L121):
```common.WMIQueryWithContext(ctx, q, &ret)```

You may actually get an error return [see here](https://github.com/shirou/gopsutil/blob/master/internal/common/common_windows.go#L123-L128), returning from the caller before the go routine has guaranteed completion; this triggers a race condition to read and write from `ret` in the return expression and in the actual StackExchange `wmi` module.

Addressed in the same manner here in the [host subpackage](https://github.com/shirou/gopsutil/blob/master/host/host_windows.go#L123-L126)


shirou:
Good catch! Thank you!

However since `WMIQueryContext` is introduced before other `WithContext` methods, we will remove `WMIQueryContext` and use `Context` directory. So I apologize if your valuable commit is removed by future commits.

PR #488, opened 2018-02-12
windows: fix WithContext functions

sadag:
set timeout only if parent context doesn't set.

Found an issue with host.Info() and other calls during windows boot. It takes 13s to complete these calls. This pull request enables the user set the timeout in `context`, also fixes passing `ctx` across API boundaries.


shirou:
Perhaps this is related to #487. 

Thank you for PR! It is useful for windows WMI Query, so I merge.
But as I described at #487, we will merge WMIQueryContext related functions. So please forgive us if we delete your cool commits.

PR #502, opened 2018-03-21
Use static initialisation for invoke instances, instead of init funcs

lclarkmichalek:
The order of init function execution is dependant on the order that the
source files are passed to the compiler. This causes issues when
building under other build systems, such as bazel or buck, as they are
not guarenteed to maintain the same file order as the default go tool.

This is required because other init functions in other files were referring
to this invoke variable, which caused null pointer errors when the init
functions were invoked in a different order.

### DIFF
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -54,10 +54,9 @@ type lastPercent struct {
 }
 
 var lastCPUPercent lastPercent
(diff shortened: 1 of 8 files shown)


shirou:
yes, it's reasonable. Thanks!

PR #503, opened 2018-03-26
Fix float64 casting

cshunger:


cshunger:
@shirou is this possible to get it merged and have and update?

cshunger:
@shirou can you take a look at this? This casting issue is causing swap percentage to always be 0.

shirou:
ah, thank you so much!

PR #504, opened 2018-03-26
Always calculate linux used mem as total-(free+buffered+cache)

dimrozakis:
Used memory was calculated as Total - Available.
For newer kernels (3.14+), available memory is taken from /proc/meminfo,
whereas for older kernels it is computed as free+buffered+cached.

This commit changes that behavior. Available memory is still taken from
/proc/meminfo, if available, but used memory is always computed as
total-free-buffered-cached.

This way, it matches the output of `free` for used memory (in the -/+
buffers/cache line) and other tools.

Prior to this change, I'd see a reported used memory of 600MiB whereas
free, htop and other tools would report a used memory of 1.8GiB. And
adding used, cached, buffered and free memory would leave ~1.2GiB
unaccounted for.

### DIFF
--- a/mem/mem_linux.go
+++ b/mem/mem_linux.go
@@ -70,8 +70,8 @@ func VirtualMemoryWithContext(ctx context.Context) (*VirtualMemoryStat, error) {
 	if !memavail {
 		ret.Available = ret.Free + ret.Buffers + ret.Cached
 	}
-	ret.Used = ret.Total - ret.Available
-	ret.UsedPercent = float64(ret.Total-ret.Available) / float64(ret.Total) * 100.0
+	ret.Used = ret.Total - ret.Free - ret.Buffers - ret.Cached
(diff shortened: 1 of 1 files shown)


Lomanic:
Thank you for your contribution, you are totally right, this is how it is calculated in [python's psutil too](https://github.com/giampaolo/psutil/blob/d6445f68faaa332612cebd91bfbfc56f24daed77/psutil/_pslinux.py#L449).

shirou:
@dimrozakis Neat! Thank you! 

@Lomanic Also thank you for confirming.

Aulilino:
@shirou sth. weird confused me. 

CentOS 7.4  

 total-(free+buffered+cache) and total-available are so different.  And the result print by top indicates the latter(total-available) is right.

image


image

Aulilino:
image

PR #509, opened 2018-04-08
[host] linux/mips: mips64le definition from mipsle

maddie:
Same as commit fc04d2d, but for mips64le from mipsle definition.
It would be nice if anyone can check with the correctness, since the
qemu environment for mips64/mips64le can't run Go properly due to this
issue: https://github.com/golang/go/issues/15416


shirou:
Thank you. no one confirm it, but I merge it. If someone has problem, he/she might open an issue.

Issue #515, opened 2018-04-22
Get elapsed time (etime) of a process 


dastergon:
Would be great to have function to support ps's [etime](http://linuxcommand.org/lc3_man_pages/ps1.html). It is supported by both `linux` and `darwin`.

Explanation from the man page:
```
etime  ELAPSED  elapsed time since the process was started, in the form [[dd-]hh:]mm:ss.
```
You can get an example by running ps:  `ps -eo etime`

PR #516, opened 2018-04-22
add support for etime (elapsed time)

dastergon:
This is the implementation of my feature request #515. 

According to the [ps man page](http://linuxcommand.org/lc3_man_pages/ps1.html):
```
etime  ELAPSED  elapsed time since the process was started, in the form [[dd-]hh:]mm:ss.
```

The `ElapsedTime()` function returns the elapsed time of the process indicating how long process has been running for.

Usage example:
```
package main

import (
	"fmt"

)

func main() {
	pid := os.Getpid()
	p, err  := process.NewProcess(pid)
        if err != nil {
          log.Fatalf("error: %v", err)
        } 
        etime, err := p.ElapsedTime()
        if err != nil {
          log.Fatalf("error: %v, err)
        }
        fmt.Println("PID: ", pid, "Elapsed Time: ", etime)
}
```

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -1,9 +1,12 @@
 package process
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"math"
 	"runtime"
 	"time"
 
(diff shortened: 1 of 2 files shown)


shirou:
Thanks your PR. However I agree to @Lomanic . It is easily calculated by software layer.

dastergon:
Ok, then I am closing it now. If someone believes it could be a good fit for the future, we can reopen it.

Issue #413, opened 2017-08-05
Reading stats for all running processes is inefficient


conorbranagan:
Hi,

We are using the gopsutil as part of a tool that's collecting information about all running processes. With the existing gopsutil public APIs in process_* you often end up doing the same work several times to get different data. For example calling `p.Status()` and `p.UIDs()` reads through `/proc/$pid/status` twice.

Right now we've forked and added an which adds an `AllProcesses` function (https://github.com/DataDog/gopsutil/blob/dd/process/process_linux.go#L794) that performs each proc file read once and then populates a `FilledProcess`. But ideally we'd like to stay in sync with upstream and make this work well in the normal case.

Right now some fields (`uids`, `status`) will populate in the `Process` instance (at least for linux) but it's never reused and it's not public.

So the questions are: have you given any thought to this use case? Is adding `AllProcesses` something that might be able to be accepted upstream OR could we change the existing API so it's possible to get all the fields we want with the fewest number of syscalls.

Thanks!

shirou:
Because if a user want to read just a NumFD, getting other information is waste. So I separate it.

However, I know this is not efficient if user like you want all information related to the process. Then, I agree with your `AllProcesses` and will accept if it is implemented all of the supported platforms.

And how about introduce a cache mechanism to `AllProcesses`?

conorbranagan:
Yes I you are definitely right that in most cases you're just getting a couple bits of information and pulling everything does not make sense.

That's great news that you might accept `AllProcesses` upstream. With that in mind I'll work on clean up what we have and getting it to support other platforms (so far we just have linux, mac and freebsd) and see what you think. Thanks for the quick response!

shirou:
I think `cache` can fascinate other users to use `AllProcesses`.  Perhaps #286 might be helpful.

Issue #394, opened 2017-07-06
Reading spend too much time


gonboy:
```
Microsoft Windows 7
Intel(R) Core(TM) i5-4590 CPU @ 3.30GHz  4核
4G
```
test :
```
osInfo =  100.0058ms
memInfo =  0s
cpuInfo =  1.0430596s
cpuProcInfo=  276.0158ms
cpuPerfInfo =  268.0153ms
```

shirou:
On Windows, we use WMI to get ProcInfo and PerfInfo and WMI is not so fast.

Issue #286, opened 2016-12-12
Prepopulating a Process struct with a set of fields


idank:
When calling a function on a Process instance, it will go and get the most fresh information. If the pieces of information are all in the same file, that file will be read on each call. This is really wasteful if you know what you need upfront.

It would be nice if a Process could be created such that a set of fields are cached on initialization. Obviously if we're getting fields A, B and C from file F, we should only read F once.

It would be a lot easier to pull this off if Process was an interface rather than a struct.

shirou:
Sorry for late response.

You are right. current approach is wasteful if user want to read information from same file. 
However, I think it is difficult to change this behavior from these reasons.

1. We can not expect what information actually needed. If there are only one access, cache is wasteful.
2. memory, cpu etc. changes very quick. If cache is introduced, the cache invalidation mechanism is also required. And invalidate time might be quite short time.

Do you have any idea?

idank:
Hi, I can think of several ways.

Add a new `NewProcessWithFields(pid, fields)` function. The user can then pass in `fields` and then the Process struct will prepopulate those fields. It should be smart enough to ask for the information once per set of related fields. For example, process_linux.go:fillFromStat is capable of returning 5 fields. Getting those 5 fields in the new scheme will only go to disk once.

Regarding invalidation, I think it's fine if we let the user control it. Alternatively, we can say that this way of creating a Process makes it immutable and it will not refresh data after construction.

I'd like gopsutil to have reasonable performance when getting information for a set of pids & predefined fields. If you think of how it works now, take darwin for example, for N pids we're making N calls to ps just to get a single field. If I have M fields, that's N*M calls to ps, which is very slow. If we're calling an API to get a piece of information (whether it is a shell command like ps, or a library call), we should get all the information we can so the cost of getting those M fields is as low as possible.

shirou:
How about add `CacheEnable(expire time.Duration)` to `Process` ? If it set, methods returns from cache until expire time comes.  It requires `fillFromStat` or other function changes but no API change.
I use the word `Cache` to explicitly it is a cache. So user understand the invalidation problem might be occurred. (intrducte `CacheInvalidate()` function if needed)

Cache can be used for other platform. So ps problem you told can be solved, I think.

What do you think?

idank:
Sounds like a good start. Can we also support the zero time.Duration so a user must call CacheInvalidate to get new values? This should avoid the need to check the current time on every call to get the value of something.

shirou:
[...]

Issue #250, opened 2016-08-25
when I List process info on win10, it running a long time(30s), cpu usage 30~40%


jindongyi011039:
when I List processes info on win10, it running a long time(30s), cpu usage 30~40%
but on linux , it just need 100ms

shirou:
To get processes on Windows, gopsutil WMI. and WMI is slow.
We plan to change other method like https://github.com/lxn/win, but actually, not started yet.

kondr1:
When you fix this?

shirou:
Sorry, I haven't have a time for a while. But, PR is always welcome!

pqant:
is there any progress about it ?  WMI is really slow..

shirou:
Implementation becomes slightly better from this issue first opened. Actually, which API is slow and/or could you give us a sample code and benchmark?

pqant:
@shirou 

Sure : 

This is the method that burst cpu usage from %2-5 to %60-80. 

`
package main

import (
	"github.com/gin-gonic/gin"
	"github.com/bamzi/jobrunner"
	"github.com/matishsiao/goInfo"
	"github.com/shirou/gopsutil/process" 
        ...
)

func FindProcessByCommandLineInfo(processInfo string) bool {
	if processInfo == "" {
		log.Fatalf("processInfo not specified!")
	}
	v, _ := process.Processes()
	for _, value := range v {
		cmdLine, _ := value.Cmdline()
		if strings.Contains(cmdLine, processInfo) {
			return true
		}
	}
	return false
}
`

shirou:
```
func BenchmarkProcesses(b *testing.B){
	for i := 0; i < b.N; i++ {
		Pids()
    }
}
func BenchmarkCmdline(b *testing.B){
	v, _ := Processes()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		v[0].Cmdline()
    }
}
func BenchmarkTimes(b *testing.B){
	v, _ := Processes()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		v[0].Times()
    }
}
```
Then get this result

```
BenchmarkProcesses-4         500           2366454 ns/op
BenchmarkCmdline-4            50          36758234 ns/op
BenchmarkTimes-4         2000000               623 ns/op
```

This means `Processes` is slow and `Cmdline` is very very slow. Eventually `GetWin32Proc` uses WMI and it is a problem. In fact, `Times` which does not use WMI is very fast.

[...]

PR #517, opened 2018-04-24
add snapshot feature for collecting process info at a point in time

charliesignalfx:
* capture all information about a process at a given point in time as a snapshot 
* de-duplicate underlying calls to fill in the snapshot
* allow a filter system for only filling the snapshot with desired information

Hi,
I am proposing a new structure called Snapshot that inherits from Process, but loads all information at the time the snapshot is created.
The goals of this addition are to:
* reduce the time it takes to collect process information
  I see a ~30% reduction ( 2.8 ms compared to 3.9 ms) in the average time to load all information for a single process.
* maintain feature parity with Process
* de-duplicate the calls to "fill" functions in Process.

There is also a filtering system so static snapshots can be taken with only a subset of process information.

This proposed solution was written to avoid making large changes to the existing Process design.  I appreciate any suggestions you may have for achieving the goals I listed above.

If this is of no interest to this project, then I ask that you consider exposing the underlying fill functions.  This would allow the snapshot feature to be developed in a separate downstream project.

Thanks!

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -230,7 +230,10 @@ func (p *Process) CPUPercentWithContext(ctx context.Context) (float64, error) {
 	if err != nil {
 		return 0, err
 	}
+	return p.calculateCPUPercent(ctx, crt_time, cput)
+}
 
+func (p *Process) calculateCPUPercent(ctx context.Context, crt_time int64, cput *cpu.TimesStat) (float64, error) {
 	created := time.Unix(0, crt_time*int64(time.Millisecond))
 	totalTime := time.Since(created).Seconds()
 	if totalTime <= 0 {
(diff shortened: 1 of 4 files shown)


Lomanic:
Thanks for your work and this good idea @charliesignalfx!

Caching process stats was discussed in #413, prepopulating a process with given fields in #286.

Your implementation is interesting, quite different than what was discussed before, but limited and Linux-only currently, it might require a bit of code duplication and even double the work to have the same API for Snapshot and Process types… for each OS gopsutil supports. Plus, it does not solve another problem we have on Windows when reading all processes stats (highly inefficient to call WMI for each process when we could want a "snapshot" of all the running processes with a single WMI call, re #394, #250).

I don't think we will include this new Snapshot type (as-is at least, too much work to re-implement on other platforms, much work to keep its API against Process API  evolutions), but we might need this code to implement some `NewProcessWithFields(pid, fields)` function (#286).

PR #519, opened 2018-04-26
[disk][linux] add label based on devicemapper name of disk

ofek:
This is a fixed up version of https://github.com/shirou/gopsutil/pull/327

Resolves https://github.com/shirou/gopsutil/issues/326

cc @shirou

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -42,6 +42,7 @@ type IOCountersStat struct {
 	WeightedIO       uint64 `json:"weightedIO"`
 	Name             string `json:"name"`
 	SerialNumber     string `json:"serialNumber"`
+	Label            string `json:"label"`
 }
 
 func (d UsageStat) String() string {
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -5,6 +5,7 @@ package disk
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os/exec"
 	"path/filepath"
 	"strconv"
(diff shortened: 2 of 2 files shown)


disk/disk_linux.go
```diff
@@ -406,6 +409,26 @@ func GetDiskSerialNumberWithContext(ctx context.Context, name string) string {
 	return ""
 }
 
+// GetLabel returns label of given device or empty string on error.
+// Name of device is expected, eg. /dev/sda
+// Supports label based on devicemapper name
+// See https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-block-dm
+func GetLabel(name string) string {
+	// Try label based on devicemapper name
+	dmname_filename := fmt.Sprintf("/sys/block/%s/dm/name", name)
```
shirou: Could you use `common.HostSys` instead of use `/sys/` ?

disk/disk.go
```diff
@@ -42,6 +42,7 @@ type IOCountersStat struct {
 	WeightedIO       uint64 `json:"weightedIO"`
 	Name             string `json:"name"`
 	SerialNumber     string `json:"serialNumber"`
+	Label            string `json:"label"`
```
shirou: I still have not know `label` is a good name or not. But other platform has some `label` even if the actual meaning is different on each platform.  Could you add some document on README at least Linux?

ofek: @shirou I added it to the readme, thanks!

ofek:
@shirou Is this ready to be merged?

h8liu:
this change does not build (on linux)

src/github.com/shirou/gopsutil/disk/disk_linux.go:428:3: cannot use dmname (type []byte) as type string in return argument

PR #527, opened 2018-05-19
mem[windows]: if no swap, just return 0 used percent default

### LINKED ISSUE #525: swap memory percent get NAN when swap memory total is zero
``` go
func SwapMemoryWithContext(ctx context.Context) (*SwapMemoryStat, error) {
	var perfInfo performanceInformation
	perfInfo.cb = uint32(unsafe.Sizeof(perfInfo))
	mem, _, _ := procGetPerformanceInfo.Call(uintptr(unsafe.Pointer(&perfInfo)), uintptr(perfInfo.cb))
	if mem == 0 {
		return nil, windows.GetLastError()
	}
	tot := perfInfo.commitLimit * perfInfo.pageSize
	used := perfInfo.commitTotal * perfInfo.pageSize
	free := tot - used
	ret := &SwapMemoryStat{
		Total:       tot,
		Used:        used,
		Free:        free,
		UsedPercent: float64(used) / float64(tot),
	}

	return ret, nil
}
```

My windows machine has no swap memory configured, so get a swap memory total is zero, which make the `UsedPercent: 0/0`, final get a NAN out of the function.

I don't known why the `procGetPerformanceInfo`  function not return an error or mem is not none.

geekerlw:
Just return 0 used percent when swap memory total is zero and the other will be the current zero.

There's no swap total or used is not a error.

fix #525 , please check it, thanks:)

### DIFF
--- a/mem/mem_windows.go
+++ b/mem/mem_windows.go
@@ -4,7 +4,6 @@ package mem
 
 import (
 	"context"
-	"errors"
 	"unsafe"
 
 	"github.com/shirou/gopsutil/internal/common"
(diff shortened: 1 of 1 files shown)


shirou:
Thank you!

Issue #524, opened 2018-05-17
Windows services 


pytimer:
Hi,
I already use gopsutil to gather vm metrics, such as memory and cpu. But now want to get windows services, but gopsutl currently not supported. I see https://github.com/giampaolo/psutil already support windows services. Whether consider this feature to the project?

shirou:
Interesting. But I don't have a time to do, currently. PR is always welcome!

pytimer:
Thanks for your reply. I don't know about more windows, but i will try to implement it.

pytimer:
@shirou 
I have implement about windows service package,  i hope you can check it if you have free.thanks

pytimer:
Hi, @shirou ,Thank you for merge this PR. :-) 

I think this issue can close.

shirou:
Thank you for your great contribution!

PR #528, opened 2018-05-21
[windows]services: add windows service feature

pytimer:
Hi, @shirou 
I write windows service package to implement like psutil win_services.

usage:
```go

import (
	"fmt"
	"log"

	"github.com/shirou/gopsutil/winservices"
)

func main() {
        name := "MySQL57"
	srv, err := winservices.NewService(name)
	if err != nil {
		log.Fatalf("get service by name error, %s", err)
	}
	if err := srv.GetServiceDetail(); err != nil {
		log.Fatalf("get service %s error, %s", name, err)
	}

       services, err := winservices.ListServices()
	if err != nil {
		log.Fatal(err)
	}
}
```

implement #524, please check it.
I don't know how to write test code for this feature, can you tell me?
thanks :)


shirou:
Great!! 

However, When I run `ListServices`, `openSCManager()` on my Windows 10 PC, it returns `Access is denied` error. Do you know the reason?

pytimer:
Thanks for your testing. :)

I don't know about this question, but i will try my best to solved it.
My pc os is windows7, so i can only test code on windows7.
I will find windows 10 PC to test it.

pytimer:
@shirou 
I test code on windows 10 pc, i used administrator account is correct, but when i used my owner user account to run, it also have your error `Access is denied`.  

I think this problem should be caused by permissions. openSCManager() will call windows api OpenSCManager, it need administrator privileges. So i think if someone want to use winservices package, it requires that application must be running under the administrator privileges. 

How do you think this problem?

danielnelson:
We have some [similar code](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/win_services/win_services.go) to this in Telegraf, but we also call [func (*Service) Config](https://godoc.org/golang.org/x/sys/windows/svc/mgr#Service.Config) to gather a few other pieces of information.  Have you considered adding some of the config info?

pytimer:
Hi, @danielnelson 

I read Telegraf win_services package code, this PR already has function QueryServiceConfig, like func (*Service) Config.

you can use gopsutil winservices like:

```go
srv, _ := winservices.NewService(name)
srv.QueryServiceConfig()
```

func QueryServiceConfig() will call func (*Service) Config and return [mgr.Config](https://godoc.org/golang.org/x/sys/windows/svc/mgr#Config). 

I will remove this PR  file `winservices/zsyscall_windows_ex.go`, because I already submit PR to [golang/sys/windows ](https://github.com/golang/sys/commit/56ad15cc2170219c9fed5b7d059e825c97588071).

Does this feature support telegraf win_services input plugin?
If you have any idea, welcome! And i change it together.

Thanks. :)

danielnelson:
Great, I think Telegraf would be covered here and we could switch over to these functions.

I wonder if it makes sense to have golang/sys/windows always use `QueryServiceStatusEx` when `Query()` is called?

pytimer:
golang/sys/windows `Query()` function now only return service CurrentState and ControlsAccepted,  `QueryServiceStatusEx` or `QueryServiceStatus` can return this value, so i am not sure if it need to change.

@danielnelson ,  WDYT?

danielnelson:
IMO, it would be nice if the golang/sys/windows `Query()` function returned the additional info in the existing `svc.Status` object.  The `QueryServiceStatus` is deprecated and the new values would be useful, it feels like it just isn't hooked up completely yet.

pytimer:
I agree with your option, i will submit a issue to golang/sys/windows about it. But now if Telegraf use `winservices` package, this package maybe still call `windows.QueryServiceStatusEx`.

If any progress about this, i will tell you. :)

shirou:
sorry for late response.  I am not sure about `golang/sys/windows` side, but I can run this PR with admin right. Can I merge this PR?

pytimer:
@shirou golang/sys/windows `Query()` does not change currently, so winservices still call `windows.QueryServiceStatusEx`.

If you think there is no problem, you can merge it. Thanks. :)

@danielnelson i create a issuse,  now gopsutil/winservices call `windows.QueryServiceStatusEx`  get more information, golang/sys/windows think if add `ProcessId` to `Query()`, it will brake existing programs. So if we want to get more information, we should call Windows function.

pytimer:
Hi, @shirou Can you merge this PR?

PR #532, opened 2018-06-01
Read disk model and serial from udev data or sysfs on Linux

jsoriano:
It reads model and serial data directly from files instead of using
`udevadm` command. This way obtaining the disk serial number doesn't
depend on command execution, and can be also possible even if udev or
udevadm are not available.

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -3,10 +3,11 @@
 package disk
 
 import (
+	"bufio"
+	"bytes"
 	"context"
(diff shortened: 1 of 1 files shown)


Lomanic:
Thanks a lot for this awesome contribution!

jsoriano:
@Lomanic thanks for merging it :) are you planning to do a release soon?

shirou:
we usually release end of month.

PR #537, opened 2018-06-17
[host][windows] Remove (slow) WMI calls, rely on registry or win32 API to get OS infos

Lomanic:
As in python psutil, we rely on [`GetTickCount64`](https://msdn.microsoft.com/library/windows/desktop/ms724411(v=vs.85).aspx) on Vista+ and [`GetTickCount`](https://msdn.microsoft.com/library/windows/desktop/ms724408(v=vs.85).aspx) on WinXP  to get uptime (and Bootup Time). It should be noted that GetTickCount overflows to 0 after 49.7 days.

We use RtlGetVersion to circumvent the limitations of GetVersionEx on Win8.1 and up: GetVersionEx returns OS version as Win8 on Win8.1+ unless you specify in executable manifest that you are compatible with Win8.1 (and Win10).  RtlGetVersion doesn't lie and returns the proper version.

I couldn't find a way to get the exact same thing as the [`Win32_OperatingSystem.Caption`](https://msdn.microsoft.com/en-us/library/aa394239(v=vs.85).aspx) outside WMI, so we now pull `HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProductName` to get a similar "friendly name". This one is not localized (on a french locale it is _Microsoft Windows 7 Édition Familiale Premium_ whereas this reg key is _Windows 7 Home Premium_). The service pack version (found in adjacent registry key) is now appended to the `platform`.

The `getMachineGuid` function was reworked to properly use `golang.org/x/sys/windows/registry` instead of manually calling the win32 RegOpenKeyEx/RegQueryValueEx functions.


Lomanic:
Can someone check that the `platform` var is properly set on Windows 8.1/10? I've read that it might lie as Windows 8 and I don't have these versions to test.

Lomanic:
> The getMachineGuid function was reworked to properly use golang.org/x/sys/windows/registry instead of manually calling the win32 RegOpenKeyEx/RegQueryValueEx functions.

I will rollback last commit as per https://github.com/shirou/gopsutil/pull/312#issuecomment-277422612

shirou:
Sorry for late response. I can check `platform` on Windows 10. but could you solve the conflict? Thank you!

Lomanic:
The merge conflict was resolved @shirou. Thanks for testing.

shirou:
Thanks. The variable is `Windows 10 Pro`.

Lomanic:
Excellent, merging then.

Issue #435, opened 2017-10-06
How to get serial number of disk on windows?


onkarsutar:
For linux there is a func to get HDD Serial number **GetDiskSerialNumber**.
But this method is missing in windows.
Is there any other way to get serial number?

shirou:
Sorry, I don't know about windows. If you find, PR is always welcome!

PR #541, opened 2018-06-22
Implement disk.GetDiskSerialNumber for Windows via WMI

sify21:
Fix [shirou/gopsutil#435](https://github.com/shirou/gopsutil/issues/435). Parameter should be drive letter, so it can be used in IOCountersStat directly. See last part of https://msdn.microsoft.com/en-us/library/windows/desktop/aa394592(v=vs.85).aspx.

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -32,13 +32,11 @@ type Win32_PerfFormattedData struct {
 	AvgDisksecPerRead       uint64
 	AvgDisksecPerWrite      uint64
 }
-type Win32_DiskDrive struct {
-	DeviceID         string
-	FirmwareRevision string
-	InterfaceType    string
-	SerialNumber     string
+type win32_DiskDrive struct {
+	DeviceID     string
+	SerialNumber string
 }
-type Win32_DiskPartition struct {
+type win32_DiskPartition struct {
 	DeviceID string
 }
 
@@ -187,8 +185,8 @@ func GetDiskSerialNumber(name string) string {
 }
 
 func GetDiskSerialNumberWithContext(ctx context.Context, name string) string {
-	var diskPart []Win32_DiskPartition
-	var diskDrive []Win32_DiskDrive
+	var diskPart []win32_DiskPartition
+	var diskDrive []win32_DiskDrive
 	err := common.WMIQueryWithContext(ctx, "Associators of {Win32_LogicalDisk.DeviceID='"+name+"'} where AssocClass=Win32_LogicalDiskToPartition", &diskPart)
 	if err != nil || len(diskPart) <= 0 {
 		return ""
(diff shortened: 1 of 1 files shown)


sify21:
Tested on windows server 2016. The result is right, but slow.

disk/disk_windows.go
```diff
@@ -32,6 +32,15 @@ type Win32_PerfFormattedData struct {
 	AvgDisksecPerRead       uint64
 	AvgDisksecPerWrite      uint64
 }
+type Win32_DiskDrive struct {
+	DeviceID         string
+	FirmwareRevision string
+	InterfaceType    string
+	SerialNumber     string
+}
+type Win32_DiskPartition struct {
```
Lomanic: Please don't export these types (I know there are similarly exported wmi types in other gopsutil packages but that's a mistake as Windows-only and thus should not be part of the public API).

shirou:
Thank you! This works on my Windows 10 environment.

Lomanic:
I wouldn't have merged this PR @shirou, on Windows a letter C:, D:, etc is a disk partition and not a physical disk (\\.\PHYSICALDRIVEX is), see my commit message at https://github.com/Lomanic/gopsutil/commit/8905b9265323166baad47d6b9014c161df93f233.

Plus, this result should be compared to the one found on Linux, see https://stackoverflow.com/q/27297348/ (that's why I did not create a PR for my commit).

shirou:
ahhh.. I'm sorry. Could you revert this merge ? > @Lomanic

shirou:
I have reverted this PR. sorry.

sify21:
Hi @Lomanic , to your first question, I did this on purpose so that the serial number can be returned with iocounters, like that on linux (/boot/efi, /, /home also not physical disk). Otherwise users of this library will have to find the physical disk id related to the iocounter's name themselves.
As to the second question, This method is supposed to be used on windows, so I think it's ok if all windows system do the reverse thing.

Lomanic:
I think we should have the following:

- `GetDiskSerialNumber` on Windows should accept both `\.\PHYSICALDRIVEX` and `X:` paths (there is `\Device\HarddiskVolumeX\` too, let's skip it for the moment, we would have to port [this function](https://github.com/giampaolo/psutil/blob/ac9dccab6b038835b5d612f92cf4804ec2662c2e/psutil/_psutil_windows.c#L1370), that also will be needed to get rid of some wmi in `process` AFAIR). Please note that this is the behavior of gopsutil on Linux where `GetDiskSerialNumber` obtains the right serial number if we give it a partition `/dev/sda1` or a device `/dev/sda`.
- (optional but nice to have) a `disk.PhysicalDisks` function to return physical disks instead of partitions (with `\.\PHYSICALDRIVEX` paths on Windows). `disk.PartitionStat` includes a device field, but the user has to deduplicate results and [it's not even implemented on Windows](https://github.com/shirou/gopsutil/blob/8563278/disk/disk_windows.go#L128).
- the same serial numbers should be returned no matter the OS, if wmi inverts them on Windows, we shall reverse them back

So we might merge back this PR, but with some checks added if we query a physical drive path.

PR #544, opened 2018-07-02
Use HOST_* environment variables for getting disk serial number in Linux

jsoriano:
Code introduced in #532 was not making use of these environment variables.


Lomanic:
Good catch, thank you

Issue #164, opened 2016-02-25
replace lsof to platform dependent method


shirou:
#142 , #143 are related to using lsof to get Network connection info.

To replace lsof, we needs at least these functions.  inet_ntop porting, sysctl decode functions,  and so on are little bit heavy to implement.  I will update this issue step by step.

- [ ] Linux
 - [ ] getting inode list from processes
 - [ ] porting socket.inet_ntop to pure golang (linux struct)
 - [ ] IPv6
 - [ ] Unix Pipe
- [ ] windows
 - [ ] needs research
 - [ ] rtlIpv4AddressToStringA
 - [ ] rtlIpv6AddressToStringA
 - [ ] getExtendedTcpTable
- [ ] darwin
 - [ ] proc_pidinfo and retrieve proc_fdtype
 - [ ] inet_ntop 
- [ ] freebsd
 - [ ] sysctl net.inet.tcp.pcblist for TCP
 - [ ] sysctl net.inet.udp.pcblist for UDP 
 - [ ] sysctl net.local.stream.pcblist for socket stream
 - [ ] sysctl net.local.dgram.pcblist for socket dgram
 - [ ] porting inet_ntop to pure golang (freebsd struct)
 - [ ] porting kinfo_file

vrecan:
:+1: great to see that you are taking this on!

sparrc:
looking forward to this, thanks for your work @shirou!

shirou:
OK. now I have implemented linux version at #169. But more tests are required.

shirou:
### FreeBSD

sysctl `net.inet.tcp.pcblist` returnes `xtcpcb` struct. But since it is complicated to parse to golang struct, time should be required.

### Darwin

To get NetConnections, `proc_pidinfo()` is required and it seems to invoke by only CGO. I am thinking `net` is also use CGO or not. Or, the `net.inet.tcp.pcblist` is exists on Darwin also, the FreeBSD code may be used on Darwin, too. I will challenge after FreeBSD.

### windows

not yet...

## v2 migration

It takes a time to implement, I think prioritize v2.

NullSoldier:
Is this dead in the water? I'm seeing flakiness with using lsof where the command never finishes.

shirou:
@NullSoldier what platform do you want to replace lsof?

PR #546, opened 2018-07-03
Support for unix sockets in ConnectionsPid()

arvenil:
Currently:
```golang
	case "unix":
		return ret, common.ErrNotImplementedError
```
 With `-U` option passed to `lsof` we can get the list of unix sockets so we can change this to:
```patch
 	case "unix":
-		return ret, common.ErrNotImplementedError
+		args = []string{"-U"}
 	}
```
However I'm not really sure about some other required changes... maybe someone could help me with this PR?

Right now it works but take a look at the code. An example output below.
```go
import (
        "fmt"
        "log"

        "github.com/shirou/gopsutil/net"
)

func main() {
        pids, err := process.Pids()
        if err != nil {
                panic(err)
        }
        for _, pid := range pids {
                proc, err := process.NewProcess(pid)
                if err != nil {
                        log.Println(err)
                        continue
                }
                if procName, err := proc.Name(); err != nil {
                        log.Println(err)
                        continue
                } else {
                        if procName == "mysqld" {
                                cons, err := net.ConnectionsPid("unix", pid)
                                if err != nil {
                                        log.Fatal(err)
                                }
                                if len(cons) > 0 {
                                        fmt.Printf("mysql with PID %d is using socket: %s\n", pid, cons[0].Laddr.IP)
                                }
                        }
                }
        }
}
```
```sh
# go run main.go
mysql with PID 15612 is using socket: ->0x6be97f0b39f04933
mysql with PID 15612 is using socket: ->0x6be97f0b39f05d83
mysql with PID 15612 is using socket: /tmp/mysql.sock
mysql with PID 15612 is using socket: /tmp/mysqlx.sock
```

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -71,6 +71,7 @@ type FilterStat struct {
 }
 
 var constMap = map[string]int{
+	"unix": syscall.AF_UNIX,
 	"TCP":  syscall.SOCK_STREAM,
 	"UDP":  syscall.SOCK_DGRAM,
 	"IPv4": syscall.AF_INET,
@@ -178,8 +179,13 @@ func getIOCountersAll(n []IOCountersStat) ([]IOCountersStat, error) {
 
 func parseNetLine(line string) (ConnectionStat, error) {
 	f := strings.Fields(line)
-	if len(f) < 9 {
+	if len(f) < 8 {
 		return ConnectionStat{}, fmt.Errorf("wrong line,%s", line)
+	}
+
+	if len(f) == 8 {
+		f = append(f, f[7])
+		f[7] = "unix"
 	}
 
 	pid, err := strconv.Atoi(f[1])
@@ -199,9 +205,14 @@ func parseNetLine(line string) (ConnectionStat, error) {
 		return ConnectionStat{}, fmt.Errorf("unknown type, %s", f[7])
 	}
 
-	laddr, raddr, err := parseNetAddr(f[8])
-	if err != nil {
-		return ConnectionStat{}, fmt.Errorf("failed to parse netaddr, %s", f[8])
+	var laddr, raddr Addr
+	if f[7] == "unix" {
+		laddr.IP = f[8]
+	} else {
+		laddr, raddr, err = parseNetAddr(f[8])
+		if err != nil {
+			return ConnectionStat{}, fmt.Errorf("failed to parse netaddr, %s", f[8])
+		}
 	}
 
 	n := ConnectionStat{
--- a/net/net_unix.go
+++ b/net/net_unix.go
@@ -63,7 +63,7 @@ func ConnectionsPidWithContext(ctx context.Context, kind string, pid int32) ([]C
 	case "udp6":
 		args = append(args, "6udp")
 	case "unix":
-		return ret, common.ErrNotImplementedError
+		args = []string{"-U"}
 	}
 
 	r, err := common.CallLsofWithContext(ctx, invoke, pid, args...)


shirou:
Thank you for your awesome contribution! I can confirm it works like a charm on my FreeBSD environment!

arvenil:
But I have doubts now. I think it may not work on other systems as it turns out `lsof` output can be different on different os.
There is `-F` option for `lsof` and `OUTPUT FOR OTHER PROGRAMS` description how to parse it:
```
OUTPUT FOR OTHER PROGRAMS
       When the -F option is specified, lsof produces output that is suitable for processing by another program - e.g, an awk or Perl script, or a C program.

       Each unit of information is output in a field that is identified with a leading character and terminated by a NL (012) (or a NUL (000) if the 0 (zero) field identifier character is specified.)  The data of the field follows immediately after the field identification character and extends to the field terminator.

       It is possible to think of field output as process and file sets.  A process set begins with a field whose identifier is `p' (for process IDentifier (PID)).  It extends to the beginning of the next PID field or the beginning of the first file set of the process, whichever comes first.  Included in the process set are fields that identify the command, the
       process group IDentification (PGID) number, the task (thread) ID (TID), and the user ID (UID) number or login name.

       A file set begins with a field whose identifier is `f' (for file descriptor).  It is followed by lines that describe the file's access mode, lock state, type, device, size, offset, inode, protocol, name and stream module names.  It extends to the beginning of the next file or process set, whichever comes first.

       When the NUL (000) field terminator has been selected with the 0 (zero) field identifier character, lsof ends each process and file set with a NL (012) character.

       Lsof always produces one field, the PID (`p') field.  All other fields may be declared optionally in the field identifier character list that follows the -F option.  When a field selection character identifies an item lsof does not normally list - e.g., PPID, selected with -R - specification of the field character - e.g., ``-FR'' - also selects the list-
       ing of the item.

       It is entirely possible to select a set of fields that cannot easily be parsed - e.g., if the field descriptor field is not selected, it may be difficult to identify file sets.  To help you avoid this difficulty, lsof supports the -F option; it selects the output of all fields with NL terminators (the -F0 option pair selects the output of all fields with
       NUL terminators).  For compatibility reasons neither -F nor -F0 select the raw device field.

       These are the fields that lsof will produce.  The single character listed first is the field identifier.

            a    file access mode
            c    process command name (all characters from proc or
                 user structure)
            C    file structure share count
            d    file's device character code
            D    file's major/minor device number (0x<hexadecimal>)
            f    file descriptor (always selected)
            F    file structure address (0x<hexadecimal>)
            G    file flaGs (0x<hexadecimal>; names if +fg follows)
            g    process group ID
            i    file's inode number
            K    tasK ID
            k    link count
            l    file's lock status
            L    process login name
            m    marker between repeated output
            n    file name, comment, Internet address
            N    node identifier (ox<hexadecimal>
            o    file's offset (decimal)
            p    process ID (always selected)
            P    protocol name
            r    raw device number (0x<hexadecimal>)
            R    parent process ID
            s    file's size (decimal)
            S    file's stream identification
            t    file's type
            T    TCP/TPI information, identified by prefixes (the
                 `=' is part of the prefix):
                     QR=<read queue size>
                     QS=<send queue size>
                     SO=<socket options and values> (not all dialects)
                     SS=<socket states> (not all dialects)
                     ST=<connection state>
                     TF=<TCP flags and values> (not all dialects)
                     WR=<window read size>  (not all dialects)
                     WW=<window write size>  (not all dialects)
                 (TCP/TPI information isn't reported for all supported
                   UNIX dialects. The -h or -? help output for the
                   -T option will show what TCP/TPI reporting can be
                   requested.)
            u    process user ID
            z    Solaris 10 and higher zone name
            Z    SELinux security context (inhibited when SELinux is disabled)
            0    use NUL field terminator character in place of NL
            1-9  dialect-specific field identifiers (The output
                 of -F? identifies the information to be found
                 in dialect-specific fields.)

       You can get on-line help information on these characters and their descriptions by specifying the -F?  option pair.  (Escape the `?' character as your shell requires.)  Additional information on field content can be found in the OUTPUT section.

       As an example, ``-F pcfn'' will select the process ID (`p'), command name (`c'), file descriptor (`f') and file name (`n') fields with an NL field terminator character; ``-F pcfn0'' selects the same output with a NUL (000) field terminator character.

       Lsof doesn't produce all fields for every process or file set, only those that are available.  Some fields are mutually exclusive: file device characters and file major/minor device numbers; file inode number and protocol name; file name and stream identification; file size and offset.  One or the other member of these mutually exclusive sets will appear
       in field output, but not both.

       Normally lsof ends each field with a NL (012) character.  The 0 (zero) field identifier character may be specified to change the field terminator character to a NUL (000).  A NUL terminator may be easier to process with xargs (1), for example, or with programs whose quoting mechanisms may not easily cope with the range of characters in the field  output.
       When the NUL field terminator is in use, lsof ends each process and file set with a NL (012).

       Three aids to producing programs that can process lsof field output are included in the lsof distribution.  The first is a C header file, lsof_fields.h, that contains symbols for the field identification characters, indexes for storing them in a table, and explanation strings that may be compiled into programs.  Lsof uses this header file.

       The second aid is a set of sample scripts that process field output, written in awk, Perl 4, and Perl 5.  They're located in the scripts subdirectory of the lsof distribution.

       The third aid is the C library used for the lsof test suite.  The test suite is written in C and uses field output to validate the correct operation of lsof.  The library can be found in the tests/LTlib.c file of the lsof distribution.  The library uses the first aid, the lsof_fields.h header file.
```
Seems like keybase did parser for this https://github.com/keybase/client/tree/master/go/lsof so if someone complains that it doesn't work for other os then probably `lsof -F` should be used for all calls and such parsing should be introduced - but that's another improvement you can make over current `lsof` output parsing - or just finally get rid of `lsof` https://github.com/shirou/gopsutil/issues/164 ;)

Issue #472, opened 2017-12-18
Temperature on Windowsz


LeMoussel:
Hi,

On Windows (Intel CPU) `host.SensorsTemperatures()` return `[]`

Lomanic:
It returns a `common.NotImplementedError` too.
Help to implement this on Windows with WMI or (preferably) win32 API very much appreciated 😉

pytimer:
@Lomanic 

I want to use wmi implement this issue.

I use https://github.com/stackexchange/wmi on my demo, it works. But when i add code to gopsutil, use common.WMIQueryWithContext(), i always return error like below.

```
context deadline exceeded
exit status 1
```

Is it call this method wrong?

pytimer:
@LeMoussel @Lomanic , i already implement it. 

I hope you can check it, and if any problem about it, welcome to tell me, i will fix it.

Thanks.

Lomanic:
For future reference, it might be doable to use [Open Hardware Monitor](http://openhardwaremonitor.org/) (via [this SO answer](https://stackoverflow.com/a/9564862), by loading its DLL) so we don't rely on the slow and unreliable wmi requests implemented in #550.

LeMoussel:
Open Hardware Monitor API documentation is poor. For GPU you can see code in 
https://github.com/openhardwaremonitor/openhardwaremonitor/blob/master/Hardware/ATI/ATIGPU.cs 
https://github.com/openhardwaremonitor/openhardwaremonitor/blob/master/Hardware/Nvidia/NvidiaGPU.cs

Lomanic:
> But if we use Open Hardware Monitor to implement it, all user or application should include this DLL, does it increase the complexity of application deployment?

Not necessarily if we use something like [github.com/gobuffalo/packr/v2](https://github.com/gobuffalo/packr/tree/master/v2) which would embed the DLL in binaries. To be tested.

danielnelson:
It seems like this could complicate the build process especially when cross compiling.  Also, I believe users of this library would need to provide the source code for Open Hardware Monitor when distributing their applications.

Lomanic:
As I see it, packr converts an asset to a .go source code file with the asset's contents serialized to a `[]byte` variable. We could commit this (probably huge) file and gopsutil users would be able to easily (cross-)compile without the `packr build` step. I hope the binary size for Windows users using the `host` package will not explode (the DLL is just a few kB).

Skyline404:
[...]

PR #550, opened 2018-07-09
[host] Implement windows temperature use wmi

pytimer:
implement #472 .

I want to use windows api to implement it, but i can not found any idea about it.
So i use WMI get windows temperature.

`host.SensorsTemperatures()` return value:

`[{"sensorKey":"ACPI\\ThermalZone\\TZ00_0","sensorTemperature":27.85} {"sensorKey":"ACPI\\ThermalZone\\TZ01_0","sensorTemperature":29.85}]`

Welcome to check it, and if it have any problem, please tell me. Thanks. :-)

### DIFF
--- a/host/host_windows.go
+++ b/host/host_windows.go
@@ -5,6 +5,7 @@ package host
 import (
 	"context"
 	"fmt"
+	"math"
 	"os"
 	"runtime"
 	"strings"
@@ -13,6 +14,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/StackExchange/wmi"
 	"github.com/shirou/gopsutil/internal/common"
 	process "github.com/shirou/gopsutil/process"
 	"golang.org/x/sys/windows"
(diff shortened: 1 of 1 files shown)


host/host_windows.go
```diff
@@ -261,3 +287,10 @@ func KernelVersionWithContext(ctx context.Context) (string, error) {
 	_, _, version, err := PlatformInformation()
 	return version, err
 }
+
+type MSAcpi_ThermalZoneTemperature struct {
```
Lomanic: Don't export this type. There are similarly exported types but this is a mistake as Windows-only, while gopsutil aims to provide a platform-agnostic API.

pytimer: >Don't export this type. 

MSAcpi_ThermalZoneTemperature struct only used for WMI . I found other package use WMI also definded like it. 

>There are similarly exported types but this is a mistake as Windows-only

I don't understand what's mean similarly exported types, can you tell me detail, thanks.

shirou: Lomanic is correct. We want to gopsutil platform independent. If this API exported only for Windows, it becomes Windows only.

shirou: Actually, we already has exported structs on platform dependent part, but these are fault and should be fixed.

pytimer: I update `MSAcpi_ThermalZoneTemperature` to `msacpi_ThermalZoneTemperature`, but `WMI` query error, so i think this struct not modify. If you know how to change, can you tell me? I will change it. Thanks.

Lomanic: `msAcpi_ThermalZoneTemperature` works properly on Windows 7 (needs admin rights though). I'm merging and fixing this unwanted export. Thanks for your contribution @pytimer.

pytimer: Thanks for your fix.

shirou:
I'm really thank you to  @pytimer and @Lomanic 
 contribution!

Issue #552, opened 2018-07-12
TestVirtual_memory failure


elboulangero:
```
=== RUN   TestVirtual_memory
--- FAIL: TestVirtual_memory (0.00s)
	mem_test.go:29: 
			Error Trace:	mem_test.go:29
			Error:      	Not equal: 
			            	expected: 0x3d93b5000
			            	actual  : 0x3dc0ef000
			Test:       	TestVirtual_memory
			Messages:   	Total should be computable from available + used: {"total":16529444864,"available":12705124352,"used":3871744000,"usedPercent":23.423315373599713,"free":2846973952,"active":9378164736,"inactive":3537682432,"wired":0,"buffers":308453376,"cached":9502273536,"writeback":0,"dirty":46673920,"writebacktmp":0,"shared":475365376,"slab":543125504,"pagetables":60813312,"swapcached":84303872,"commitlimit":25318756352,"committedas":11271778304,"hightotal":0,"highfree":0,"lowtotal":0,"lowfree":0,"swaptotal":17054035968,"swapfree":16142458880,"mapped":439349248,"vmalloctotal":35184372087808,"vmallocused":0,"vmallocchunk":0,"hugepagestotal":0,"hugepagesfree":0,"hugepagesize":2097152}
=== RUN   TestSwap_memory
--- PASS: TestSwap_memory (0.00s)
=== RUN   TestVirtualMemoryStat_String
--- PASS: TestVirtualMemoryStat_String (0.00s)
=== RUN   TestSwapMemoryStat_String
--- PASS: TestSwapMemoryStat_String (0.00s)
FAIL
FAIL	github.com/shirou/gopsutil/mem	0.003s
```

Lomanic:
Fixed in #553

PR #553, opened 2018-07-12
Fix TestVirtual_memory #552

elboulangero:
Based on https://stackoverflow.com/a/30772733/776208

Signed-off-by: Arnaud Rebillout <arnaud.rebillout@collabora.com>

### DIFF
--- a/mem/mem_test.go
+++ b/mem/mem_test.go
@@ -26,8 +26,8 @@ func TestVirtual_memory(t *testing.T) {
 	assert.True(t, v.Available > 0)
 	assert.True(t, v.Used > 0)
 
-	assert.Equal(t, v.Total, v.Available+v.Used,
-		"Total should be computable from available + used: %v", v)
+	assert.Equal(t, v.Total, v.Used + v.Free + v.Buffers, v.Cached,
+		"Total should be computable (used + free + buffers + cached): %v", v)
 
 	assert.True(t, v.Free > 0)
 	assert.True(t, v.Available > v.Free,


mem/mem_test.go
```diff
@@ -26,8 +26,8 @@ func TestVirtual_memory(t *testing.T) {
 	assert.True(t, v.Available > 0)
 	assert.True(t, v.Used > 0)
 
-	assert.Equal(t, v.Total, v.Available+v.Used,
-		"Total should be computable from available + used: %v", v)
+	assert.Equal(t, v.Total, v.Used + v.Free + v.Buffers, v.Cached,
```
Lomanic: `s/v.Buffers, v.Cached/v.Buffers + v.Cached/`, but tested OK on Linux otherwise.

shirou: Good catch. I will fix after merge.

shirou:
Thank you so much!

mem/mem_test.go
Lomanic: This test is now broken on Windows (it wasn't before this merge), because v.Free+v.Buffers+v.Cached equals 0.

elboulangero: Oh sorry, I have no idea how this thing works on Windows. Looks like the formula to get `v.Total` varies with the OS...

shirou: So, is this test itself wrong because platform dependent? if so, we'd better to delete this test.

elboulangero: I believe that the fix I provided is the right thing for Linux. I have no idea about Window. Not even talking about MacOS :) Looks like something platform dependent indeed.

Lomanic: I have a PR ready that makes the test branch out against `runtime.GOOS` so the test passes on Windows (that's not the only issue for this test on Windows), I will submit it for review in the upcoming days. 

I don't even have a darwin machine anymore to check if the test passes on this platform.

PR #556, opened 2018-07-18
[mem][windows] Fix TestVirtual_memory test

Lomanic:
Following https://github.com/shirou/gopsutil/pull/553#discussion_r202643975, here is the fix for the failing test on Windows. Don't hesitate to criticize, `runtime.GOOS == "windows" || v.Free > 0` is quite ugly for example and might not be merged as-is.


shirou:
Thank you ! if other platform should be checked different way, change to use `switch` at that time.

PR #557, opened 2018-07-23
[disk][unix] Fix #555 Unescape escaped sequences in fstab path in disk.Usage

### LINKED ISSUE #555: Issue with Mountpoint field of PartitionStat if mounting path has space in it
On Linux, if a partition mount path has a space in it, then the string in the PartitionStat returned by disk.Partitions() replaces that space with the string "\040" (not the character).

Lomanic:
Fix #555

### DIFF
--- a/disk/disk_unix.go
+++ b/disk/disk_unix.go
@@ -4,6 +4,7 @@ package disk
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


shirou:
Great! Thank you!

PR #559, opened 2018-07-25
Calculating pss/uss from /proc/<pid>/smaps

ronklein:
Added pss and uss metrics.
Fixed issue in MemoryMapsWithContext. Parsing of /proc/<pid>/smaps failed when field "Name:" exists.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -46,21 +46,33 @@ func (m MemoryInfoExStat) String() string {
 }
 
 type MemoryMapsStat struct {
-	Path         string `json:"path"`
-	Rss          uint64 `json:"rss"`
-	Size         uint64 `json:"size"`
-	Pss          uint64 `json:"pss"`
-	SharedClean  uint64 `json:"sharedClean"`
-	SharedDirty  uint64 `json:"sharedDirty"`
-	PrivateClean uint64 `json:"privateClean"`
-	PrivateDirty uint64 `json:"privateDirty"`
-	Referenced   uint64 `json:"referenced"`
-	Anonymous    uint64 `json:"anonymous"`
-	Swap         uint64 `json:"swap"`
+	Path           string `json:"path"`
+	Rss            uint64 `json:"rss"`
+	Size           uint64 `json:"size"`
+	Pss            uint64 `json:"pss"`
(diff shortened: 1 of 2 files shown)


Lomanic:
We can't accept your additions as they are Linux-only currently, try to run `GOOS=windows go test github.com/shirou/gopsutil/process` to see undefined types/methods errors at compile time. 

Please document what these values represent and provide a source re: how you calculate them.

ronklein:
Moved specific linux test, to a dedicated test file. 
USS/PSS calculation is described in https://lwn.net/Articles/230975/. Added comments to emphasis that.

Lomanic:
The `MemoryInfoSmaps` type and function are still Linux-only, gopsutil is platform-agnostic and provides the same public API (types, methods) across all the supported platforms, moving around this new test to make it Linux-only does not fix anything. The `GOOS=windows go test ` command was here to underline the undefined type/method errors on this platform (any platform not Linux). 

Also, do we need a new type `MemoryInfoSmaps` where PSS is already found in `MemoryMapsStat`? Wouldn't it be sensible  to just add the new `USS` field to `MemoryMapsStat`?

ronklein:
Currently MemoryMapsWithContext is called with a boolean "grouped" variable. But this variable is not used. Maybe the intention was to use this flag to decide whether to sum up all memory maps into a single group? If so, then we can add USS and use the grouped flag to decide to either return all memory maps, or a single aggregate memory map (sum of each memory map type).

shirou:
`MemoryMaps` is a port of [psutil memory_map](http://psutil.readthedocs.io/en/latest/#psutil.Process.memory_maps).  This (perhaps only this) method is allowed to use platform dependent struct, because psutil does. 

`grouped` is not implemented, but it is just because my laziness. I will appreciate if you would make a PR    with PSS and grouped feature.

ronklein:
Sure, i will make a PR for it.
I will also make a PR for implementing [memory_full_info()](http://psutil.readthedocs.io/en/latest/#psutil.Process.memory_full_info), which returns both USS and PSS.

tagirb:
Just noticed, that you've been up to the same stuff. Here's a PR that adds grouping of MemoryMaps, please have a look: https://github.com/shirou/gopsutil/pull/567.

PR #561, opened 2018-07-26
Fix undefined process.Processes when building telegraf on Solaris-based system

szaydel:
I attempted to build `telegraf` today on `OmniOS`, which is a `Solaris` derived system, but based on the `Illumos` kernel, and ran into this issue. A bit of digging revealed that nothing builds that would satisfy this requirement. As such, this hack happened.
```
# github.com/influxdata/telegraf/plugins/inputs/procstat
plugins/inputs/procstat/native_finder.go:24:16: undefined: process.Processes
plugins/inputs/procstat/native_finder_notwindows.go:18:16: undefined: process.Processes
plugins/inputs/procstat/native_finder_notwindows.go:43:16: undefined: process.Processes
```
It seems like there's not really much there to support process information on `Solaris` and friends, and I am wondering if there is any demand, and if so, I have done some work to expose but a small subset of what seems to be made available from other operating systems, but perhaps if there is interest, I can at least add a few of those bits as time permits and maybe enhance from there.

Thanks, Sam.

### DIFF
--- a/process/process_fallback.go
+++ b/process/process_fallback.go
@@ -36,6 +36,14 @@ func PidsWithContext(ctx context.Context) ([]int32, error) {
 	return []int32{}, common.ErrNotImplementedError
 }
 
+func Processes() ([]*Process, error) {
+	return nil, common.ErrNotImplementedError
+}
+
+func ProcessesWithContext(ctx context.Context) ([]*Process, error) {
+	return nil, common.ErrNotImplementedError
+}
+
 func NewProcess(pid int32) (*Process, error) {
 	return nil, common.ErrNotImplementedError
 }


Lomanic:
You might be interested in https://github.com/shirou/gopsutil/pull/378 that adds some support for Illumos. 

About your PR, I don't have any objections, it would fix compilation for any platform not currently implemented, the error will appear at runtime instead.

szaydel:
This sounds good. I did look at that PR, and I see that there's nothing for process, which is what `telegraf` actually demands. I think I could add memory bits if there's interest, at least 64bit support initially. If you don't mind pulling in this PR, it will make it easier for me to build `telegraf`. :)

Thanks!

shirou:
I think this PR is not only Solaris-based system but also other systems. And we can check `err` to avoid runtime error after build success. Thank you for your contribution!

szaydel:
Glad to contribute in a tiny way. I will look into getting Solaris-centric process information and making it available through `psutil`. It would be nice to have.

PR #563, opened 2018-07-30
Fixing calculation of UsedPercent.

tomekkolo:
Now it is calculated based of available disk space for user,
not including reserved space for root. This is compatible
with psutil and df command.


shirou:
Great! Thank you!

PR #566, opened 2018-08-07
Add mem(laundry) numbers on FreeBSD.

zyun-i:
After FreeBSD 11.1, mem number of `cached` replaced by `laundry`.

https://reviews.freebsd.org/D8302

And Fix FreeBSD total memory. Like this.

FreeBSD: fix total memory giampaolo/psutil@bd9a58b giampaolo/psutil@bd9a58b

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -42,6 +42,10 @@ type VirtualMemoryStat struct {
 	Inactive uint64 `json:"inactive"`
 	Wired    uint64 `json:"wired"`
 
+	// FreeBSD specific numbers:
+	// https://reviews.freebsd.org/D8467
+	Laundry  uint64 `json:"laundry"`
+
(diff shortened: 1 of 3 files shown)


Lomanic:
Thanks for the fixed way to calculate total memory as in python psutil, but I don't find any trace of `laundry` in https://github.com/giampaolo/psutil/blob/f4c6b02461cc89a7f9a98a164d346b4dfb72bee0/psutil/arch/freebsd/specific.c, won't it break on previous versions (lower than 11.1) of FreeBSD?

shirou:
Thank you so much! I confirmed this cool PR works on my FreeBSD-11 environment. And I agree to add FreeBSD specific `laundary` to memstat struct, because Linux already has.

And, this PR contains a version check so I think it can run on FreeBSD-10! cool!!

PR #577, opened 2018-09-03
Calucation for Windows Total CPU utlilization changed

adamiklukasz:
This PR changes the way total CPU metrics are calculated on Windows. Previous version didn't provide good result (commit from 4 years: "implements CPU info total on Windows. but it may be not correct.").

New version is based on best answer from following topic:
https://stackoverflow.com/questions/23143693/retrieving-cpu-load-percent-total-in-windows-with-c

Generally new code rely on the fact that values from GetSystemTimes are relative. To calculate current CPU usage you need two separate calls to that API. That's why some global variables was introduced.

There is also a question about how to handle following situations:
1) First call to GetSystemTimes() - in my case I'm doing this is init(), totalCPUTimes() returns empty array
2) How to handle Consecutive calls to GetSystemTimes() withing small amount of time - in this case values returned from WinAPI may be that same as for previous call. In that case calculation of CPU is not possible. In that case I also return empty array.

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -38,6 +38,18 @@ type Win32_PerfFormattedData_Counters_ProcessorInformation struct {
 	DPCRate               uint32
 }
 
+type systemTimes struct {
+	idleTime   uint64
+	kernelTime uint64
+	userTime   uint64
+}
+
+var lastSystemTimes systemTimes
+
+func init() {
+	totalCPUTimes()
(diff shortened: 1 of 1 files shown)


adamiklukasz:
Added @lootek

lootek:
cc @tomekkolo

shirou:
I am not familiar with WIndows. Could someone join this PR?

PR #581, opened 2018-09-11
get docker container cpu usage from cpuacct.usage

CMGS:
This PR for:

* Get cgroup cpu usage from cpuacct.usage, it's a nanoseconds data, so I trans it to seconds.
* Make ```CgroupCPUWithContext``` return same unit like ```cpu.Times```. I saw cpu.Times will trans jiffies to seconds, so here should divide ```CPUTick```.

After this change, we can calculate container CPU percent in host view and container view like this:

```
interval := Seconds
containerCPUStat1 := CgroupCPUDocker(cid)
hostCPUStat1 := cpu.Times(false)[0]
time.sleep(interval)
hostCPUStat2 := cpu.Times(false)[0]
containerCPUStat2 := CgroupCPUDocker(cid)

percentInHostView := (containerCPUStat2.Usage - containerCPUStat1.Usage) / (hostCPUStat2.Total() - hostCPUStat1.Total())

percentInContainerView := percentInHostView * CPUNumbers
```

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -13,7 +13,7 @@ import (
 	"github.com/shirou/gopsutil/internal/common"
 )
 
-var cpu_tick = float64(100)
+var CPUTick = float64(100)
 
 func init() {
 	getconf, err := exec.LookPath("/usr/bin/getconf")
@@ -25,7 +25,7 @@ func init() {
 	if err == nil {
 		i, err := strconv.ParseFloat(strings.TrimSpace(string(out)), 64)
 		if err == nil {
-			cpu_tick = float64(i)
+			CPUTick = i
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


shirou:
Thank you for your PR. but it seems this PR breaks compatibility about `CgroupCPU`. (`*cpu.TimesStat` -> `CgroupCPUStat`) We don't want to change API compatiblity, so it should be merge on v3.

or, because `CgroupCPUUsage` is just added and not break compatibility, we can merge it if you separate this PR.

CMGS:
@shirou So, could I add Usage into cpu.TimesStat? In this way, ```CgroupCPU``` can return ```*cpu.TimesStat```

shirou:
No. `cpuacct`. If you add only `CgroupCPUUsage` users who use cgroup can get usage without breaking. compatibility. No need to add usage to `CgroupCPU`, I think.

CMGS:
OK, let me think about it...

shirou:
Sorry for late response and thank you for your contribution!

PR #583, opened 2018-09-13
net: use multi-value case clause instead of fallthrough

Quasilyte:
Found using https://go-critic.github.io/overview#emptyFallthrough-ref


Lomanic:
Didn't know about go-critic, thanks 👍

PR #590, opened 2018-09-21
get docker container cpu usage from cpuacct.usage

CMGS:
This is for v3


shirou:

PR #595, opened 2018-10-19
add CgroupCPUUsageDocker just like CgroupCPUDocker

liu4480:
In line with CgroupCPUDocker, because I used :

CgroupCPUDockerUsageWithContext(context.Background(), containerid)

But I do not want to import extra package like context.

Feel free to close it if you think this is not necessary


shirou:
Oh, you are right! Thank you!

liu4480:
thanks for merging

PR #602, opened 2018-11-06
CPU(win): pass context through the Times

requilence:
Context is missed through [TimesWithContext](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_windows.go#L54)

As a result the default timeout is set([3sec](https://github.com/shirou/gopsutil/blob/master/internal/common/common_windows.go#L119)) for the WMI call even if other timeout is specified via Context

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -46,40 +46,19 @@ type Win32_PerfFormattedData_PerfOS_System struct {
 
 // Times returns times stat per cpu and combined for all CPUs
 func Times(percpu bool) ([]TimesStat, error) {
-	return TimesWithContext(context.Background(), percpu)
-}
-
-func TimesWithContext(ctx context.Context, percpu bool) ([]TimesStat, error) {
 	if percpu {
 		return perCPUTimes()
 	}
 
-	var ret []TimesStat
-	var lpIdleTime common.FILETIME
-	var lpKernelTime common.FILETIME
-	var lpUserTime common.FILETIME
-	r, _, _ := common.ProcGetSystemTimes.Call(
-		uintptr(unsafe.Pointer(&lpIdleTime)),
-		uintptr(unsafe.Pointer(&lpKernelTime)),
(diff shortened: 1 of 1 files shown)


Lomanic:
Thanks for this PR, unfortunately this was superseded by #601.

Re: code-style, you are redefining `cpu.Times()` instead of just being a dumb call to `cpu.TimesWithContext()` with a default context.Background() and the same parameters, which conflicts with the rest of the lib. In the future we are going to only have defined `*WithContext` functions in OS-specific files and the "withoutContext" functions will be moved in the common non-OS-specific files (here, `cpu/cpu.go`).

Re: making the call to `common.ProcGetSystemTimes` context-aware as in your newly defined `allCPUTimesWithContext` function, I'm not convinced of the usefulness, this is the last function called and Win32 syscalls are pretty fast (a few milliseconds on my old machine, in this case the measured maximum is 4.0002ms).

If you can convince me of the usefulness of making `common.ProcGetSystemTimes` context-aware you should rebase your PR against the current master and not redefine `cpu.Times()`.

Lomanic:
Thanks for your PR, unfortunately it was superseded by #601 which fixes the context forwarding.

About code-style, you redefined `cpu.Times()` instead of a dumb call to `cpu.TimesWithContext()` with a default context.Background(), like how it is done in the rest of the library: in the (near) future we will have all the "*WithoutContext" functions defined the same way for all OSes (which should be the case currently) in the non-OS-specific source files (in this case `cpu/cpu.go`) where this will not be possible, and with only `*WithContext` functions defined in OS-specific source files.

Besides forwarding context to the WMI call, you made context-aware the call to `common.ProcGetSystemTimes`: Win32 syscalls are usually pretty fast (max time measured here is 4 milliseconds on my old machine) and AFAIK there is no such other handling in gopsutil for Win32 syscalls, so I'm not convinced of its usefulness here.

Lomanic:
No news from PR opener, closing.

Issue #350, opened 2017-04-12
HostID or HardwareID?


robmuh:
Although there is a [long running history of figuring out what to name `boot_id` and why it goes in the `proc` namespace](https://lists.linuxfoundation.org/pipermail/containers/2012-September/030194.html) in the first place, I think this is worth replacing with the HardwareID of the system. Although, since this is a Python port maybe that's how they do it.

https://github.com/shirou/gopsutil/blob/master/host/host_linux.go#L79

I am almost certain this number changes every time the system is rebooted. 

The number that does not change is the Hardware ID (as I understand it) available from About This Mac and with the following command:

```go
func UUID() string {
        info, err := exec.Command("/usr/sbin/system_profiler", "SPHardwareDataType").Output()
        if err != nil {
                return ""
        }
        x := regexp.MustCompile(`(?m)^\s*Hardware UUID:\s*(\S+)`)
        m := x.FindSubmatch(info)
        if len(m) < 1 {
                return ""
        } else {
                return string(m[1])
        }
}

```
![screen_shot_2017-04-11_at_8_26_08_pm](https://cloud.githubusercontent.com/assets/3749067/24936201/4082577a-1ef5-11e7-8fbb-08f9bcb46d86.png)

I found this package looking for a faster way to get the HardwareUUID from a syscall instead of the unbearably slow `exec`ing out for the profiler call (which other libraries do).

shirou:
On linux, I think HostID is defined as follows.

1. read `/sys/class/dmi/id/product_uuid` if available
2. If no, use `kernel.random.boot_id`

I think `product_uuid` is not changed after reboot, so use this as much as possible. But, if not available, use random.boot_id anyway.

And on darwin, we use `sysctl kern.uuid` https://github.com/shirou/gopsutil/blob/master/host/host_darwin.go#L67 to get system UUID. I think it is not changed after the reboot.

davidbirdsong:
@shirou the method of using `kern.uuid` comes with some pretty nasty side effect any hosts that image off of a blessed osx image will share a `kern.uuid`. We run a few hundred OSX nodes and imaging them is the only sane way of managing OS versions. Once imaged, `kern.uuid` is immutable as we understand it.

Newer consul versions use this lib and this tis attribute to generate host-specific id's. When we upgraded consul, we ran into massive failures when it detected duplicate host-id's.  There are work-arounds, but the root of the problem would persist and would smack anyone else using this library in the face like it did for us.

[...]

PR #603, opened 2018-11-06
Have a real fixed HostID on Linux

pierresouchay:
On Linux, most golang programs do not run as root (or at least, they should not),
by default, the kernels uses strict permissions, so most userland programs cannot
read `/sys/class/dmi/id/product_uuid`. However, programs such as Consul are relying
on it to get fixed IDs, instead they have a different ID on each boot.

We propose to use `/etc/machine-id` as fallback https://www.freedesktop.org/software/systemd/man/machine-id.html

In order to fix this, this patch does the following:
 - if `/sys/class/dmi/id/product_uuid` can be read, use it for HostID
 - else if `/etc/machine-id` exists and has 32 chars, use it and add '-' to have the same format as product_uuid
 - finally, if notthing works, use the `kernel.random.boot_id`

This will greatly increase the number of programs having correct behaviour when
those rely on having a fixed HostID.

This will fix the following issues:
 - https://github.com/shirou/gopsutil/issues/350
 - https://github.com/hashicorp/consul/issues/4741

### DIFF
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -73,11 +73,21 @@ func InfoWithContext(ctx context.Context) (*InfoStat, error) {
 	}
 
 	sysProductUUID := common.HostSys("class/dmi/id/product_uuid")
+	machineID := common.HostEtc("machine-id")
 	switch {
+	// When not running as root, lib cannot read the value
 	case common.PathExists(sysProductUUID):
 		lines, err := common.ReadLines(sysProductUUID)
 		if err == nil && len(lines) > 0 && lines[0] != "" {
 			ret.HostID = strings.ToLower(lines[0])
+			break
+		}
+		fallthrough
+	case common.PathExists(machineID):
+		lines, err := common.ReadLines(machineID)
+		if err == nil && len(lines) > 0 && len(lines[0]) == 32 {
+			st := lines[0]
+			ret.HostID = fmt.Sprintf("%s-%s-%s-%s-%s", st[0:8], st[8:12], st[12:16], st[16:20], st[20:32])
 			break
 		}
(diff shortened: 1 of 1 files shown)


host/host_linux.go
```diff
@@ -73,14 +73,24 @@ func InfoWithContext(ctx context.Context) (*InfoStat, error) {
 	}
 
 	sysProductUUID := common.HostSys("class/dmi/id/product_uuid")
+	machineID := common.HostEtc("machine-id")
 	switch {
+	// When not running as root, lib cannot read the value
```
Lomanic: This is not the only way gopsutil could not be able to read this value, the file can be missing for example (this is the case on arm on Debian 9). Remove this comment or completely rephrase it to explain what this `switch case fallthrough` construction does.

pierresouchay:
@lomanic yes, but this case is already properly handled with error management below. 

What I mean is, thus case (file not being readable is not the exception: some weird kernel config or unusual arch), but the rule (I mean, only programs running as root can use it, which in 2018 should be the exception, not the rule)

pierresouchay:
@Lomanic Thank you for the review

shirou:
Wow, Great! Your are thinking backward compatibility so much!

pierresouchay:
@shirou Thank you! Do you think you might release a new version, so I can fix the bug in Consul by updating this library? Kind Regards

Lomanic:
Releases are once per month, following [calver](http://calver.org/) like Ubuntu, see [Tag semantics](https://github.com/shirou/gopsutil#tag-semantics)

linydquantil:
+1

PR #605, opened 2018-11-08
Fix #596 Implement process.Background and process.Foreground functions

### LINKED ISSUE #596: Determine if process is running in the foreground
Is there any way to determine if a process loaded with `process.NewProcess(<pid>)` is running in the foreground? 

I am looking for similar functionality to what `ps -O -p <pid>` returns in the `STAT` column; if there is a `+` in that column then the process is running in the foreground.

Lomanic:
You can test on Unix by comparing the same output between these two programs (might be interesting to get go-tests for this):

```bash
ps aux | tail -n +2 | awk '$8 !~ /'\\+'/ { print $2 }' # get processes whose status doesn't contain the '+' character, remove the first header line with tail (matches 'PID' otherwise)
```

```go
package main

import (
	"fmt"
	"os"

)

func main() {
	if err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
	for _, p := range procs {
		ok, err := process.PidExists(p.Pid)
		if !ok {
			continue
		}
		bg, err := p.Background()
		if err != nil {
			fmt.Println(err)
			os.Exit(2)
		}
		if bg {
			fmt.Println(p.Pid)
		}
	}
}
```

You can also test foreground status with the below scripts, but keep in mind they will also show your current command(s) running if run interactively (two more PIDs for ps+awk, one more for the go program):

```bash
ps aux | awk '$8 ~ /'\\+'/ { print $2 }' # get processes whose status contains the '+' character
```

```go
package main

import (
	"fmt"
	"os"

)

func main() {
	if err != nil {
		fmt.Println(err)
		os.Exit(-1)
	}
	for _, p := range procs {
		ok, err := process.PidExists(p.Pid)
		if !ok {
			continue
		}
		fg, err := p.Foreground()
		if err != nil {
			fmt.Println(err)
			os.Exit(-2)
		}
		if fg {
			fmt.Println(p.Pid)
		}
	}
}
```

Tested successfully on Debian 9 amd64 and armv7l. Might need the same testing on BSDs/macOS.

### DIFF
--- a/process/process_freebsd.go
+++ b/process/process_freebsd.go
@@ -181,7 +181,7 @@ func (p *Process) ForegroundWithContext(ctx context.Context) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	out, err := invoke.CommandWithContext(ctx, ps, "-o", "stat=", "-p", string(pid))
+	out, err := invoke.CommandWithContext(ctx, ps, "-o", "stat=", "-p", strconv.Itoa(int(pid)))
 	if err != nil {
 		return false, err
 	}
--- a/process/process_freebsd.go
+++ b/process/process_freebsd.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/binary"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	cpu "github.com/shirou/gopsutil/cpu"
--- a/process/process_freebsd.go
+++ b/process/process_freebsd.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"os/exec"
 	"strings"
 
 	cpu "github.com/shirou/gopsutil/cpu"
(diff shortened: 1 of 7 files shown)


Lomanic:
I got my hands on a macOS VM to test this PR, I could notice that the implementation on darwin was faulty (this is now fixed, see second commit) and really slow (expected).
I had to add a `process.PidExists()` check in the test above because `process.Pids()` reports its own `ps` process when called, which is already dead when iterating over the found processes. This makes matters worse as `process.PidExists()` is completely inefficient (retrieves all pids and iterates over them), this should be reworked. Python psutil is very defensive and checks for every process-related functions if the pid exists, gopsutil should do the same in every process-related functions, for better error reporting.

process/process_freebsd.go
```diff
@@ -168,6 +169,25 @@ func (p *Process) StatusWithContext(ctx context.Context) (string, error) {
...
+}
+
+func (p *Process) ForegroundWithContext(ctx context.Context) (bool, error) {
+	// see https://github.com/shirou/gopsutil/issues/596#issuecomment-432707831 for implementation details
+	pid := p.Pid
+	ps, err := exec.LookPath("ps")
+	if err != nil {
+		return false, err
+	}
+	out, err := invoke.CommandWithContext(ctx, ps, "-o", "stat=", "-p", strconv.Itoa(int(pid)))
```
shirou: plz add `strconv` to import.

shirou:
It works like a charm except only one missing import on FreeBSD. Awesome!

process/process_freebsd.go
Lomanic: Added, thank you

PR #610, opened 2018-11-18
[disk][linux] Fix #555 Unescape escaped sequences in fstab path in disk.Partitions

### LINKED ISSUE #555: Issue with Mountpoint field of PartitionStat if mounting path has space in it

Lomanic:


Lomanic:
I did not have time to test on Linux,@cjbassi can you confirm this fixes your issue?

cjbassi:
Works for me now! :+1:

Lomanic:
Thanks, merging then.

Issue #582, opened 2018-09-12
A and B drives excluded


sparkblaze:
Is there a particular design reason that A and B drives are excluded inside disk_windows.go? Both letters can be assigned to drives within Windows without causing issues.

https://github.com/shirou/gopsutil/blob/eb1f1ab16f2ef35f23024e10888dfdb5c17fcb92/disk/disk_windows.go#L86

shirou:
because original psutil [skips it](https://github.com/giampaolo/psutil/blob/master/psutil/_psutil_windows.c#L2634). The comment says "it introduces a considerable slowdown."

egonelbre:
My guess is that by reading `A:` or `B:` on a machine with floppy drive it will cause the floppy drive to wake up and start reading it, which is slow.

PR #613, opened 2018-11-20
do not ignore A: and B: drives on windows

tmm1:
from https://www.howtogeek.com/122891/what-are-the-windows-a-and-b-drives-used-for/

>if your computer does not have floppy disk drives, you can assign A and B to volumes

I ran into an issue where a network drive was mounted as b:\ and would not show up through `disk.Partitions(true)`

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -83,9 +83,6 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 	for _, v := range lpBuffer {
 		if v >= 65 && v <= 90 {
 			path := string(v) + ":"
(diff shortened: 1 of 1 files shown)


shirou:
related #582

I think even if you can assign to A or B, you'd better to not use those.

shirou:
hmm. Ok, @Lomanic approved, so I merge it. Thank you for your contribution.

Lomanic:
> I think even if you can assign to A or B, you'd better to not use those.

Problem is, you can't tell that to your end-users, they can still do that. Python psutil is much older, when floppies were more common, we don't have to have the same legacy support. 

Either way, thanks for merging.

PR #614, opened 2018-11-27
[net][linux] Support socket tables in nested namespaces

joestringer:
Support dumping the sockets for nested namespaces.

Previously only sockets in the root network namespace would be dumped. Now, if the PID is nonzero, attempt to dump the sockets associated with the network namespace of the PID. This may be the same if the PID is in the root network namespace, or may be different if it uses a different network namespace.


shirou:
Looks great! thank you!

PR #616, opened 2018-12-13
[process][linux] Export process page fault info

dtnaylor:
Add a function for retrieving page fault counts from /proc/[pid]/stat on
Linux.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -241,6 +241,7 @@ parent              x            x       x       x
 children            x     x      x       x       x
 connections         x            x       x
 is_running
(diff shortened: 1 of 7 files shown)


Lomanic:
This function is lacking on all the other OSes supported (should return ErrNotImpmemented), where it will simply fail to compile.

dtnaylor:
Whoops, you're right—I just added stub functions that return `common.ErrNotImpmementedError` for other OSes.

shirou:
Thank you!

PR #617, opened 2018-12-14
SReclaimable should be summed to Cached Memory

mznet:
psutil sums `SReclaimable` metric to `Cached` when calculating cached metric on linux. And Free command also  sums SReclaimable to Cached.
SReclaimable can be get back and reclaim if under memory pressure, that means SReclaimable can be consider a kind of cached memory.

* psuti https://github.com/giampaolo/psutil/issues?utf8=%E2%9C%93&q=is%3Aissue+slab
```python
 # "free" cmdline utility sums reclaimable to cached.
# Older versions of procps used to add slab memory instead.
# This got changed in:
# https://gitlab.com/procps-ng/procps/commit/
#     05d751c4f076a2f0118b914c5e51cfbb4762ad8e
cached += mems.get(b"SReclaimable:", 0)  # since kernel 2.6.19
```

* procps(free) https://gitlab.com/procps-ng/procps/commit/05d751c4f076a2f0118b914c5e51cfbb4762ad8e

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -44,7 +44,7 @@ type VirtualMemoryStat struct {
 
 	// FreeBSD specific numbers:
 	// https://reviews.freebsd.org/D8467
-	Laundry  uint64 `json:"laundry"`
+	Laundry uint64 `json:"laundry"`
 
 	// Linux specific numbers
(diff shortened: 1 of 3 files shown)


shirou:
Thank you!

`SReclaimable` default value is 0, so I think it works even if a kernel is below 2.6.19.

PR #619, opened 2018-12-17
disk: trim the device mapper name.

remeh:
On Linux, while reading the name of a logical volume in the file with `ioutil.ReadFile`, the resulting name is suffixed with a `\n`.

This commit remove trailing spaces in the read device mapper name.

Tested on Ubuntu 18.10


shirou:
Good catch! thank you!

PR #624, opened 2019-01-02
cpu: implement Mhz and Cores on OpenBSD

ajacoutot:
Adapted from the FreeBSD code.
Successfully tested with Nomad.


shirou:
I have not checked yet but from
https://man.openbsd.org/sysctl.2#HW_NCPUONLINE_2
and 
https://man.openbsd.org/sysctl.2#HW_CPUSPEED_2
it can be work. Thank you!

PR #625, opened 2019-01-04
Add VmHWM

wcc526:
VmHWM: Peak resident set size ("high water mark").

http://man7.org/linux/man-pages/man5/proc.5.html


shirou:
Thank you !

PR #626, opened 2019-01-05
Manual available memory calculation

mznet:
At kernel 3.14 `MemAvailable` metric was introduced, but under 3.14, `MemAvailable` didn't exist and needed to calculate manully to get, gopsutil have [estimated it summing free, cached, and buffer metric](https://github.com/shirou/gopsutil/pull/84). Summing metrics was good idea, could be approached to `MemAvailable` nearly, not perfectly. Calculating the metric more precisely, I am referring to `psutil` and kernel algorithm. 

**psutil**
https://github.com/giampaolo/psutil/commit/8504d930cb68e1e079e03df55112312505ed260f

**kernel**
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -50,8 +50,8 @@ type VirtualMemoryStat struct {
 	// https://www.centos.org/docs/5/html/5.1/Deployment_Guide/s2-proc-meminfo.html
 	// https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 	// https://www.kernel.org/doc/Documentation/vm/overcommit-accounting
-	ActiveFile     uint64 `json:activefile`
-	InactiveFile   uint64 `json:inactivefile`
+	ActiveFile     uint64 `json:"activefile"`
+	InactiveFile   uint64 `json:"inactivefile"`
 	Buffers        uint64 `json:"buffers"`
 	Cached         uint64 `json:"cached"`
 	Writeback      uint64 `json:"writeback"`
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -50,8 +50,6 @@ type VirtualMemoryStat struct {
 	// https://www.centos.org/docs/5/html/5.1/Deployment_Guide/s2-proc-meminfo.html
 	// https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 	// https://www.kernel.org/doc/Documentation/vm/overcommit-accounting
-	ActiveFile     uint64 `json:"activefile"`
-	InactiveFile   uint64 `json:"inactivefile"`
 	Buffers        uint64 `json:"buffers"`
 	Cached         uint64 `json:"cached"`
 	Writeback      uint64 `json:"writeback"`
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -50,6 +50,8 @@ type VirtualMemoryStat struct {
 	// https://www.centos.org/docs/5/html/5.1/Deployment_Guide/s2-proc-meminfo.html
 	// https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 	// https://www.kernel.org/doc/Documentation/vm/overcommit-accounting
+	ActiveFile     uint64 `json:activefile`
+	InactiveFile   uint64 `json:inactivefile`
 	Buffers        uint64 `json:"buffers"`
 	Cached         uint64 `json:"cached"`
 	Writeback      uint64 `json:"writeback"`
--- a/mem/mem_linux.go
+++ b/mem/mem_linux.go
@@ -13,6 +13,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+type VirtualMemoryExStat struct {
+	ActiveFile		uint64	`json:"activefile"`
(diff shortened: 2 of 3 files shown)


mem/mem_linux.go
```diff
@@ -156,3 +159,39 @@ func SwapMemoryWithContext(ctx context.Context) (*SwapMemoryStat, error) {
 	}
 	return ret, nil
 }
+
+func calcuateAvailVmem(ret *VirtualMemoryStat) uint64 {
+	var watermarkLow uint64
+	fn := common.HostProc("zoneinfo")
+	lines, _ := common.ReadLines(fn)
```
shirou: from psutil, [kernel < 2.6.13 zoneinfo is not included](https://github.com/giampaolo/psutil/pull/890/files#diff-befef210ffc1d19a84ee5969886021a3R322). So could you add err check and if err, fall back to original `ret.Free + ret.Buffers + ret.Cached` ?

mem/mem_linux.go
```diff
@@ -156,3 +159,39 @@ func SwapMemoryWithContext(ctx context.Context) (*SwapMemoryStat, error) {
 	}
 	return ret, nil
 }
+
+func calcuateAvailVmem(ret *VirtualMemoryStat) uint64 {
```
shirou: Could you add psutil PR link, https://github.com/giampaolo/psutil/pull/890 which is merged PR as comment?

mem/mem_linux.go
mznet: @shirou such a good point. `calcuateAvailVmem` should check kernel version and error. The fallback will be needed if some error or under kernel version 2.6.13. 

But, `psutil` fallback is the sum of `free` and `cached`, not `buffer`. I think that `buffer` is also needed to be add to fallback. Adding `buffer` is more precise calculation, but they didn't. I need to know why they didn't.

mem/mem_linux.go
mznet: 👍 i will

mem/mem_linux.go
shirou: hmm. So something changed after 2.6.13 about calculation, perhaps. Then, could you change to the sum of free and cached, same as psutil?

mznet: @shirou I found out that [very old linux kernel calculates memory available adding free and cached](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773). I should have read the document carefully. 

Adding free and cached is sufficient calculating memory available for old kernel and that is more suitable for kernel. 

I will commit your review soon.

mznet: @shirou Sorry for late commit. Previous my commit wrongly implemented the algorithm. [I should have used `active(file)` and `inactive(file)` metrics instead of active, inactive](https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L352). 
[On kernel 2.6.28, the kernel specified active and inactive metrics to active(file), active(anon), inactive(file), and inactive(anon)](https://kernelnewbies.org/Linux_2_6_28#Memory_management_Scalability_improvements), I overlooked that. 
`To properly calculate available memory, active(file) and inactive(file) are needed and these should be added to Memory struct in mem.go

mem/mem.go
```diff
@@ -50,6 +50,8 @@ type VirtualMemoryStat struct {
 	// https://www.centos.org/docs/5/html/5.1/Deployment_Guide/s2-proc-meminfo.html
 	// https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 	// https://www.kernel.org/doc/Documentation/vm/overcommit-accounting
+	ActiveFile     uint64 `json:"activefile"`
+	InactiveFile   uint64 `json:"inactivefile"`
```
Lomanic: I'm against adding these two new fields into the `VirtualMemoryStat` type,  as this type is common to all the platforms gopsutil is compatible with and these values don't make sense on other OSes than Linux.

mznet: @Lomanic I worried that too, but I couldn't find a way to calculate what I want without these two new fields. I thought `VirtualMemoryStat` struct already had some fields which not compatible with other platforms such as `SReclaimable` and the two new fields value return 0 on other platforms so that this way didn't make a big issue. 

I fully realized that this way isn't graceful at all and will try to investigate getting memavailable without adding these two new fields. Let me know if you have a good idea.

mznet: @Lomanic I updated this issue. Review, please.

Lomanic: Good job on this, I prefer this way and it reflects what has been done in the `process` package with [MemoryInfoExStat](https://godoc.org/github.com/shirou/gopsutil/process#MemoryInfoExStat).

shirou:
@Lomanic Thank you for reviewing. I think this is good to merge now. How do you think?

shirou:
Merged it. Thank you for your contribution!

PR #629, opened 2019-01-14
Skip non-integer meminfo values instead of error

requilence:
On some systems(at least DD-WRT Linux) `/proc/meminfo` can contain values that are not an integer, but multiple integers split by space. 
It will be better to skip these values instead of returning an error

I don't know what exactly the meminfo key that has this value. I got this from users' error logs. But it was at the start of the file, so the other results were not populated.


shirou:
Could you show us the `/proc/meminfo` ?

shirou:
Since no response, close this PR.

PR #634, opened 2019-02-05
Support Linux kernels without /proc/self/mountinfo (<2.6.26)

qaz52897:
if /proc/self/mountinfo not found, then use /proc/self/mounts


qaz52897:
1 In my case, the server is now running rhel 5. kernel version is v2.6.18.
2 completed
3 completed

qaz52897:
Thanks for your suggestion！

PR #641, opened 2019-02-22
Eliminate WMI queries when calling cpu.Times with percpu=True

nikita-vanyasin:
The current implementation of cpu.Times(true) invokes WMI request to retrieve information about cpu usage. 
This pull request changes the implementation to call Windows API function NtQuerySystemInformation:
https://docs.microsoft.com/en-us/windows/desktop/api/winternl/nf-winternl-ntquerysysteminformation

Working much faster and reliable.
Tested on Windows 10, Windows 7 and Windows Server 2012.

The implementation based on https://github.com/signalfx/signalfx-agent/blob/e89f2c5fad499a2637010ac32a48b16a0d7b4173/internal/monitors/cpu/cpu_windows.go

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -23,18 +23,18 @@ type Win32_Processor struct {
 	MaxClockSpeed             uint32
 }
 
-type win32_PerfRawData_Counters_ProcessorInformation struct {
-	Name                  string
-	PercentDPCTime        uint64
-	PercentIdleTime       uint64
-	PercentUserTime       uint64
-	PercentProcessorTime  uint64
-	PercentInterruptTime  uint64
-	PercentPriorityTime   uint64
-	PercentPrivilegedTime uint64
-	InterruptsPerSec      uint32
-	ProcessorFrequency    uint32
-	DPCRate               uint32
+// SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION
+// defined in windows api doc with the following
+// https://docs.microsoft.com/en-us/windows/desktop/api/winternl/nf-winternl-ntquerysysteminformation#system_processor_performance_information
+// additional fields documented here
+// https://www.geoffchappell.com/studies/windows/km/ntoskrnl/api/ex/sysinfo/processor_performance.htm
+type win32_SystemProcessorPerformanceInformation struct {
+	IdleTime       int64 // idle time in 100ns (this is not a filetime).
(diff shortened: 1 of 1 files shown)


cpu/cpu_windows.go
```diff
@@ -151,21 +144,58 @@ func ProcInfoWithContext(ctx context.Context) ([]Win32_PerfFormattedData_PerfOS_
 }
 
 // perCPUTimes returns times stat per cpu, per core and overall for all CPUs
-func perCPUTimesWithContext(ctx context.Context) ([]TimesStat, error) {
+func perCPUTimesWithContext() ([]TimesStat, error) {
```
Lomanic: Rename this function to `perCPUTimes` if it doesn't accept a context anymore.

cpu/cpu_windows.go
```diff
@@ -151,21 +144,58 @@ func ProcInfoWithContext(ctx context.Context) ([]Win32_PerfFormattedData_PerfOS_
...
+			Idle:   float64(v.IdleTime) / win32_TicksPerSecond,
+			Irq:    float64(v.InterruptTime) / win32_TicksPerSecond,
 		}
 		ret = append(ret, c)
 	}
 	return ret, nil
 }
+
+// makes call to Windows API function to retrieve performance information for each core
+func perfInfoWithContext() ([]win32_SystemProcessorPerformanceInformation, error) {
```
Lomanic: Rename this function to `perfInfo` if it doesn't accept a context anymore.

Lomanic:
Thanks a lot for this PR, removing calls to WMI is always appreciated.
Will test it soon on various Windows versions.

shirou:
I confirmed this works like a charm on Windows10 Pro!

After @Lomanic confirmed also, I or Lomanic will merge. Thank you for your great contribution!

Lomanic:
Tested OK on good old Windows XP 32 bit, works great. However, had an issue on a Tiny7 VM (which is a heavily trimmed down version of Windows 7 32 bit) where the return number was not 0 but the error message was (in substance) "call was successful". We should just merge this MR, it is probably an artifact of the heavy customisation of Tiny7 and @nikita-vanyasin says it works OK on Windows 7.

nikita-vanyasin:
Today I've tested the changes for Windows Server 2012, 2016 and 2019. Works fine too.

shirou:
I really appreciate your all contribution!

PR #642, opened 2019-02-23
Update process_linux.go for Add process ConnectionsMax 

wcc526:
Add ConnectionsMax for /proc/pid/fd open thousands incode conditions.


Lomanic:
You need to implement this function for other OSes (it might return `common.ErrNotImplementedError`). Run `go fmt` on your modifications.

wcc526:
OK,Thank you!

@Lomanic  I have modified,Please review #643

Lomanic:
Please don't duplicate the same PR as you did, just update your branch instead.

PR #643, opened 2019-02-23
Update process_linux.go for Add process ConnectionsMax

wcc526:
Add ConnectionsMax for /proc/pid/fd open thousands incode conditions.


Lomanic:
I'm going to merge this, the changes are innocent. By the way it highlighted to me that `net.ConnectionsPidMax` is only defined on Linux, will have to fix this (so `process.ConnectionsMax` can be the same on all platforms).

PR #645, opened 2019-03-05
add bedrock linux support 

superwhiskers:
on [bedrock linux](https://bedrocklinux.org/) installs, gopsutil improperly detects the operating system as the operating system of the strata the program is being run from. this pull request adds in another utility function that detects the `/bedrock` folder at the root of the system and if it finds it, it changes the files being read from and properly detects the operating system as being bedrock linux

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   test:
     docker:
       - image: circleci/golang:1.9
-    working_directory: /home/circleci/go/src/github.com/shirou/gopsutil
+    working_directory: /home/circleci/go/src/github.com/superwhiskers/gopsutil
     environment:
       GOPATH: /home/circleci/go
     steps:
--- a/README.rst
+++ b/README.rst
@@ -7,8 +7,8 @@ gopsutil: psutil for golang
 .. image:: https://coveralls.io/repos/shirou/gopsutil/badge.svg?branch=master
         :target: https://coveralls.io/r/shirou/gopsutil?branch=master
 
-.. image:: https://godoc.org/github.com/shirou/gopsutil?status.svg
-        :target: http://godoc.org/github.com/shirou/gopsutil
+.. image:: https://godoc.org/github.com/superwhiskers/gopsutil?status.svg
+        :target: http://godoc.org/github.com/superwhiskers/gopsutil
 
 This is a port of psutil (https://github.com/giampaolo/psutil). The challenge is porting all
 psutil functions on some architectures.
@@ -61,7 +61,7 @@ Note: gopsutil v2 breaks compatibility. If you want to stay with compatibility, 
    import (
        "fmt"
 
-       "github.com/shirou/gopsutil/mem"
+       "github.com/superwhiskers/gopsutil/mem"
    )
 
    func main() {
@@ -92,7 +92,7 @@ You can set an alternative location to :code:`/var` by setting the :code:`HOST_V
 Documentation
 ------------------------
 
-see http://godoc.org/github.com/shirou/gopsutil
+see http://godoc.org/github.com/superwhiskers/gopsutil
 
 Requirements
 -----------------
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -10,7 +10,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 )
 
 // TimesStat contains the amounts of time the CPU has spent performing different
--- a/cpu/cpu_darwin_nocgo.go
+++ b/cpu/cpu_darwin_nocgo.go
@@ -3,7 +3,7 @@
 
 package cpu
 
-import "github.com/shirou/gopsutil/internal/common"
+import "github.com/superwhiskers/gopsutil/internal/common"
 
 func perCPUTimes() ([]TimesStat, error) {
 	return []TimesStat{}, common.ErrNotImplementedError
--- a/cpu/cpu_fallback.go
+++ b/cpu/cpu_fallback.go
@@ -5,7 +5,7 @@ package cpu
 import (
 	"context"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 )
 
 func Times(percpu bool) ([]TimesStat, error) {
--- a/cpu/cpu_freebsd.go
+++ b/cpu/cpu_freebsd.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"unsafe"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 	"golang.org/x/sys/unix"
 )
 
--- a/cpu/cpu_freebsd_test.go
+++ b/cpu/cpu_freebsd_test.go
@@ -5,7 +5,7 @@ import (
 	"runtime"
 	"testing"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 )
 
 func TestParseDmesgBoot(t *testing.T) {
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -10,7 +10,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 )
 
 var CPUTick = float64(100)
--- a/cpu/cpu_openbsd.go
+++ b/cpu/cpu_openbsd.go
@@ -11,7 +11,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 	"golang.org/x/sys/unix"
 )
 
--- a/cpu/cpu_solaris.go
+++ b/cpu/cpu_solaris.go
@@ -10,7 +10,7 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 )
 
 var ClocksPerSec = float64(128)
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -8,7 +8,7 @@ import (
 	"unsafe"
 
 	"github.com/StackExchange/wmi"
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 	"golang.org/x/sys/windows"
 )
 
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -3,7 +3,7 @@ package disk
 import (
 	"encoding/json"
 
-	"github.com/shirou/gopsutil/internal/common"
+	"github.com/superwhiskers/gopsutil/internal/common"
 )
 
 var invoke common.Invoker = common.Invoke{}
--- a/disk/disk_darwin.go
+++ b/disk/disk_darwin.go
@@ -7,7 +7,7 @@ import (
 	"path"
 	"unsafe"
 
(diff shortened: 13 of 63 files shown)


Lomanic:
I think this is needlessly complicated (and includes changes to solaris disk package that look unrelated) when you can simply apply the following patch:

```patch
diff --git a/host/host_linux.go b/host/host_linux.go
index 508d8c9..20dc93c 100644
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -248,7 +248,7 @@ func getOSRelease() (platform string, version string, err error) {
                switch field[0] {
                case "ID": // use ID for lowercase
                        platform = field[1]
-               case "VERSION":
+               case "VERSION", "VERISON": // VERISON for bedrock Linux compatibility
                        version = field[1]
                }
        }
@@ -435,6 +435,8 @@ func PlatformInformationWithContext(ctx context.Context) (platform string, famil
                family = "alpine"
        case "coreos":
                family = "coreos"
+       case "bedrock":
+               family = "bedrock"
        }
 
        return platform, family, version, nil
```

Tested with a "hijacked" Debian 9.8.0 with `strat bedrock ./host_platform`, this binary simply doing `fmt.Println(host.PlatformInformation())` (it prints `bedrock bedrock "0.7.1 (Poki)" <nil>`). Without the `strat bedrock` prefix the binary indeed shows `debian debian 9.8.0 <nil>`, because the underlying OS is _still_ a Debian (the rest of the OS is unaware of the bedrock installation, like `lsb_release -a` for example, as it would for a "guest" chroot or container).

superwhiskers:
the main point of it was to detect it regardless of the strata it's being run on, and having to explicitly run it under the context of the bedrock strata defeats that

superwhiskers:
oh, and the solaris file modifications were the result of `gofmt`-ing it all, i can undo that if it's required

Lomanic:
> the main point of it was to detect it regardless of the strata it's being run on, and having to explicitly run it under the context of the bedrock strata defeats that

My reply to this is that the "hijacked" host OS is still the same distro type (see `lsb_release -a` or `/etc/os-release`), that there is a `/bedrock` dir and that the global vars (e.g $PATH) are appending/prepending it in `/bedrock/run/profile` (sourced in `/etc/profile`) doesn't change that my hijacked Debian 9.8.0 is still (identifying as) a Debian 9.8.0. There shouldn't be an exception in gopsutil (as expected in python `platform.linux_distribution()` which host.PlatformInformation is a bit the port reports a debian host in my example).

superwhiskers:
```
My reply to this is that the "hijacked" host OS is still the same
```
it's not the host os though? bedrock itself is technically the host as you can remove the hijacked stratum and it still functions. so therefore whichever stratum you decide to init from is the system that gopsutil reports, which is a little inconsistent. this pr adds consistency by detecting a bedrock install and rightfully marking the system as one. with bedrock, you don't have a debian install that happens to have bedrock on it, you have a bedrock install that happens to have debian on it.

JoshuaDoes:
Can confirm, a Bedrock "hijack" is a Bedrock operating system install with various strata on top of it, not a Debian/whatever install with Bedrock and other strata on top of it. You can simply remove the Debian strata and still have other strata available, so the inconsistency in what is reported is not very helpful when running a program across various strata.

Adding onto a point made earlier, the reason you can run `lsb_release -a` and get the current strata is because that's intended for the rest of the system to realize and understand what strata it's running in and not freak out without having to teach it about Bedrock. But the suggested change to gopsutil would allow for Go programs to consistently understand that they are consistently ran on Bedrock and to be aware of such an environment, not under the current strata.

paradigm:
Not to beat a dead horse here, but I thought maybe some additional information may help clarify things:

The idea behind Bedrock Linux is to get features from other distros.  For example, one could get their kernel from one distro, window manager from another, and web browser from a third.  In this context, the install process is considered such a feature.  Hijacking is how Bedrock gets the install process from another distro.  Once the hijack is completed, you're effectively running Bedrock with the hijacked distro's features made available.  Some ways to think about this are:

- If you mount Bedrock's filesystem somewhere without Bedrock's runtime virtual filesystem subsystems running, you'll find Bedrock's files on the root of the filesystem.  The hijacked distro's files, like those of all the other strata, are in `/bedrock/strata`.  When viewed in this manner, there's nothing special about the hijacked distro's files to qualify it as the "real" distro rather than any of the other strata sitting in directories along side it.
- As both superwhiskers and JoshuaDoes pointed out, once you've fetched some other strata to provide essential features like the init, you can run `brl remove hijacked` to remove the hijacked distro's files.  Since one is running Bedrock, not the hijacked distro, this works fine.  If Debian was the main/base/primary OS, the system would no longer work after removing its files.
- The past releases of Bedrock supported an alternative to the hijack install process where one installs Bedrock by hand first, after which other strata files are placed on top of it.  This workflow was removed from the current release to shift development resources elsewhere, but in principle if it was provided again as an alternative installation option for the current release the resulting system would be functionally the same.

When a process goes to use some file, Bedrock has to determine which stratum's instance of a file to provide.  Often, it's best to provide the instance of a file corresponding to the same stratum as the process.  For example, if a given Bedrock system has (musl-based) Alpine and (glibc-based) Debian, Alpine's `htop` process should utilize Alpine's ncurses library, rather than Debian's.  For another example, if a given Bedrock system had Debian and Ubuntu, `apt` should see Debian's `/etc/apt/sources.list` rather than Ubuntu's.  Other times, getting files across strata boundaries is perfectly fine, such as Debian's `bash` running Arch's `pacman`, or OpenSUSE's `kate` using Gentoo's `terminus` font.

In the past, I found some software (I think Ubuntu's `add-apt-repository`, but I could be misremembering) became upset if `lsb_release` reported another distro.  I thought it not unlikely that other software I am not familiar with may have similar concerns.  Thus, by default Bedrock ensures things like `lsb_release` and `/etc/os-release` and other similar subsystems indicate the same distro as the process that runs/reads them, using the same subsystems as `/etc/apt/sources.list` and the ncurses library in the examples above.  When they reported Debian for Lomanic, that's because Debian's processes were the ones to run/read them.  If another stratum's processes used them, they'd have reported that stratum's distro instead.  Bedrock could have just as easily had them all report they're running Bedrock had there not been concerns about that confusing software running under Bedrock.

> Without the strat bedrock prefix the binary indeed shows debian debian 9.8.0 <nil>

When you selected Debian's init at boot, `strat` (or rather, an under-the-hood equivalent) was run.

If your system never ran `strat` or an equivalent, `cat /etc/os-release` would report the system as Bedrock.  `lsb_release` would not run at all, as Bedrock itself does not provide it; like most features, it gets it from other distros.  You can confirm this by providing `init=/bin/sh` to your bootloader's kernel line in your Bedrock test environment to bypass Bedrock's virtual filesystem setup code (amongst other things).  Then run `cat /etc/os-release` and `lsb_release -a` in the resulting shell.

> the rest of the OS is unaware of the bedrock installation, like `lsb_release -a` for example, as it would be unaware of a "guest" chroot or container

Your mental model here would be closer to what's actually happening if you swapped host and guest.  `lsb_release -a` reports Debian because you're running it in a rough analogue to a chroot or container (or, rather, the opposite of those, as Bedrock's goal is to integrate things, not segregate them).  In fact, when you select an init at boot, run `strat`, run a binary in `/bedrock/cross/bin/`, use a `.desktop` file in `/bedrock/cross/applications`, or likely other things I'm forgetting, the `chroot()` system call gets called under-the-hood.

That having been said, I'd caution against thinking of Bedrock as a host and the other strata as guests.  Every userland process - including Bedrock's - is running "in" a `chroot()`.  Every file on the system - including Bedrock's - sits in `/bedrock/strata/<stratum>`.  The main thing that makes Bedrock's files special is that they're the glue that holes the system together.  If you remove them, the system won't work.

If anyone has advice to make this more evident to new Bedrock users, such as improvements to the documentation, I'm all ears, although it may be best to contact me by some other means to avoid taking this PR off topic.

-----

I worked with neofetch to add Bedrock Linux support.  The way we handled Bedrock detection there was by checking if `/bedrock/etc/bedrock-release` exists *and* if `/bedrock/cross` shows up in the `$PATH`.  I suspect that, if Bedrock support is to be added here, that method is preferable here as well.

Bedrock has a feature available via `strat -r` to disable some cross-stratum hooks for a given process.  This is primarily used when building software that automatically detects dependencies in the environment to avoid pulling in dependencies across distro boundaries.  For example, when creating an Arch Linux package with `makepkg`, it should be prefixed with `strat -r`:

	strat -r arch makepkg

Or when performing the `./configure && make && sudo make install` dance:

	strat -r debian ./configure
	strat -r debian make
	sudo strat -r debian make install

When adding Bedrock support to neofetch, it became evident some workflows may wish to detect the system as Bedrock, and others may wish to report the specific stratum providing neofetch.  Thus:

	$ neofetch | grep OS | cut -d: -f2
	 Bedrock Linux 0.7.1 Poki x86_64
	$ strat arch neofetch | grep OS | cut -d: -f2
	 Bedrock Linux 0.7.1 Poki x86_64
	$ strat -r arch neofetch | grep OS | cut -d: -f2
	 Arch Linux x86_64
	$ strat void neofetch | grep OS | cut -d: -f2
	 Bedrock Linux 0.7.1 Poki x86_64 
	$ strat -r void neofetch | grep OS | cut -d: -f2
	 Void Linux x86_64

If Bedrock support is added to gopsutil, it may be desirable to support this workflow as well.  For example, if someone uses gopsutil as part of some build system, it may be desirable to have a way to tell it to ignore the fact it's running in Bedrock.

All `strat -r` actually does is remove `/bedrock/cross` entries from environment variables like `$PATH`.  Thus, perhaps in addition to checking for `/bedrock/etc/bedrock-release` existing, consider checking for `/bedrock/cross` in the `$PATH`.

It'd be interesting if `strat -r` would also remove `/bedrock/` from the given processes virtual filesystem.  However, with Bedrock's current architecture this would be a lot of work with insufficient benefit.  It doesn't actually help `strat -r`'s originally intended use case, and Bedrock-aware software should be able to just check the `$PATH`.

superwhiskers:
> All strat -r actually does is remove /bedrock/cross entries from environment variables like $PATH . Thus, perhaps in addition to checking for /bedrock/etc/bedrock-release existing, consider checking for /bedrock/cross in the $PATH .

i didn't think of doing that originally, but i'll look into adding that functionality into this pr. i just need to modify `InclusionaryPath` to check for that in `$PATH`

JoshuaDoes:
Very well written @paradigm, nice to see you here! Given what all has been said, I'd find it best to support a way of detecting both the existence of Bedrock and the current stratum the process is being ran under, keeping to the methodology used to report either Bedrock or the current stratum via neofetch.

paradigm:
I should clarify that I meant `$PATH` entries that *start* with `/bedrock/cross`.  For example:

	$ echo $PATH | sed 's/:/\n/g' | grep '\/bedrock\/cross'
	/bedrock/cross/pin/bin
	/bedrock/cross/bin

I'm not planning on adding other `/bedrock/cross` directories to the `$PATH`, but in theory there may be need to do so later.  There's also a possibility a user configures additional items.

superwhiskers:
done @paradigm. running `strat -r <stratum name> <name of program that uses gopsutil>` yields the stratum's information rather than bedrock's

Lomanic:
Thanks for the explanations on strata, I discovered bedrock via this PR so I didn't have the full picture at first.

gopsutil wants to be faithful to other implementations (this is a _port_ of psutil and influenced by [other libraries](https://github.com/shirou/gopsutil#related-works)). Other implementations just report the stratum information, neofetch (a quite different use-case) being the only exception.

I think reporting the /bedrock information should be opt-in (the developer explicitly checks for /bedrock dir existence and in $PATH and sets HOST_ETC with `os.Setenv` accordingly) instead of opt-out (the developer has to explicitly set $HOST_ETC to /etc to be sure that what is reported is information from the chroot/stratum), because of the least surprise principle (similar to all other implementations). Bear in mind that it is acceptable for neofetch to display information from /bedrock, but programs in the wild may break while running on a bedrock instance because they would be relying on what is returned by host.Info.

shirou:
I also don't know bedrock linux, but adding changes which affects all distribution/platform is not a good thing. 

I think just adding a check of `/bedrock/etc/bedrock-release` and PATH, same as [neofetch does](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L852) is acceptable.

superwhiskers:
closing for cleanliness

Issue #646, opened 2019-03-12
[OpenBSD][cpu] errors and incorrect results


omar-polo:
**Describe the bug**
There are a couple of things that don't work on OpenBSD:

- [`cpu.Info`](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_openbsd.go#L137) fails because `unix.SysctlUint32("hw.ncpuonline")` fails
   This happens both on OpenBSD 6.3 (i386) and OpenBSD-CURRENT (amd64). atm I cannot test on other versions/archs. I'm not too familiar with the sysctl api, but I'll try to look into that.

- `cpu.Percent`: if `percpu` is true, the library returns n times the percentages of the average of all cpus, not the individual cpu percentage ([see here](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_openbsd.go#L87)). I'll make a pr for that. However, even after the fix, the results are strange: the cpu1 and cpu3 are always at 0%. Again, I'm not familiar with sysctl, so while I'll try to fix that, help or ideas are welcome =)

**To Reproduce**
```go
// perc.go
package main

import (
        "fmt"
        "log"
        "time"

        "github.com/shirou/gopsutil/cpu"
)

func main() {
        percents, err := cpu.Percent(2*time.Second, true)
        if err != nil {
                log.Fatal(err)
        }

        fmt.Println(percents)
}
```

```sh
for i in `jot 10`; do sleep 1; go run perc.go; done # jot is like linux' seq
# sample output:
[10.555555556157394 0 8.55614973282839 0]
[14.285714285879273 0 10.052910052573967 0]
[14.054054053898863 0 13.297872340392594 0]
[26.94610778451984 0 19.879518072701064 0]
[10.58201058254832 0 10.52631578989694 0]
[14.361702127453713 0 7.3298429320049845 0]
[10.582010581778377 0 10.994764398388417 0]
[28.97727272778009 0 27.071823204744145 0]
[3.7433155084042373 0 10.309278350043751 0]
[9.8958333335386 0 3.6269430051793936 0]
```

**Expected behavior**
- `cpu.Info` should return without erros
- `cpu.Percent` should return percentages consistent with what `top` reports

**Environment:**
 - [x] OpenBSD: `uname -a`: `OpenBSD venera.rosetta.moe 6.5 GENERIC.MP#760 amd64`
Note that while the version reported is 6.5, this is a ~7 days old snapshot

omar-polo:
[...]

PR #647, opened 2019-03-12
[OpenBSD][CPU] fix per-cpu

omar-polo:
#646 related

While there, rearrange the last if statement to make it more similar to the other two.

### DIFF
--- a/cpu/cpu_openbsd.go
+++ b/cpu/cpu_openbsd.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/shirou/gopsutil/internal/common"
 	"golang.org/x/sys/unix"
@@ -101,7 +102,11 @@ func TimesWithContext(ctx context.Context, percpu bool) ([]TimesStat, error) {
 	}
 
 	smt, err := smt()
-	if err != nil {
+	if err == syscall.EOPNOTSUPP {
+		// if hw.smt is not applicable for this platform (e.g. i386),
+		// pretend it's enabled
+		smt = true
+	} else if err != nil {
 		return nil, err
 	}
 
--- a/cpu/cpu_openbsd.go
+++ b/cpu/cpu_openbsd.go
@@ -79,7 +79,7 @@ func smt() (bool, error) {
 
 	var ret bool
 	br := bytes.NewReader(buf)
-	if err := binary.Read(br, binary.LittleEndian, ret); err != nil {
+	if err := binary.Read(br, binary.LittleEndian, &ret); err != nil {
 		return false, err
 	}
 
--- a/cpu/cpu_openbsd.go
+++ b/cpu/cpu_openbsd.go
@@ -29,6 +29,7 @@ var (
 const (
 	CTLKern     = 1  // "high kernel": proc, limits
 	CTLHw       = 6  // CTL_HW
+	SMT         = 24 // HW_SMT
 	NCpuOnline  = 25 // HW_NCPUONLINE
 	KernCptime  = 40 // KERN_CPTIME
 	KernCptime2 = 71 // KERN_CPTIME2
@@ -69,6 +70,22 @@ func init() {
 	}()
 }
 
+func smt() (bool, error) {
+	mib := []int32{CTLHw, SMT}
+	buf, _, err := common.CallSyscall(mib)
+	if err != nil {
+		return false, err
+	}
+
+	var ret bool
+	br := bytes.NewReader(buf)
+	if err := binary.Read(br, binary.LittleEndian, ret); err != nil {
+		return false, err
+	}
+
+	return ret, nil
+}
+
 func Times(percpu bool) ([]TimesStat, error) {
 	return TimesWithContext(context.Background(), percpu)
 }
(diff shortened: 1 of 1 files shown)


omar-polo:
I've pushed a fix for the first issue, the one regarding `cpu.Info()`. I haven't been able to find out why, but `unix.Sysctl("hw.ncpuonline")` fails on:
 - 6.3 on i386
 - 6.4 on amd64
 - CURRENT on amd64

As a workaround, use `unix.Syscall6`. Tested on amd64 (6.4 and CURRENT)

Lomanic:
Thanks for the `cpu.Info` fix, I was working on exactly the same fix this morning. I can confirm it works in OpenBSD 6.4 amd64.

I don't have a multi-core physical machine with OpenBSD (my Virtualbox VM doesn't seem to see more than one core after adding more in Virtualbox machine configuration) to test atm, but this looks OK to me.

omar-polo:
I've managed to fix also the other issue I mentioned on #646 

The reason I got that strange percentage I reported is because of `hw.smt`: `runtime.NumCPU()` returns 4 on my system, but `hw.ncpus` is 8. Only `cpu{0,2,4,6}` are effectively used by the scheduler, but the loop was asking only for `cpu{0..3}`. The last commit is an attempt to detect `hw.smt` and act accordingly. Tested on yesterday' snapshot on amd64.

----

@Lomanic thanks for reviewing the other commits. I'm sorry, I know nothing about running OpenBSD on virtualbox, so I cannot help there :/

omar-polo:
fixed a typo: I forgot to pass the address of the variable instead of the variable itself.

I've tested on the latest snapshot on amd64 with `hw.smt` both enabled and disabled and the results are correct.

omar-polo:
Hi!

I've tested these changes on i386 and noticed that:
 - `hw.ncpuonline` is "operation not supported" on OpenBSD 6.3 (it was added with 6.4 IIRC). Since 6.3 is going EOL soon I don't think it's an issue;
 - if `hw.smt` is not present (ie on i386) the sysctl returns `EOPNOTSUPP`. I've pushed a commit to handle that case;
 - `cpuTimes` is a `[]int64`: on i386 `binary.Read` returns `ErrUnexpectedEOF`. I don't know how to handle this case (binary.Read won't accept an int slice).

Except for this last issue, I think this pr is finished :)

shirou:
OpenBSD released every six months and only support latest two releases.
http://www.openbsd.org/faq/faq5.html#Flavors

So 6.3 will be EOL as you said, but not yet. I don't want remove supported release, so after [6.5 released](https://www.openbsd.org/65.html), I will merge this PR. thank you for you contribution.

Lomanic:
OpenBSD 6.5 got released https://marc.info/?l=openbsd-announce&m=155611207805565&w=2

Does it mean 6.3 is now EOL?

> cpuTimes is a []int64: on i386 binary.Read returns ErrUnexpectedEOF. I don't know how to handle this case (binary.Read won't accept an int slice).

This is problematic and should be fixed before accepting this PR. Instead of doing manual syscalls, wouldn't it be possible (and would it fix it on i386) to use `common.SysctlUint` instead?

omar-polo:
Sorry for the delay.

> OpenBSD 6.5 got released https://marc.info/?l=openbsd-announce&m=155611207805565&w=2
> 
> Does it mean 6.3 is now EOL?

Yup, they support only the two most recent OpenBSD releases (that is, only
the last two release get security fixes [1].)

> > cpuTimes is a []int64: on i386 binary.Read returns ErrUnexpectedEOF. I don't know how to handle this case (binary.Read won't accept an int slice).
> 
> This is problematic and should be fixed before accepting this PR. Instead of doing manual syscalls, wouldn't it be possible (and would it fix it on i386) to use `common.SysctlUint` instead?

Luckily I've just finished updating a i386 from 6.3 to 6.5. I'll try to fix that
and test on both amd64 and i386.

Thanks for the suggestion of `common.SysctlUint`, I think I've missed it.

--

[1]: https://www.openbsd.org/faq/faq5.html#Flavors

omar-polo:
I've recalled why I didn't used `unix.SysctlUint32`: on OpenBSD hw.ncpuonline does not work:

```
import (
        "fmt"
        "golang.org/x/sys/unix"
)

func main() {
        fmt.Println(unix.SysctlUint32("hw.ncpuonline"))
        // outputs: "0 invalid argument"
}
```

Anyway, I've just tried the code on amd64 and i386 (both running 6.5-STABLE) and everything works. As I've been told, OpenBSD uses 32bit integers for both archs, so the error I got was probably due to something not working with 6.3.

I still have to try on 6.4.

omar-polo:
Tested on 6.4-STABLE on i386 (emulated under qemu, it's slow...) and it works.

Tested on 6.4-STABLE on amd64 (under qemu) and it works.

So, the `ErrUnexpectedEOF` I got was probably due to something 6.3 specific that I haven't tracked down.

Lomanic:
> Thanks for the suggestion of `common.SysctlUint`, I think I've missed it. 

It's only declared for freebsd, that's probably why you missed it 😉. Setting it in `internal/common/common_openbsd.go` and calling it with `common.SysctlUint("hw.ncpuonline")` I had the same `invalid argument` error unfortunately (only tested on OpenBSD 6.4 amd64).

Thanks for testing on different archs and versions, I will check also. When it's merged, if we have reports of users under 6.3, we might perhaps implement a workaround for `hw.ncpuonline` by calling `sysctl -n` if the syscall fails.

omar-polo:
ping :)

shirou:
Thank you reporting about OpenBSD 6.3 has some issue, but as you said, 6.3 is now EOL. So I merge this PR. If anyone meet this issue on OpenBSD 6.3, please open a new ticket or PR.

Issue #640, opened 2019-02-21
Support changes in the number of cpus given by cpu.Counts


danielnelson:
**Is your feature request related to a problem? Please describe.**

In Telegraf, it has been suggested that we report changes to the number of cpus without restarting the process.  In KVM, and many other hypervisors the number of cpus can be hotplugged on a running VM.

Currently we are calling `runtime.NumCPU()` in Telegraf directly, but I would like to move this to use `cpu.Counts`.  However, the implementation on Linux is the same `runtime.NumCPU()` call.

**Describe the solution you'd like**

Instead of using the cached number of CPU from Go, each call to cpu.Counts() should attempt to discover the number of CPUs again.

**Describe alternatives you've considered**

None

**Additional context**

https://github.com/influxdata/telegraf/issues/5451

Lomanic:
As raised in #628 we should have OS-defined functions to gather the number of cores, `runtime.NumCPU()` is not sufficient (it returns the number available cores for the process, which can change afterwards on VMs indeed and we can't distinguish between logical and physical cores).

mznet:
@Lomanic Refer to the [document](https://golang.org/pkg/runtime/#NumCPU) about `runtime.NumCPU()`, this method shows only the number of CPUs at process startup and can't reflect the changes after startup. To reflect the change of CPU properly, we should not use `runtime.NumCPU` so that need to find an alternative way, IMHO. 
What about using `SC_NPROCESSORS_ONLN` or `/proc/cpuinfo` to get the number of logical or physical CPUs like [original psutil](https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L628) implementation

danielnelson:
What do you think about just pulling up the implementations from Go without the caching?  Every platform would of course need a different implementation.

Lomanic:
I think we shouldn't reuse `runtime.NumCPU` code with caching removed, we will still have to port psutil code to implement logical/physical cores count (the `logical` argument) afterwards. 

Also almost all the platforms implement processor affinity (how many cores the process can run on) in `runtime.NumCPU`, which is not exactly the number of physical or logical cores on the machine. Plus I'm not even sure it gets updated after process startup and at least one implementation relies on assembly.

PR #650, opened 2019-03-17
[Darwin][CPU] retrieve a cpu count depends on an boolean argument

mznet:
Related to #640, I've implemented darwin cpu count logic based on @Lomanic 's [works](https://github.com/shirou/gopsutil/pull/644).

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -88,5 +88,18 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 }
 
 func CountsWithContext(ctx context.Context, logical bool) (int, error) {
-	return runtime.NumCPU(), nil
+	var cpuArgument string
+	if logical {
+		cpuArgument = "hw.logicalcpu"
+	} else {
+		cpuArgument = "hw.physicalcpu"
+	}
+
+	count, err := unix.SysctlUint32(cpuArgument)
+
+	if err != nil {
+		return runtime.NumCPU(), nil
+	}
+
+	return int(count), nil
 }
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -4,7 +4,6 @@ package cpu
 
 import (
 	"context"
-	"runtime"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cpu/cpu_darwin.go
```diff
@@ -88,5 +88,18 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
...
+	var cpuArgument string
+	if logical {
+		cpuArgument = "hw.logicalcpu"
+	} else {
+		cpuArgument = "hw.physicalcpu"
+	}
+
+	count, err := unix.SysctlUint32(cpuArgument)
+
+	if err != nil {
```
mingrammer: I think it would be better to put the 98, 100 lines together without a new line. That would be more go-ish.

mznet: @mingrammer Thank you for your comment. However, I don't understand why those two lines are less go-ish. Could you let me know what real go-ish way is or send me any articles to help to understand more go-ish way?

Lomanic:
Similar to reference implementation https://github.com/giampaolo/psutil/blob/b31f3bf/psutil/_psutil_osx.c#L348-L385, all good. Will just have to test before merging, but globally seems good otherwise. 

I have an objection regarding the fallback to `runtime.NumCPU` if the sysctl call fails though (not the case in cpu_linux.go), might return the error instead.

cpu/cpu_darwin.go
mingrammer: It's closer to convention rather than rules. Most Go projects including this project treating this kind of code as a single block by placing the error handling line directly underneath the expressions. So I said it would be more go-ish code.

mznet: @mingrammer I've understood what you mean. Error handling and error assignment should be together to distinguish those codes from others, right? Good point 👍

mingrammer: Yap, you're right!

mznet:
@lomanic I went over linux physical count implementation which you did and recognized your review was reasonable. Returning zero and error would be better if an errer exists.

shirou:
Thank you all for your contribution and discussion!

PR #651, opened 2019-03-17
Support go modules

mingrammer:


Lomanic:
Didn't have the chance to play with go modules yet, what was the command you ran to generate this?

mingrammer:
I just ran the `go mod init github.com/shirou/gopsutil` command.

I think you should support Go modules for all of who use the Go modules in their workspace. Without it, module users can't run the tests because some test codes needed to run with internal packages where not accessible from outside of GOPATH.

shirou:
Go modules will enable by default on Go 1.13, release planned Aug. So module is standard way even if I haven't used. And we already use `dep`, ancestor of `go module` so this PR is very good point. thank you!

PR #653, opened 2019-03-18
Remove an unnecessary return variable

mingrammer:
`ret` variable is only used at return. I think the immediate return is better.

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -40,8 +40,6 @@ func Usage(path string) (*UsageStat, error) {
 }
 
 func UsageWithContext(ctx context.Context, path string) (*UsageStat, error) {
-	ret := &UsageStat{}
-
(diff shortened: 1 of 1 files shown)


Lomanic:
I find it less explicit in my opinion, sorry, will probably reject this.

mingrammer:
Yep. I respect your opinion. 

But the first line `ret := &UsageStat{}` is never used before 56 line. So you can use just `ret := &UsageStat{...}` in 56 lines, and remove the first short declaration.

shirou:
I agree with @Lomanic. This does not big improvement and less extendability. so close this PR. sorry.

However, your comment is important. it use memory even if error happened. so I change at b2eb6640fbe09968f82a42f6a993b854af8b62cd.

Thank you for your great suggestion!

Issue #654, opened 2019-03-18
[Darwin][process_darwin.go] convertCPUTimes does not handle cpu time string more than 1 hour.


keefo:
**Describe the bug**
Function convertCPUTimes only handle cpu time string within 1 hour range.
if the string is xx:xx.xx format then the function work great.
But if the string is xx:xx:xx.xx format then the function will fail.

Mac process would exceed hour time very often.
![Screen Shot 2019-03-18 at 3 19 38 PM](https://user-images.githubusercontent.com/809588/54567521-57955200-4991-11e9-8610-76bfac99596d.png)

**To Reproduce**
```go
// +build darwin

package main

import "fmt"

import (
    "strconv"
    "strings"
)

const (
    ClockTicks = 100 // C.sysconf(C._SC_CLK_TCK)
)

func convertCPUTimes(s string) (ret float64, err error) {
    var t int
    var _tmp string
    if strings.Contains(s, ":") {
        _t := strings.Split(s, ":")
        hour, err := strconv.Atoi(_t[0])
        if err != nil {
            return ret, err
        }
        t += hour * 60 * 100
        _tmp = _t[1]
    } else {
        _tmp = s
    }

    _t := strings.Split(_tmp, ".")
    if err != nil {
        return ret, err
    }
    h, err := strconv.Atoi(_t[0])
    t += h * 100
    h, err = strconv.Atoi(_t[1])
    t += h
    return float64(t) / ClockTicks, nil
}

func main() {
    timeVal, err := convertCPUTimes("01:59:23.20")
    if err != nil {
        fmt.Printf("err=%s\n", err)
        return
    }
    fmt.Printf("utime=%f\n", timeVal)
}
```

**Expected behavior**
7163.20000 should be the correct output value for the demo code.

**Environment (please complete the following information):**
 - [ ] Windows: [paste the result of `ver`]
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [X] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**
[Cross-compiling? Paste the command you are using to cross-compile and the result of the corresponding `go env`]

shirou:
your PR is merged so I think this is resolved. Feel free to reopen if something problem. Thank you!

PR #655, opened 2019-03-18
[Darwin][process]Add hour handling in convertCPUTimes function

keefo:
Address #654 

This commit add hour handling in convertCPUTimes function.

The time string usually comes from macOS command line:
ps -a -o stime,utime -p <pid>

which could contain hour string.

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -389,12 +389,31 @@ func convertCPUTimes(s string) (ret float64, err error) {
 	var _tmp string
 	if strings.Contains(s, ":") {
 		_t := strings.Split(s, ":")
-		hour, err := strconv.Atoi(_t[0])
-		if err != nil {
-			return ret, err
-		}
-		t += hour * 60 * 100
-		_tmp = _t[1]
+		if len(_t) > 3 {
+            return ret, err
+        } else if len(_t) == 3 {
+            hour, err := strconv.Atoi(_t[0])
+            if err != nil {
+                return ret, err
+            }
+            t += hour * 60 * 60 * ClockTicks
+
+            mins, err := strconv.Atoi(_t[1])
+            if err != nil {
+                return ret, err
+            }
+            t += mins * 60 * ClockTicks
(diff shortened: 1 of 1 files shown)


mingrammer:
Your code seems to use 4 spaces for indents. But gopsutil is using 8 spaces for it.

keefo:
Hi @mingrammer I have fixed this indentation issue.  Please help me with this pull request review. Let me know if there is anything else I need to do. Thank you.

process/process_darwin.go
```diff
@@ -389,12 +389,31 @@ func convertCPUTimes(s string) (ret float64, err error) {
 	var _tmp string
 	if strings.Contains(s, ":") {
 		_t := strings.Split(s, ":")
-		hour, err := strconv.Atoi(_t[0])
-		if err != nil {
+		if len(_t) > 3 {
```
mznet: @keefo In this method, multiple nested `if` clause is used and it leads to low readability. What about using `switch`  for better readability?

keefo: @mznet  No problem, I will do that.

keefo:
@mznet  I have updated the code to use Switch.

process/process_darwin.go
```diff
@@ -389,12 +389,32 @@ func convertCPUTimes(s string) (ret float64, err error) {
...
+			mins, err := strconv.Atoi(_t[0])
+			if err != nil {
+				return ret, err
+			}
+			t += mins * 60 * ClockTicks
+			_tmp = _t[1]
+		case 1, 0:
+			_tmp = s
+		default:
 			return ret, err
```
Lomanic: From reading the code it looks like this `err` is undefined (not explicitly defined), please use something like `return ret, fmt.Errorf("wrong cputime format")`

keefo: Good point.

keefo: @Lomanic  Done.

shirou:
Thank you so much!!

Issue #665, opened 2019-04-09
No data on one temperature sensor prevents reading any other temp sensor


marcv81:
**Describe the bug**

host_linux.go / SensorsTemperature() returns an error if any of the temperature sensors does not have any data available. It is however an acceptable state for a sensor and should not be treated as an error. Furthermore a single sensor with no data available should not prevent reading the temperatures from any of the other sensors.

Example of a sensor which functions correctly but does not provide any data under certain conditions: Intel Wireless LAN hardware and the iwlwifi module.

WiFi on:

    $ cat /sys/class/hwmon/hwmon4/name 
    iwlwifi
    $ cat /sys/class/hwmon/hwmon4/temp1_input 
    41000

WiFi off:

    $ cat /sys/class/hwmon/hwmon4/temp1_input 
    cat: /sys/class/hwmon/hwmon4/temp1_input: No data available

**To Reproduce**

Please see downstream bug: https://github.com/influxdata/telegraf/issues/5692.

**Expected behavior**

When any of the temperature sensors does not have any data available, host_linux.go / SensorsTemperature() should still read and return the temperature values for the remaining sensors.

**Environment (please complete the following information):**
 - [ ] Windows: [paste the result of `ver`]
 - [X] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [ ] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

`/etc/os-release`

    NAME="Ubuntu"
    VERSION="16.04.6 LTS (Xenial Xerus)"
    ID=ubuntu
    ID_LIKE=debian
    PRETTY_NAME="Ubuntu 16.04.6 LTS"
    VERSION_ID="16.04"
    HOME_URL="http://www.ubuntu.com/"
    SUPPORT_URL="http://help.ubuntu.com/"
    BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
    VERSION_CODENAME=xenial
    UBUNTU_CODENAME=xenial

`uname -a`

    Linux 5.0.7-050007-generic #201904052141 SMP Fri Apr 5 21:43:20 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

Ubuntu 16.04 + Kernel 5.0.7.

Lomanic:
[...]

PR #668, opened 2019-04-11
Fix for #665

marcv81:
SensorsTemperatures() and SensorsTemperaturesWithContext() remain fully backward compatible.

SensorsTemperaturesDetails() and SensorsTemperaturesDetailsWithContext() return an error and an array of warnings. The error (if not nil) is fatal and no temperature sensor was read. The warnings are not fatal and only the impacted temperature sensors were not read.

### DIFF
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -622,17 +622,33 @@ func SensorsTemperatures() ([]TemperatureStat, error) {
 }
 
 func SensorsTemperaturesWithContext(ctx context.Context) ([]TemperatureStat, error) {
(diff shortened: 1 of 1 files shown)


shirou:
Thank you for your contribution. But I don't want to add almost same API (and this PR is only includes for Linux). As @Lomanic said in #572, how about define custom error type which can include multiple error?

marcv81:
Thank you for your reply @shirou. Here is a new PR https://github.com/shirou/gopsutil/pull/675 based on your kind feedback.

PR #674, opened 2019-04-24
Add pages in/out/fault statistics (as read from vmstat)

lootek:
see: https://superuser.com/questions/785447/what-is-the-exact-difference-between-the-parameters-pgpgin-pswpin-and-pswpou

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -84,6 +84,9 @@ type SwapMemoryStat struct {
 	UsedPercent float64 `json:"usedPercent"`
 	Sin         uint64  `json:"sin"`
 	Sout        uint64  `json:"sout"`
+	PgIn        uint64  `json:"pgin"`
+	PgOut       uint64  `json:"pgout"`
+	PgFault     uint64  `json:"pgfault"`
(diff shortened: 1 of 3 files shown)


lootek:
cc @tomekkolo @adamiklukasz

Lomanic:
I don't have a functional computer at hand right now but this will break the `mem` package tests (`TestSwapMemoryStat_String`), so you should amend your commit to fix them. 

This is also a breaking change in the API (`SwapMemoryStat.Sin` and `SwapMemoryStat.Sout` renamed). This is a small change so I might be personally OK with that.

lootek:
@Lomanic thanks for looking at this PR

1. Added missing tests
2. Reverted name change. Shame on me here - normally I'm super-cautious about _not_ making breaking changes... Here I must have (incorrectly) assumed that I was only changing the internal field, leaving the json tag untouched.

shirou:
Thank you for your contribution! page in/out/fault is good information. and other platform at least FreeBSD has same information, so we can expand to others at another PR.

PR #675, opened 2019-04-25
Fix for #665

marcv81:
Remains backward compatible.

When encountering non-fatal errors SensorsTemperatures() returns the temperatures for all the sensor we could read successfully. In that case the custom error contains a list of all the non-fatal errors encountered.

Example usage:

	_, err := SensorsTemperatures()
	if err != nil {
		warns, ok := err.(*Warnings)
		if ok {
			fmt.Printf("%v\n", err)
			for i, w := range warns.List {
				fmt.Printf("Warning %v: %v\n", i+1, w)
			}
		} else {
			t.Errorf("%v", err)
		}
	}

### DIFF
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -635,6 +635,7 @@ func SensorsTemperaturesWithContext(ctx context.Context) ([]TemperatureStat, err
 			return temperatures, err
 		}
 	}
+	var warns Warnings
 
 	// example directory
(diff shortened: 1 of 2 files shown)


marcv81:
I know Warnings is defined in the wrong place. It should be at least in host.go so that the type is defined regardless of the OS, and can be used in downstream apps supporting more than one OS. Should I move it to host.go, or is there a more sensible location?

marcv81:
Update: I defined Warnings in a probably better place, but possibly still not the best place.

shirou:
we should move `warning` to correct place. but I merge at this time to solve the original issue. Thank you for the PR and provide very good discussion point!

Issue #580, opened 2018-09-10
use of internal package not allowed


sundarv85:
Hello,

I’m trying to use the function `DoSysctrl` from external go app available through the library `"github.com/shirou/gopsutil/internal/common"`. 

Since internal is not supposed to be used by Go, I get the error `use of internal package not allowed`

Is there a way to overcome this or another way to reuse `DoSysctrl`

shirou:
Because of [golang specification](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit),  `internal` package is a private and you can not use it from outside of the package.

I have changed these functions to `internal`, however, after several years experiments, I think `internal` package is wrong for OSS library. So if you have some reason to use `DoSysctrl`, could you tell us why and how use it? Thanks.

Lomanic:
> Is there a way to overcome this or another way to reuse DoSysctrl

Copy the [function definition](https://github.com/shirou/gopsutil/blob/master/internal/common/common_linux.go#L11) in another package, while resolving its dependencies ([getSysctrlEnv](https://github.com/shirou/gopsutil/blob/master/internal/common/common.go#L380)), or make the "internal" sub-packages not internal anymore in your vendored gopsutil, by removing "internal" in their import path (you would have to fix the gopsutil packages using them to use the new path, but it is as easy as a `sed -i 's@github.com/shirou/gopsutil/internal/@github.com/shirou/gopsutil/@g'`).

I don't share the view of @shirou, we already export too much internals and platform-specific functions (notably exposing wmi requests on Windows), if someone wants to use gopsutil they don't need to access the "internal" sub-packages and if they want to use these non-exported functions and packages they can fork gopsutil and use them in their own fork as I explained above.

sundarv85:
@shirou @Lomanic The reason I’m looking at using it is, we are looking to identify each and every system uniquely while it is running (ie. when the system reboots, it could register itself as another entity - which is OK for us).

In gopsutil, the hostID is the ideal option for this. However, by default in linux the hostID is given as `product_uuid`. Only as a fallback you provide `boot_id`. In our case, in some VMs we have seen that the `product_uuid` is being duplicated. So here, we have the issue that two machines that are currently running at the same time have a clash in the unique identification. 

Hence, we would prefer to use the `boot_id` in the case of linux. While falling back to the hostId that psutil returns when it is not available. 

[...]

PR #678, opened 2019-04-27
move common package to public from internal.

shirou:
This PR move `internal/common` to top `common` package, outside of `internal`. So users can use common package contents. 

related: #665 

----
After several years passed after `internal` is introduced in Go, I personally think library does not need to use this feature, especially OSS package. There are many users and many kind of demands. `internal` prevents their use case.


Lomanic:
I'm glad that you bring this up.
I understand your concern, but it removes even more flexibility for us maintainers. We are already really blocked for API breaking changes, we would be even more constrained if we make this part of the library public. As I was saying in https://github.com/shirou/gopsutil/issues/580#issuecomment-420180804, there are solutions for developers wanting to use internal functions in their own work. 

But I would agree with this PR if we had more flexibility for API breaking changes, we could assume downstream users vendor gopsutil or use modules, that both pin their version of gopsutil.

marcv81:
This PR was started because there is no external common package, and we would like to add a custom error type shared across all gopsutil packages. Maybe we could instead create a new common package for that is not internal?

Lomanic:
That's indeed an option. For me this is also the occasion to talk about the huge limitation we imposed ourselves of not breaking public APIs between non-major versions. I think this is paralyzing development of the library.

danielnelson:
I think you won't want to make breaking changes between non-major version changes, it breaks with the semantic versioning scheme which is now integral to versioning in Go.  I also think if you break compatibility too frequently you will just end up with many projects using an out-of-date version.

That said, there is some sweet spot to be found and v2 came out years ago.  Maybe it's worth working on a v3 with context support across the board and better error reporting methods?

shirou:
So, there are some options.

1. Move internal common to public.
2. Create `hosts.Warning`.
3. Create `common` package other than current internal.
4. Any other good error reporting and will introduce on v3.

I can understand @Lomanic concern. So, I will close this issue and take option 2. However, if we can found something on option 4, I will create v3 (finally). Let's discuss on another issue.

shirou:
and also an option.

5. any other good solution without breaking API

Lomanic:
Maybe we could make public the `common` package but make it clear (in documentation) that its API is not stable. While still keeping the rest of the library stable API-wise between non-major versions?

shirou:
I will create another PR which creates `general` package on the top directory and listen opinions. Thank you for joining discussion.

Issue #667, opened 2019-04-10
Missing ProcsTotal in load


genofire:
https://godoc.org/github.com/shirou/gopsutil/load

```
type MiscStat struct {
    ProcsTotal int `json:"procsTotal"`
    ProcsRunning int `json:"procsRunning"`
    ProcsBlocked int `json:"procsBlocked"`
    Ctxt         int `json:"ctxt"`
}
```

Lomanic:
What do you mean?

genofire:
> The value after the slash is the number of kernel scheduling entities that currently exist on the system. 
```
genofire@ ~ % cat /proc/loadavg
0.92   0.60   0.50         2/916    22393
load1 load5 load15  running/total   highestpid
```

pytimer:
@genofire You mean that the `MiscStat` struct should add process total, is it? I can take it and implement it. But i want to know which struct should add the process total, `AvgStat` or `MiscStat`?

@genofire @Lomanic What do you think?

genofire:
I prefer `MiscStat` together with the count of `ProcsRunning`

shirou:
I agree to add `ProcsTotal` to `MiscStat`.

pytimer:
@shirou I found you create a issue about [Rename load.MiscStat](https://github.com/shirou/gopsutil/issues/680), if i implement `ProcsTotal`,  do i need to care it?

pytimer:
I send a pr to implement it, but i only know linux about it, so i only implement it.

shirou:
Thank you for your contribution! I will add same information on other platform.

PR #682, opened 2019-05-05
add ProcsTotal in load

pytimer:
The pr add total process threads in load package. Implement https://github.com/shirou/gopsutil/issues/667.


shirou:
I could confirm it works on Linux. Thank you!

PR #685, opened 2019-05-12
Adds ConntrackStats (/proc/net/stat/nf_conntrack)

cmattoon:
Provides more in-depth info on the conntrack tables from `/proc/net/stat/nf_conntrack`

I was debating rolling it into the `FilterStat` struct, but opted to keep it separate for now, since it's a separate source of info. Happy to refactor

    per_cpu_results, err := net.ConntrackStats(true)
    if err != nil {
        log.Fatal(err)
    }

    for i, s := range per_cpu_results {
        fmt.Printf("CPU%d has %d insert_failed entries\n", i, s.InsertFailed)
    }

```
// ConntrackStat has conntrack summary info
type ConntrackStat struct {
        Entries       uint32 `json:"entries"`        // Number of entries in the conntrack table
        Searched      uint32 `json:"searched"`       // Number of conntrack table lookups performed
        Found         uint32 `json:"found"`          // Number of searched entries which were successful
        New           uint32 `json:"new"`            // Number of entries added which were not expected before
        Invalid       uint32 `json:"invalid"`        // Number of packets seen which can not be tracked
        Ignore        uint32 `json:"ignore"`         // Packets seen which are already connected to an entry
        Delete        uint32 `json:"delete"`         // Number of entries which were removed
        DeleteList    uint32 `json:"delete_list"`    // Number of entries which were put to dying list
        Insert        uint32 `json:"insert"`         // Number of entries inserted into the list
        InsertFailed  uint32 `json:"insert_failed"`  // # insertion attempted but failed (same entry exists)
        Drop          uint32 `json:"drop"`           // Number of packets dropped due to conntrack failure.
        EarlyDrop     uint32 `json:"early_drop"`     // Dropped entries to make room for new ones, if maxsize reached
        IcmpError     uint32 `json:"icmp_error"`     // Subset of invalid. Packets that can't be tracked d/t error
        ExpectNew     uint32 `json:"expect_new"`     // Entries added after an expectation was already present
        ExpectCreate  uint32 `json:"expect_create"`  // Expectations added
        ExpectDelete  uint32 `json:"expect_delete"`  // Expectations deleted
        SearchRestart uint32 `json:"search_restart"` // Conntrack table lookups restarted due to hashtable resizes
}
```

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -213,6 +213,12 @@ func ReadInts(filename string) ([]int64, error) {
 	return ret, nil
 }
 
+// Parse Hex to uint32 without error
+func HexToUint32(hex string) uint32 {
+	vv, _ := strconv.ParseUint(hex, 16, 32)
+	return uint32(vv)
+}
+
 // Parse to int32 without error
 func mustParseInt32(val string) int32 {
 	vv, _ := strconv.ParseInt(val, 10, 32)
--- a/internal/common/common_test.go
+++ b/internal/common/common_test.go
@@ -51,6 +51,12 @@ func TestByteToString(t *testing.T) {
 	}
 }
 
+func TestHexToUint32(t *testing.T) {
+	if HexToUint32("FFFFFFFF") != 4294967295 {
+		t.Error("Could not convert")
+	}
+}
+
 func TestmustParseInt32(t *testing.T) {
 	ret := mustParseInt32("11111")
 	if ret != int32(11111) {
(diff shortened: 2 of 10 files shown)


shirou:
It works on my Linux, Ubuntu 16.04. This PR has functions for other platforms (NotImplemented is ok). And I have not studied yet but other platforms at least *BSD should have same function (because pf is developing on BSD). So I think it is great PR to merge. Thank you!

Memo for future developing : `pfctl -s states` for BSD.

PR #687, opened 2019-05-21
Avg disk queue length windows

tomekkolo:
weighted IO statistic is equivalent of AvgDiskQueueLength from Win32_PerfFormattedData_PerfDisk_LogicalDisk


shirou:
gopsutil is aim to be multi-platform. Do you have some information about other platform such as Linux, Darwin about avg disk queue length?

Lomanic:
WeightedIO is only used [on Linux atm](https://github.com/shirou/gopsutil/blob/5335e3fd506df4cb63de0a8239c7461d23063be6/disk/disk_linux.go#L410), it was introduced in https://github.com/shirou/gopsutil/pull/263. From [Documentation/iostats.txt](https://www.kernel.org/doc/Documentation/iostats.txt)

> Field 11—weighted # of milliseconds spent doing I/Os This field is incremented at each I/O start, I/O completion, I/O merge, or read of these stats by the number of I/Os in progress (field 9) times the number of milliseconds spent doing I/O since the last update of this field. This can provide an easy measure of both I/O completion time and the backlog that may be accumulating. 

Compared to AvgDiskQueueLength from [Win32_PerfFormattedData_PerfDisk_LogicalDisk](https://technet.microsoft.com/en-ca/aa394261(v=vs.71))

> AvgDiskQueueLength
Average number of both read and write requests that were queued for the selected disk during the sample interval.

They are **not** the same (on Linux, a time in milliseconds is returned, in Windows WMI a number of requests during a  time is returned).

It may be possible to get similar info on Windows (search [this technet blog post](https://blogs.technet.microsoft.com/robertsmith/2012/02/07/analyzing-storage-performance-using-the-windows-performance-analysis-toolkit-wpt/) for "I/O completion time") but it would be probably by reverse-engineering this tool…

Lomanic:
Closing, see my previous comment about WeightedIO not being the same thing as AvgDiskQueueLength.

PR #688, opened 2019-05-22
use /proc/buc/pci/devices to identify a VM is a guest

asnowfox:
I tested gopsutil on openstack VMs and Physical Machines. Current code will idenetify a VM with system is **kvm** and role is **host**. it should be guest

so, I use /proc/buc/pci/devices  to identify this machine is host or guest role


shirou:
From my understanding, this PR means because virtio is used on a guest only, so if `virtio-pci` in the `bus/pci/devices`, it should be a guest.
I don't have a knowledge about virtio and KVM, I think it is appropriate. so now merge this. If someone find `virtio` used in a host, please open a PR.

Thank you!

PR #689, opened 2019-05-23
Remove cycle between process and host packages

ZymoticB:
gopsutil is a transitive dependency of another project that I am integrating
into an internal build system. We target multiple platforms and as a part
of the build system for the large internal repo, we calculate the build
graph used to determine what targets have changed and need to be build /
tested as a single [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) for all platforms.

gopsutil currently does not form a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) if linux and any other platform are
considered at the same time. linux is the only platform where the process
package imports the host package.

To remove this cycle, the relevant methods have been moved to internal/common
with the linux build tag and are consumed the host and process packages.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -1235,7 +1235,7 @@ func (p *Process) fillFromTIDStatWithContext(ctx context.Context, tid int32) (ui
 		System: float64(stime / ClockTicks),
 	}
 
-	bootTime, _ := common.BootTime()
+	bootTime, _ := common.BootTimeWithContext(ctx)
 	t, err := strconv.ParseUint(fields[i+20], 10, 64)
 	if err != nil {
 		return 0, 0, nil, 0, 0, 0, nil, err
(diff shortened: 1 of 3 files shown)


process/process_linux.go
```diff
@@ -1236,7 +1235,7 @@ func (p *Process) fillFromTIDStatWithContext(ctx context.Context, tid int32) (ui
 		System: float64(stime / ClockTicks),
 	}
 
-	bootTime, _ := host.BootTime()
+	bootTime, _ := common.BootTime()
```
Lomanic: Remove the `BootTime()` definition in the `common` package and instead propagate the passed `ctx`variable with `common.BootTimeWithContext(ctx)` here.

shirou:
Thank you! I should merge this PR first but I fix the conflict.

PR #691, opened 2019-05-24
added sles to the suse platform family

ArtieReus:
Sles15 just have the os-release file. Since the function "getOSRelease" just return the ID attribute, in that case "sles", as a platform there is no matching in the method to get the platform family "PlatformInformationWithContext". Adding "sles" in the switch case will fix the gap.


Lomanic:
Can you please paste here the content of `/etc/os-release` to have a test?

ArtieReus:
NAME="SLES"
VERSION="15"
VERSION_ID="15"
PRETTY_NAME="SUSE Linux Enterprise Server 15"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15"

Lomanic:
As-is it doesn't work with the file content you provided (we don't handle double quotes) but I'm fixing it right now. Thanks for your PR.

PR #696, opened 2019-05-29
Fix net.ConnectionsMax BUG

wcc526:
when use `net.ConnectionsMax` function getProcInodesAll will cause EOF error which return wrong answer

### DIFF
--- a/net/net_linux.go
+++ b/net/net_linux.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"io"
 
 	"github.com/shirou/gopsutil/internal/common"
 )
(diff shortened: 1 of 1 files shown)


Lomanic:
I could not reproduce it, do you have a reproduction case?

shirou:
if `os.Open(dir)` or `os.Readlink(inodePath)` in the `getProcInodes` interrupted some reason, it may return `io.EOF`. However it is not expected and should treat as an error, not continue.

Or, if you find reproducable case, please re-open and tell us it.

thank you!

wcc526:
> unhandled

	`connectionsList, err := net.ConnectionsMax("tcp4", 1000)`

when you run net.ConnectionsMax,you will find some proc is not equal with the normal

PR #698, opened 2019-05-31
Update host_darwin.go

### LINKED ISSUE #694: platformFamily is empty on macOS
**Describe the bug**
Value platformFamily in host.InfoStat is empty on macOS.

**To Reproduce**
```go
package main

import (
	"fmt"

	"github.com/shirou/gopsutil/host"
)

func main() {
	i, _ := host.Info()
	fmt.Printf("%+v", i)
	// output: {"hostname":"laptop","uptime":124542,"bootTime":1558595441,"procs":411,"os":"darwin","platform":"darwin","platformFamily":"","platformVersion":"10.14.5","kernelVersion":"18.6.0","virtualizationSystem":"","virtualizationRole":"","hostid":"xxx-xx-xx-xx-xxx"}
}
```

**Expected behavior**
platformFamily should be "darwin"

picatz:
🍎This aims to fix #694 

The `/System/Library/CoreServices/ServerVersion.plist` file exists on macOS servers, but not on a workstation such as my laptop. The actual terminology is mostly borrowed from the [windows equivalent](https://github.com/shirou/gopsutil/blob/master/host/host_windows.go#L223) as @Lomanic suggested. In theory, this should make interpreting the results from the two platforms a bit more consistent.

> **Note:** The macOS server application can be installed on almost any macOS workstation to make it a server that can manage other apple devices.

### DIFF
--- a/host/host_darwin.go
+++ b/host/host_darwin.go
@@ -189,6 +189,16 @@ func PlatformInformationWithContext(ctx context.Context) (string, string, string
 		pver = strings.ToLower(strings.TrimSpace(string(out)))
 	}
 
+	// check if the macos server version file exists
+	_, err := os.Stat("/System/Library/CoreServices/ServerVersion.plist")
+
+	// server file doesn't exist
+	if os.IsNotExist(err) {
+		family = "Standalone Workstation"
+	} else {
+		family = "Server"
+	}
+
 	return platform, family, pver, nil
 }
 
--- a/host/host_darwin.go
+++ b/host/host_darwin.go
@@ -190,7 +190,7 @@ func PlatformInformationWithContext(ctx context.Context) (string, string, string
 	}
 
 	// check if the macos server version file exists
-	_, err := os.Stat("/System/Library/CoreServices/ServerVersion.plist")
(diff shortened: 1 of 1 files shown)


shirou:
Could some one confirm a macOS Server has this file? I can confirm a non-server does not have that file.

Lomanic:
Can confirm this file is missing on a "regular" (as regular as can be a hackintosh VM) install

```
ls -la /System/Library/CoreServices/*.plist
-rw-r--r--  1 root  wheel  4409 30 jul  2016 /System/Library/CoreServices/PlatformSupport.plist
-r--r--r--  1 root  wheel   477 14 sep  2016 /System/Library/CoreServices/SystemVersion.plist
-rw-r--r--  1 root  wheel  3333 18 aoû  2016 /System/Library/CoreServices/com.apple.NSServicesRestrictions.plist
```
For obvious reasons I don't think I will be able to install it [via the appstore](https://itunes.apple.com/app/os-x-server/id883878097) to test if the file gets created.But from a simple web search, it looks like it is present.

picatz:
Thank you @Lomanic 🙌

shirou:
great! thank you!

PR #702, opened 2019-06-12
Added a few examples for easy demonstration

souvikhaldar:


Lomanic:
1. This should be in godoc instead of dedicated files. See https://blog.golang.org/examples
2. These examples are bad practice as the returned errors are systematically discarded
3. We already have tests that can show how to use gopsutil functions

Thanks for your PR, but I can't merge it for these reasons.

Issue #607, opened 2018-11-12
process.PidExists() is inefficient, should only check given PID


Lomanic:
Instead of [listing every processes and iterating over each of them](https://github.com/shirou/gopsutil/blob/48177ef5f8809fc72b716c414435f2d4cff8e24d/process/process.go#L134-L147) it should use the OS capabilities to only check the given pid:

* Linux: check if `/proc/$PID` exists
* BSDs/darwin: check the result of `ps -o pid= -p $PID`
* [Windows](https://stackoverflow.com/a/592788): OpenProcess + GetExitCodeProcess
* or Linux+BSDs+Darwin: [try `kill -s 0 $PID`, and check the error returned](https://github.com/giampaolo/psutil/blob/a3d6a28be2631cae7f78287b0742bba36338a745/psutil/_psposix.py#L24-L49)

This might let us systematically check process existence at the beginning of every process function like Python psutil does/when an error is encountered, for better error messages. As discussed in https://github.com/shirou/gopsutil/pull/605#issuecomment-437794162

_Note: depending of the implementation, it might still be more efficient to keep the current "get all running processes and iterate over them" approach to check for PID existence, this should always be benchmarked._

Lomanic:
Looks like [as simple as using  `os.FindProcess` and `process.Signal(syscall.Signal(0))`](https://stackoverflow.com/questions/15204162/check-if-a-process-exists-in-go-way), only problem is on Windows[ `os.FindProcess`](https://golang.org/src/os/exec_windows.go#L89) doesn't do as many checks as Python psutil, so we will still need to redefine the same function in `process/process_posix.go` and `process/process_windows.go` (using syscalls there).

PR #704, opened 2019-06-15
preventing file open and bad defer close call

josedh:
The previous function was calling defer Close() before checking the error. THis could potential lead to a panic if calling close on a nil pointer. Fixed by removing the os.Open call altogether and using os.Stat instead.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -69,12 +69,10 @@ func (m MemoryMapsStat) String() string {
 // to get more information about the process. An error will be returned
 // if the process does not exist.
 func NewProcess(pid int32) (*Process, error) {
-	p := &Process{
-		Pid: int32(pid),
+	if _, err := os.Stat(common.HostProc(strconv.Itoa(int(p.Pid)))); err != nil {
+		return nil, err
 	}
-	file, err := os.Open(common.HostProc(strconv.Itoa(int(p.Pid))))
-	defer file.Close()
-	return p, err
+	return &Process{Pid: int32(pid)}, nil
 }
 
 // Ppid returns Parent Process ID of the process.


Lomanic:
Looks like this kind of check (checking if PID exists) is only done on the linux platform. Shall we extend this check to other platforms or remove it here instead?

josedh:
I went ahead and added the process check to all system types. Moved away from file check and instead tried to send a 0 signal to the process to check for process existence.

Lomanic:
Please test this on Windows at least once (process.Processes() returns an empty list with this PR as an example). See #607 (signals are a POSIX thing, Windows is not POSIX).

Otherwise, please remove your last commit and this will be merged as-is.

josedh:
@Lomanic: What if I simply update the return line (631) to just return the new process and nil. It seems the call to os.FindProcess * should * handle the check for existence of the process in windows, and thus be sufficient to cover this case. See: https://golang.org/src/os/exec_windows.go#L90

Lomanic:
Please reset your branch to cd515b51f3c5cba146c9988cc1b01c3a9c2c72ca (`git reset --hard 13375e2f9c9b4f2cfca1f851acde9a674b27fce8 && git push --force`) and this will be merged as-is, as I tested this commit working properly on Linux. Your subsequent commits were obviously not tested (it currently doesn't even compile on Windows as proc variable is not used, and you didn't seem to follow links posted in the issue I linked in my previous comment to implement the check properly for POSIX OSes) and my bandwidth is limited, so let's keep this PR short. Thanks in advance.

josedh:
Done. Sorry about the mess. I'll try to revisit the rest of the platforms at some point and hopefully add an appropriate fix for those as well

Lomanic:
Your last commit was not tested (it did not compile with undefined var p). I fixed it in master.

josedh:
Thanks. Seems like my vim-go installation was messed up and it reported "build sucessful" when I ran :GoBuild

PR #706, opened 2019-06-18
[process][windows] Use win32 API in process.Nice() instead of slow WMI call

Lomanic:
Convert priority classes values to their WMI equivalent for backward compatiblity.

Equivalencies were determined with the following Go program (to be run as Administrator to get REALTIME_PRIORITY_CLASS value)

```go
package main

import (
	"fmt"
	"os"
	"syscall"

	"golang.org/x/sys/windows"
)

var Modkernel32 = windows.NewLazySystemDLL("kernel32.dll")

var priorityClasses = map[string]int{
	"ABOVE_NORMAL_PRIORITY_CLASS": 0x00008000,
	"BELOW_NORMAL_PRIORITY_CLASS": 0x00004000,
	"HIGH_PRIORITY_CLASS":         0x00000080,
	"IDLE_PRIORITY_CLASS":         0x00000040,
	"NORMAL_PRIORITY_CLASS":       0x00000020,
	"REALTIME_PRIORITY_CLASS":     0x00000100,
}

func main() {
	pid := os.Getpid()
	// 0x0200 is PROCESS_SET_INFORMATION https://docs.microsoft.com/fr-fr/windows/desktop/ProcThread/process-security-and-access-rights
	c, err := syscall.OpenProcess(0x0200, false, uint32(pid))
	if err == nil {
		defer syscall.CloseHandle(c)
	}

	proc, _ := process.NewProcess(int32(pid))
	procSetPriorityClass := Modkernel32.NewProc("SetPriorityClass")
	for k, v := range priorityClasses {
		procSetPriorityClass.Call(uintptr(c), uintptr(v))
		nice, err := proc.Nice()
		if err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
		fmt.Println(k, nice)
	}
}
```

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -34,6 +34,7 @@ var (
 	procAdjustTokenPrivileges = advapi32.NewProc("AdjustTokenPrivileges")
 
 	procQueryFullProcessImageNameW = common.Modkernel32.NewProc("QueryFullProcessImageNameW")
+	procGetPriorityClass           = common.Modkernel32.NewProc("GetPriorityClass")
 )
 
 type SystemProcessInformation struct {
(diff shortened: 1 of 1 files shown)


shirou:
before
```
$ time go test --run Test_Process_Nice
PASS
ok      github.com/shirou/gopsutil/process      0.155s
```
after
```
$ time go test --run Test_Process_Nice
PASS
ok      github.com/shirou/gopsutil/process      0.078s
```

I know this `time` is not so accurate but this shows 2x faster. Thank you so much!

PR #707, opened 2019-06-23
[cpu] WIP #599 cap percent values returned by Percent() between 0 and 100

### LINKED ISSUE #599: On CentOS, the value returned by process.Percent()  is bigger than 100%
How I get cpu usage:
```
c.CPUUsage, err = p.Percent(time.Second * 1)
```
Centos Version: 6.4

Lomanic:
[cpu] Fix #599 cap percent values returned by Percent() between 0 and 100

See https://github.com/shirou/gopsutil/issues/599#issuecomment-491942842 for a reproduction case

### DIFF
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"sync"
@@ -112,7 +113,7 @@ func calculateBusy(t1, t2 TimesStat) float64 {
 	if t2All <= t1All {
 		return 1
 	}
-	return (t2Busy - t1Busy) / (t2All - t1All) * 100
+	return math.Min(100, math.Max(0, (t2Busy-t1Busy)/(t2All-t1All)*100))
 }
 
 func calculateAllBusy(t1, t2 []TimesStat) ([]float64, error) {


cpu/cpu.go
```diff
@@ -112,7 +113,7 @@ func calculateBusy(t1, t2 TimesStat) float64 {
 	if t2All <= t1All {
 		return 1
```
Lomanic: Noticed during fixing this, shouldn't it be `return 100` @shirou @vrecan? This comes from https://github.com/shirou/gopsutil/commit/c3899894530a9b38dcad8ceca651ec44895f8441

cpu/cpu.go
```diff
@@ -112,7 +113,7 @@ func calculateBusy(t1, t2 TimesStat) float64 {
...
 	}
-	return (t2Busy - t1Busy) / (t2All - t1All) * 100
+	return math.Min(100, math.Max(0, (t2Busy-t1Busy)/(t2All-t1All)*100))
```
kconwayatlassian: Should this be
```
return math.Max(0, math.Min(...)) 
```
to further control the lower bound of zero?

Lomanic:
In fact I probably still have to do the same thing for process.Percent() (what the initial bug report was about), hence WIP.

Lomanic:
> However, is there something deeper that is causing values beyond the expected bounds to begin with?

It might be a rounding error. In the reproduction case you posted, we get CPU percent in a tight loop (for a short period each iteration) and as I understand it it artificially increases values returned, next to 100% sometimes even.

cpu/cpu.go
shirou: exactly! I will fix in another PR.

shirou:
Thank you so much!

PR #711, opened 2019-07-06
Add support for hfsplus file system

tonylambiris:


shirou:
HFSPLUS is not included in "[man stafs](http://man7.org/linux/man-pages/man2/statfs.2.html)", but I can find `HFSPLUS_SUPER_MAGIC` in [some](https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/lib/file/fileInt.h#L57) of [linux](https://elixir.bootlin.com/linux/v4.3/source/fs/hfsplus/hfsplus_raw.h#L24) source codes, so I think it is valid and useful. 

Thank you for your awesome contribution!

Issue #709, opened 2019-06-30
[host][freebsd] Users() returns incorrect "Started" timestamp


ppmathis:
**Describe the bug**
Calling `host.Users()` on a FreeBSD system which uses the `utmpx` POSIX standard (default since FreeBSD 9.0) returns a slice of `UserStat` structs, which contains an incorrect value for the `Started` attribute. Instead of containing a UNIX timestamp representing the start time of the user session, a seemingly random timestamp is being returned.

**To Reproduce**
Start a new user session at a FreeBSD system (e.g. SSH to the machine), then run the following code snippet:
```go
package main

import (
	"fmt"
	"github.com/shirou/gopsutil/host"
)

func main() {
	users, _ := host.Users()
	for _, user := range users {
		fmt.Printf("%s@%s@%s - %d", user.User, user.Terminal, user.Host, user.Started)
	}
}
```
As an example, this resulted in the following output on my system: `root@pts/0@<my reverse dns> - 1732593154` - converting the supposed UNIX timestamp to a human-readable date results in `11/26/2024 @ 3:52am (UTC)`, which is obviously way off as I am unable to start sessions in the future :)

**Expected behavior**
Calling `host.Users()` on a FreeBSD system should return a correct UNIX timestamp specifying when the session got started. This behavior would be equivalent to the other supported operating systems, e.g. Linux.

**Environment (please complete the following information):**
 - [X] FreeBSD: `12.0-RELEASE-p6 / FreeBSD gopsutil-devel 12.0-RELEASE-p6 FreeBSD 12.0-RELEASE-p6 GENERIC  amd64`

**Root cause**
gopsutil is currently using the C struct declaration `utmpx`, which is incompatible with the format used by FreeBSD for `/var/run/utx.active`. While `struct utmpx` is being represented according to the POSIX standards and used internally in the kernel, FreeBSD chose to use a different representation for the `utx.active` file.

This can be read in the [initial commit](https://github.com/freebsd/freebsd/commit/ed79cccfabe6b9ba70af232bc2fb874b3062e965#diff-9cf5ac6b956bed4e48d9d3a2ac0740bb) which introduced the utmpx implementation to FreeBSD in 2010:

> The standard allows the on-disk format to be different than the in-memory representation (struct utmpx). Most operating systems don't do this, but we do. This allows us to keep our ABI more stable, while giving us the opportunity to modify the on-disk format. It also allows us to use a common file format across different architectures (i.e. byte ordering).

[...]

PR #712, opened 2019-07-06
fix FreeBSD Users() time

shirou:
Thanks for the report from #709, this PR fixes FreeBSD Users() startsec.

Note: I only created from amd64. 386 and arm are just coped from amd64 version. so it may be broken.


ppmathis:
Thank you very much for the quick fix, @shirou. While I do not have access to any FreeBSD machines running on arm64, I can confirm that the PR works 100% fine on multiple systems running either i386 or amd64.

shirou:
Thank you confirming!

PR #713, opened 2019-07-07
[process] Fix #607 properly check PID existence

### LINKED ISSUE #607: process.PidExists() is inefficient, should only check given PID

Lomanic:
Fix #607

Use signal(0) on posix-compliant platforms (linux freebsd openbsd darwin)

Use OpenProcess+GetExitCodeProcess on windows 

Tested on linux, windows (10, broken on XP), darwin, openbsd, freebsd

Regarding WinXP, this is failing because of PROCESS_QUERY_LIMITED_INFORMATION being introduced in in Vista. Calling `windows.OpenProcess` with `windows.PROCESS_QUERY_INFORMATION` works properly on WinXP (hinting that every calls to OpenProcess with PROCESS_QUERY_LIMITED_INFORMATION in gopsutil fail on WinXP). Handling this (opening process with right parameter according to Windows version) would complicate a little bit the code in process_window.go (while being prudent not too slow down gopsutil functions), I'm not sure if we should bother so much about WinXP now that it's deprecated for quite a few years.

### DIFF
--- a/process/process_posix.go
+++ b/process/process_posix.go
@@ -4,6 +4,7 @@ package process
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


shirou:
This is really great PR! thank you so much!

process/process_posix.go
```diff
@@ -69,6 +70,34 @@ func getTerminalMap() (map[uint64]string, error) {
...
 
+func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
+	if pid <= 0 {
+		return false, fmt.Errorf("invalid pid %v", pid)
+	}
+	proc, err := os.FindProcess(int(pid))
+	if err != nil {
+		return false, err
+	}
+	err = proc.Signal(syscall.Signal(0))
```
i-prudnikov: This line actually breaks the whole story if you are running inside container, and you are tracking host processes.
More details can be found here:
https://github.com/influxdata/telegraf/issues/6813

PR #715, opened 2019-07-07
[process] Fix #599 cap percent values returned by *Percent() between 0 and 100

### LINKED ISSUE #599: On CentOS, the value returned by process.Percent()  is bigger than 100%

Lomanic:
[process] Fix #599 cap percent values returned by *Percent() between 0 and 100

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"math"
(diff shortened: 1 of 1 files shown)


shirou:
Great!

PR #720, opened 2019-07-12
[disk][openbsd] Use fallback for openBSD not on amd64

imsodin:
Currently using this module and building on OpenBSD on any other arch than amd64 results in a build failure. It would be nicer if it used the fallback as unsupported OS', which is what this PR aims to do.

### DIFF
--- a/disk/disk_fallback.go
+++ b/disk/disk_fallback.go
@@ -1,4 +1,4 @@
-// +build !darwin,!linux,!freebsd,!openbsd,!windows,!solaris
+// +build !darwin,!linux,!freebsd,!openbsd,!windows,!solaris openbsd,!amd64
 
 package disk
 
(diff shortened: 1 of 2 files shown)


shirou:
oh, you are right! Thank you!

Lomanic:
Thanks for this PR, in fact it prompted me to add 386 definitions in #721. Will amend this so we are able to properly build on 386 without using the fallback.

imsodin:
Very nice, thanks. Is there going to be a late July/7 release or will this be released in August? Just for curiosity/planning - merging is blocked due to CI failing on this, we're not going to release until August so we can just delay merging (we being Syncthing :) ).

Lomanic:
Probably by the end of the month (last one was only 2~ weeks ago). @shirou is in charge of that (I'm not even sure I have sufficient rights for that).

PR #722, opened 2019-07-16
Revert "[disk][openbsd] Use fallback for openBSD not on amd64"

imsodin:
This reverts commit 3aa75af2ac502b28e078f653497a24d3809c63d4

Merging both #720 and #721 resulted in the new i386 support to be disabled -> revert my build constraint changes.


Lomanic:
Thank you, I didn't have time to do that before.

shirou:
ah, sorry, my mistake...

PR #723, opened 2019-07-17
define CPUTick on missing platforms

tychoish:


Lomanic:
This constant is only exported in `cpu/cpu_linux.go` because it is used in `docker/docker_linux.go`, not for enduser's consumption. I'm of the advice to unexport this definition instead (it's an implementation detail leaking that's not useful for gopsutil users), but if it complicates too much our implementation (moving it to a `internal ` sub package or duplicating code), I can make an exception for this one. Can you reference where you got these values from in comments?

tychoish:
I'm just trying to figure out what the values in the cpu.TimeStat are divided by, and so I used those values. I think that needs to be exposed in a cross platform way..

shirou:
I do not want to add unused variables. If docker can run some other platforms or `CPUTick` become to be used from other packages like process, then, we will add those. But good suggestion, thank you!

PR #728, opened 2019-07-22
ignore /proc/filesystems errors on android

### LINKED ISSUE #727: disk.Partitions fails on Android 8
as of Android 8, `/proc/filesystems` is no longer readable:

```
open /proc/filesystems: permission denied
```

cc #371 #372

tmm1:
fixes #727 and makes `disk.Partitions(true)` work on Android O again.

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -244,7 +245,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 	}
 
 	fs, err := getFileSystems()
-	if err != nil {
+	if err != nil && runtime.GOOS != "android" {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


shirou:
I wonder ignoring the permission error makes `disk.Partitions(true)` work. `getFileSystems()` returns empty so no supported filesystem, nothing returns. So I agree to @Lomanic, it should handled from application side.

tmm1:
> I wonder ignoring the permission error makes `disk.Partitions(true)` work.

Yes, it works.



> `getFileSystems()` returns empty so no supported filesystem, nothing returns.

Since we're passing `all=true`, the filesystems list does not matter.

disk/disk_linux.go
```diff
@@ -244,7 +245,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 	}
 
 	fs, err := getFileSystems()
-	if err != nil {
+	if err != nil && runtime.GOOS != "android" {
```
tmm1:
```suggestion
	if err != nil && all {
```

tmm1:
I updated the PR to ignore `getFileSystem()` errors when `all=true` is passed.

shirou:
Yah! You are true. When `all=true`, `getFileSystems()` return values is not used. so it can be ignored. Great point! Thank you so much!

Issue #730, opened 2019-07-26
common.NumProc on linux reports incorrect values


traetox:
**Describe the bug**
common.NumProc on linux does not differentiate between pid dirs and other proc directories

**To Reproduce**
```go
// N/A
```

**Expected behavior**
common.NumProc only reports actual PIDs

**Environment (please complete the following information):**
 - [X] Linux: [all]

**Additional context**

shirou:
Thank you for your report! it is closed via #731

PR #731, opened 2019-07-26
Fixes issue #730

traetox:
In common.NumProcs add a loop that attempts to parse /proc/<dirs> as uint64s and only report the number that are actual base 10 integers.  This should fix the issue of improper reporting of Process counts in host.Info


shirou:
oh, indeed, you are right!  `/proc/` directory has files other than PID like `fs`, `keys`, or `kmsg` Thank you !

PR #733, opened 2019-07-31
Update disk_linux.go

eshimizu:
Modify leakage of below PullRequest.
https://github.com/shirou/gopsutil/commit/34b030b41c47513d8df6ac04de62c845c9c955a5#diff-c8dc88c5e0341d514116f32162bceff2


Lomanic:
Thanks for your vigilance, we need better CI (directly on PRs instead of only on master) with strong alerts on this kind of errors (as it's not the first time even).

PR #741, opened 2019-08-05
Fixes a problem when using Docker on a host with an encrypted LLVM root.

floren:
A docker container with a volume mounted from the host will see /dev/mapper in its mount info file, but will not be able to read it. This will allow processes running in these docker containers to just use /dev/mapper/ names instead of bailing out when it fails to read the symlink.


shirou:
Good point! thank you!

Issue #628, opened 2019-01-11
cpu.Counts


hellobinge:
**Describe the bug**
cpu.Counts return the same number no matter logical is set true or false

**To Reproduce**
```go

logicalCount, _ := cpu.Counts(true)
physicalCount, _ := cpu.Counts(false)

// and this is the source code for cpu.Counts
func Counts(logical bool) (int, error) {
	return CountsWithContext(context.Background(), logical)
}

func CountsWithContext(ctx context.Context, logical bool) (int, error) {
        // why logical is not used ?
	return runtime.NumCPU(), nil
}
```

Lomanic:
This was implemented in https://github.com/shirou/gopsutil/commit/5b605dc1a36dc1eb97e30eb01604ecae6d251f22#diff-b992981277185c467e8e9dd71d21d25dR30 by @shirou, this was probably the easiest way to implement it at the time and nobody changed this behavior afterwards… help to do so is very much welcomed!

shirou:
Because original [psutil](https://psutil.readthedocs.io/en/latest/#psutil.cpu_count) has `logical` argument. gopsutil try to implement but not yet.

Lomanic:
Rust implementation for physical cores on win7+ (with GetLogicalProcessorInformationEx) https://github.com/seanmonstar/num_cpus/pull/91/files#diff-B4AEA3E418CCDB71239B96952D9CDDB6

We should fall back to [GetLogicalProcessorInformation](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getlogicalprocessorinformation) on versions before win7.

PR #742, opened 2019-08-05
cpu: change Win32_Processor struct to use NumberOfCores

mstumpfx:
I am using gopsutil to query number of physical cores on windows machines. 

**Describe the solution**
This change will report the number of physical cores now instead of logical.
Logical number of processors can still be acquired by using the `CountsWithContext` function call in this package, which uses the go runtime package call NumCPU(), which will return the logical number of processors, per https://golang.org/pkg/runtime/#NumCPU

**Alternatives**
If there is another way to get the physical number of cores on windows using your library that I have missed, please let me know.

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -14,14 +14,14 @@ import (
 )
 
 type Win32_Processor struct {
-	LoadPercentage            *uint16
-	Family                    uint16
-	Manufacturer              string
-	Name                      string
-	NumberOfLogicalProcessors uint32
-	ProcessorID               *string
-	Stepping                  *string
-	MaxClockSpeed             uint32
+	LoadPercentage *uint16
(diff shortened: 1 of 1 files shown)


Lomanic:
Hi, thanks for your PR.
Unfortunately this is not the right thing to do for the following reasons:

* This breaks API contract for existing users of the library
* This breaks the API between the different platforms (cpu.Info() would return number of logical cores on every platform but Windows)
* The `logical` argument to `cpu.Counts()` is already made for this. I already started work implementing this parameter on Windows, related issues are #628 and  #640.

By the way, `runtime.NumCPU()`  "returns the number of logical CPUs _usable by the current process_", which can be different to the total number of logical cores on the host.

mstumpfx:
@Lomanic Thanks for the response and information, understood.

PR #747, opened 2019-08-12
Proper fix for unavailable /proc/filesystems

tmm1:
The fix in #728 was close but not correct.


tmm1:
Sorry for not testing last time. I tested this commit on Android and it is working as expected now.

shirou:
ah, sorry for not tested also. [here](https://github.com/shirou/gopsutil/pull/747/files#diff-c8dc88c5e0341d514116f32162bceff2R265) is `!all` so error checking should same as it.

PR #749, opened 2019-08-14
[cpu][windows] Add support for logical arg in Counts #640 #628

Lomanic:


shirou:
I have confirmed this works great on my Windows 10!

Lomanic:
It relies on wmi to retrieve physical cores, I will try using proper win32 APIs later.

PR #750, opened 2019-08-15
[disk][darwin] Fix #560 using github.com/lufia/iostat cgo implementation

### LINKED ISSUE #560: use of unguarded globals in disk_darwin.h can crash process under high concurrency
### Repro

On mac, call `disk.IOCountersWithContext` concurrently from many goroutines.

### Expected

Everything is fine.

### Actual

Can crash with a pretty gnarly error, like this: https://github.com/cockroachdb/cockroach/issues/27867

### Diagnosis

The functions in `disk_darwin.h` use mutable global variables that aren't guarded by a mutex: https://github.com/shirou/gopsutil/blob/master/disk/disk_darwin.h#L34-L39

### Remedy

Put some kind of a mutex around those globals or do without the globals. This [lufia/iostat](https://github.com/lufia/iostat) seems to provide the same stats without using globals: https://github.com/lufia/iostat/blob/master/iostat_darwin.c

We are temporarily switching to that library for darwin only, but would like to only have to use this library, since it provides so much other useful stuff. (Thanks!)

Lomanic:
[disk][darwin] Fix #560 using github.com/lufia/iostat cgo implementation

### DIFF
--- a/disk/disk_darwin.c
+++ b/disk/disk_darwin.c
@@ -1,0 +1,131 @@
+// https://github.com/lufia/iostat/blob/9f7362b77ad333b26c01c99de52a11bdb650ded2/iostat_darwin.c
+#include <stdint.h>
+#include <CoreFoundation/CoreFoundation.h>
+#include "disk_darwin.h"
+
+#define IOKIT	1	/* to get io_name_t in device_types.h */
+
+#include <IOKit/IOKitLib.h>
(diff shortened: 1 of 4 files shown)


Lomanic:
Tested successfully on macOS 10.12 Sierra.

shirou:
Awesome!!

PR #751, opened 2019-08-16
Add interface ifIndex support

asnowfox:
In the InterfaceStat struct. current codes don't have a ifIndex support, and I think ifIndex is a very important properity of Interface. so I suggest to add a properity named Index to InterfaceStat struct.


shirou:
It's already supported by golang net package and widely used on SNMP, so this PR is reasonable for me. Thank you for your contribution!

PR #754, opened 2019-08-24
[process] Fix #752 sort PIDS returned by process.Pids()

### LINKED ISSUE #752: PIDs returned by process.Pids() should be sorted
PIDs returned by process.Pids() should be sorted so we have consistency across different platforms (not every PIDs are sorted on Windows for example) .

This the counterpart issue of https://github.com/giampaolo/psutil/issues/1437 for gopsutil

Lomanic:
[process] Fix #752 sort PIDS returned by process.Pids()

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"math"
 	"runtime"
+	"sort"
 	"time"
 
 	"github.com/shirou/gopsutil/cpu"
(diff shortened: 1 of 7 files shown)


Lomanic:
Tested with this oneliner `for GOOS in $(go tool dist list | grep '/amd64$' | cut -d/ -f1 | sort -u); do echo $GOOS; GOOS=$GOOS go build github.com/shirou/gopsutil/process; done`

Issue #755, opened 2019-08-28
process.Percent returns incorrect value on multiple cores


Viruous:
hello:
        Some days before you push **process.go**, you change code
```
        delta_proc := t2.Total() - t1.Total()
	overall_percent := ((delta_proc / delta) * 100) * float64(numcpu)
	return overall_percent

        TO:

	delta_proc := t2.Total() - t1.Total()
	overall_percent := ((delta_proc / delta) * 100) * float64(numcpu)
	return math.Min(100, math.Max(0, overall_percent))
```
According to my observation ,this is probably  wrong.
when the process useing 60% of a six core processor ,the process.Percent return 100 to me not 6 * 100 * 60%

Lomanic:
Looks like the percent calculation was wrong all along then, as I don't think having more than 100% is expectable (it's called "Percent" after all).

Or the other way, if we assume the calculation was right before, the fix is as simple as `return math.Min(numcpu * 100, math.Max(0, overall_percent))`.

The commit you mention is https://github.com/shirou/gopsutil/commit/f4e23559a2bad85fca1eeea695331ce38ff3f494 btw (way easier to find afterwards instead of "this commit from days ago").

sajoupa:
Hi,

In general, a "percent" can totally be greater than 100 ("inflation at 250%"), or lesser than 0 ("-30% discount !"). It's just a ration, multiplied by 100.
In the context of CPU usage, the most common reference is 1 core, and percent representations often go over 100:

```
$ top -b -n 1
[...]
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
33756 libvirt+  20   0 9979.1m 4.233g  23052 S 400.0  4.5   9:53.96 qemu-system-x86
```

CPU usage is taken from /proc/<pid>/stat, which gives the amount of time that a process has been scheduled, measured in clock ticks. Several processes can be scheduled during a same clock tick, or a single process can be executed at the same time on several cores. So, I can really see no reason to cap the value at 100.

https://github.com/lrita/gopsutil/commit/98703166e094334f70eca73464708f0d0b6861ed would make things more consistent, but the reference would be the total capacity of the machine, not 1 core.

I strongly suggest https://github.com/shirou/gopsutil/commit/f4e23559a2bad85fca1eeea695331ce38ff3f494 should just be reverted.

Thanks !

[...]

PR #756, opened 2019-09-03
#755 fixes process.Percent return invalid value

lrita:
Hi @Lomanic and @shirou ,

Make it return percent of all cpu core usage which will bewteen 0 and 100.

I think this patch is match both you two's ideas and make the returned value valid.

If you want got the full core usage which was saw in `top` command, you can do this:

```go
percent, _ := process.Percent(0)
percent *= float64(runtime.NumCPU())
```

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -216,18 +216,18 @@ func (p *Process) PercentWithContext(ctx context.Context, interval time.Duration
 
 	numcpu := runtime.NumCPU()
 	delta := (now.Sub(p.lastCPUTime).Seconds()) * float64(numcpu)
-	ret := calculatePercent(p.lastCPUTimes, cpuTimes, delta, numcpu)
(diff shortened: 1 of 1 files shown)


shirou:
This PR seems break a compatibility of returned value of `PercentWithContext()` and `CPUPercent()`.

lrita:
@shirou, I think there is no any compatibility was broken in this RP.

The PR #715 break the compatibility of returned value of PercentWithContext() and CPUPercent() in really. And you accept it. But it import 2 bugs also. This PR is only fixes those bugs.

lrita:
@shirou do you want to add more collaborators for this repo, that will make this repo great and fix issue faster. I think every developer want to using a fresh upstream, that will reduce ecosystem split.

cw4t:
Now the v3 code don't work correctly on cpu_percent() either:
Please fix that, it doesn't equal to python's psutil value

PR #757, opened 2019-09-05
[darwin][process] Add cgo implementation of Exe() from PR #243

Lomanic:
Original from ppanyukov https://github.com/shirou/gopsutil/commit/447301409866fd5ca98a70339fbdcdbbdb3c280c


Lomanic:
Tested OK with CGO_ENABLED=1 and CGO_ENABLED=0 on MacOS 10.12.

shirou:
Thank you! It was not merged from 2016.

PR #758, opened 2019-09-05
[process] Propagate context internally on linux and improve tests

Lomanic:
* On windows, PID 1 doesn't exist and is not the init process so the Children test was always failing. The test now spawns a subprocess and checks it gets found by the Children function.
* For whatever reason choice.exe on windows was returning immediately during my tests, worked around it using the famous `ping localhost -n X+1` hack to have a longer running process.
* During testing, I found Connections test to be faulty, so implemented the TODO and it now opens TCP connections that are checked during the test.

If this PR gets merged I will implement the `skipIfNotImplementedErr` check for the other packages so tests run properly even on non-implemented platforms.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -70,7 +70,7 @@ func (p *Process) Ppid() (int32, error) {
 }
 
 func (p *Process) PpidWithContext(ctx context.Context) (int32, error) {
-	_, ppid, _, _, _, _, _, err := p.fillFromStat()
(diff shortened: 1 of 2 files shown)


shirou:
Awesome! without context functions such as `Process.fillFromStatm()`  are not exposed, so it is ok to delete those. and `skipIfNotImplementedErr` is a good idea!

PR #759, opened 2019-09-05
[process][windows] Implement Connections() using net.ConnectionsPid()

Lomanic:
This now makes Connections test (as changed in #758) pass successfully on windows.


shirou:
Cool!

Issue #761, opened 2019-09-06
Move non-"WithContext()" functions to platform-agnostic files


Lomanic:
There is absolutely no need to redefine the non-"WithContext()" functions on each platform when they are only calling their "WithContext()" equivalent on all platforms (same definition everywhere).

Should reduce bloat in platform-specific files, should make these files easier to manage, should more easily highlight where we don't properly pass around context between functions and we would have the same documentation for these functions on every platform  no matter the GOOS env variable godoc is running with.

I already started this work here and there when doing some refactoring, but there is still a lot to do.

shirou:
Sorry, is this means you want to remove non-"WithContext()" functions from current API?

Lomanic:
For example, instead of having [the same definition of `cpu.Times ()`](https://github.com/shirou/gopsutil/blob/86c7289/cpu/cpu_linux.go#L33-L35) on every `cpu/cpu_PLATFORM.go`, having it defined only once in `cpu/cpu.go`. The public API doesn't change. Documentation is then the same regardless of the GOOS set for running godoc.

shirou:
Now from #760, I can understand what this issue means. Yes, totally agree!

PR #762, opened 2019-09-06
[process] Fix #760 implement IsRunning by checking process with same PID has same CreateTime as current process

### LINKED ISSUE #760: Implement process.IsRunning()
Currently, process.IsRunning() returns `true` and a `ErrNotImplementedError` on every platform (`grep -A 1 'func .* IsRunningWithContext' process/*.go`). https://github.com/shirou/gopsutil/blob/e4ec7b275ada47ca32799106c2dba142d96aaf93/process/process_linux.go#L560

psutil compares creation_time (initialized during `__init__`, NewProcess() for gopsutil) to a new process with the same PID. If PIDs and creation_time are the same, the same process is still running (its PID didn't get reused by another process).

https://github.com/giampaolo/psutil/blob/3aac9483767e38995f6e5bd973e822ab0598f27a/psutil/__init__.py#L690-L695

TODO:

* add a new unexported field createTime in `process.Process` struct
* in `NewProcess()`, set `process.createTime` field to `process.CreateTime()` result
* in `process.IsRunning()`, compare `process.createTime` and `NewProcess(process.Pid).CreateTime()`

Lomanic:
[process] Fix #760 implement IsRunning by checking process with same PID has same CreateTime as current process

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -31,6 +31,7 @@ type Process struct {
 	numThreads     int32
 	memInfo        *MemoryInfoStat
 	sigInfo        *SignalInfoStat
+	createTime     int64
 
 	lastCPUTimes *cpu.TimesStat
 	lastCPUTime  time.Time
@@ -163,6 +164,7 @@ func NewProcess(pid int32) (*Process, error) {
 	if !exists {
 		return p, ErrorProcessNotRunning
 	}
+	go p.CreateTime()
 	return p, nil
 }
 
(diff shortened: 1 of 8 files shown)


shirou:
Sorry, could you resolve conflicts?

Lomanic:
Sure will do.

Lomanic:
Rebased from master

process/process.go
```diff
@@ -163,6 +164,7 @@ func NewProcess(pid int32) (*Process, error) {
 	if !exists {
 		return p, ErrorProcessNotRunning
 	}
+	go p.CreateTime()
```
shirou: If we use `go` here, create time becomes unstable. How about use `CreateTimeWithContext()` with `WithTimeout()` ?

shirou:
Great progress! And I understand what you say in #761.

process/process.go
Lomanic: I chose to spawn a goroutine so `NewProcess` performances wouldn't be impacted by this new call to `CreateTime`. I don't see how create time gets unstable with this call as it is then cached after first call.

shirou: hmmm. it depends on running environment. but I can understand. Thanks.

shirou:
Great work!

PR #764, opened 2019-09-09
ignore filesystem listing for android, not supported after android o

ilyaiqoqo:


disk/disk_linux.go
```diff
@@ -243,9 +244,16 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 		}
 	}
 
-	fs, err := getFileSystems()
-	if err != nil && !all {
-		return nil, err
+	if runtime.GOOS == "android" {
+		all = true
```
Lomanic: This breaks the function for older Android versions (where it works according to the title of the PR) and the potential error (on Android Oreo+) returned by getFileSystems should be handled by downstream users, not gopsutil. :-1: from me.

shirou:
FYI #727, #728

PR #769, opened 2019-09-17
Handle case when CPU stepping is "unknown"

robbydyer:


Lomanic:
Can you paste the content of your `/proc/cpuinfo` where you had this issue?

cpu/cpu_linux.go
```diff
@@ -168,7 +168,8 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 
 			t, err := strconv.ParseInt(val, 10, 64)
 			if err != nil {
-				return ret, err
+				// "stepping" might be 'unknown', so set to 0 in that case
+				c.Stepping = 0
```
Lomanic: This doesn't properly fix the issue as you're still setting `c.Stepping` to the result of `strconv.ParseInt` even if it's failing (thankfully this function returns 0 on error). Do the same way as for `c.Mhz`.

Issue #681, opened 2019-05-05
Run go mod vendor cause go.mod and go.sum changed


pytimer:
**Describe the bug**

I run `GO111MODULE=on go mod vendor` in my pc, it execute successfully, but it will change the `go.mod` and `go.sum` files. 

go version is `go version go1.11.4 linux/amd64`.

```sh
diff --git a/go.mod b/go.mod
index 2c21bc2..fae6f13 100644
--- a/go.mod
+++ b/go.mod
@@ -4,9 +4,9 @@ go 1.12

 require (
        github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6
-       github.com/davecgh/go-spew v1.1.1
-       github.com/go-ole/go-ole v1.2.1
-       github.com/pmezard/go-difflib v1.0.0
+       github.com/davecgh/go-spew v1.1.1 // indirect
+       github.com/go-ole/go-ole v1.2.1 // indirect
+       github.com/pmezard/go-difflib v1.0.0 // indirect
        github.com/shirou/w32 v0.0.0-20160930032740-bb4de0191aa4
        github.com/stretchr/testify v1.2.2
        golang.org/x/sys v0.0.0-20180909124046-d0be0721c37e
diff --git a/go.sum b/go.sum
index 7d37a1a..f0c5194 100644
--- a/go.sum
+++ b/go.sum
@@ -1,11 +1,14 @@
+github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6 h1:fLjPD/aNc3UIOA6tDi6QXUemppXK3P9BI7mr2hd6gx8=
 github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/go-ole/go-ole v1.2.1 h1:2lOsA72HgjxAuMlKpFiCbHTvu44PIVkZ5hqm3RSdI/E=
[...]

Issue #663, opened 2019-04-02
go.mod does not specify v2 version


hvastani:
The go.mod file in this repo does not specify the /v2 tag so if this module is imported in any project, it will be imported with +incompatible 
From the official documentation (https://github.com/golang/go/wiki/Modules)

_The module path (such as module foo/v2) declared at the start of a module's go.mod file is both:
the definitive declaration of that module's identity
the definitive declaration of how that module must be imported by consuming code_

_A module can import a v2+ package that has not opted in to modules itself. A non-module v2+ package that has a valid v2+ semver tag will be recorded with an +incompatible suffix in the importing module's go.mod file. The +incompatible suffix indicates that even though the v2+ package has a valid v2+ semver tag such as v2.0.0, the v2+ package has not actively opted in to modules and hence that v2+ package is assumed to have not been created with an understanding of the implications of Semantic Import Versioning and how to use major versions in import paths. Therefore, when operating in module mode, the go tool will treat a non-module v2+ package as an (incompatible) extension of the v1 version series of the package and assume the package has no awareness of Semantic Import Versioning, and the +incompatible suffix is an indication that the go tool is doing so._

Tested this on centos linux and saw this in my go.mod file 
` github.com/shirou/gopsutil v2.18.12+incompatible`
When I tested this, the current tag was 2.19.03
My go version 
`go version go1.11 linux/amd64`

Lomanic:
Looks like we would need a `/v2`  in the import path for it to be not flagged as incompatible (`github.com/shirou/gopsutil/v2/...`), using a `v2` branch it seems according to https://research.swtch.com/vgo-module#from_repository_to_modules

shirou:
I am not familiar with mod yet, but I think go module can adapt [git tag](https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository) and gopsutil already has version aware git tags. And I don't want to version package directory like `gopsutil/v2/disk/disk_linux.go`

ppmathis:
Please note that aside of breaking Go module standards as mentioned earlier in this issue, this repository also actively breaks the semantic versioning standard due to the patch version containing leading zeroes:

[...]

PR #772, opened 2019-09-28
remove go module settings

shirou:
from #663 (and also related to #681), I think go mod is not match for us currently. so I just remove go module settings in this PR.

When a situation changed, this change will be reverted.


jefferai:
@shirou all you needed to do was update the module path in the go.mod file. Now the library doesn't have any native handling with go modules users. Can you please revert this and just fix the go.mod file?

shirou:
I have tried update module path in go.mod, but no luck. from [this blog](https://blog.golang.org/v2-go-modules) v2 libraries with separate branches is not so supported. But I don't want to separate directories for this version. So, I decide not support go mod at least go module becomes default for all supported version (that means 1.15) because go mod breaks backward compatibility to non-go module users.

jefferai:
@shirou I think you are misunderstanding that blog. You can absolutely use different branches. You can even use the _same_ branch and simply change the module version in go.mod.

Go modules relies on tagging only. If you want to use multiple branches, that's totally supported. Simply ensure that when you cut tags that you are cutting them against the correct branch.

So for instance you could have a v1 branch with
`module github.com/shirou/gopsutil`
and any v1 tags you cut (`v1.2.3`) you cut against commits on the `v1` branch.

Then you can update `master` to have
`module github.com/shirou/gopsutil/v2`
and cut v2 tags (`v2.0.4`) against master.

When you want a v3 you can make v2 into a new branch -- _or not_, because all that actually matters is that the `go.mod` file in the commit that your tag points to has the correct version matching the tag version.

PR #775, opened 2019-10-07
[Process] Implement Groups() to get additional gids.

TakayukiBGoto:
I added Groups and fill it with additional gids. 
I think additional gids is useful information for some users.

### DIFF
--- a/process/process_openbsd.go
+++ b/process/process_openbsd.go
@@ -204,10 +204,6 @@ func (p *Process) GidsWithContext(ctx context.Context) ([]int32, error) {
 
 	return gids, nil
 }
-func (p *Process) Groups() ([]int32, error) {
-	return p.GroupsWithContext(context.Background())
-}
-
 func (p *Process) GroupsWithContext(ctx context.Context) ([]int32, error) {
 	k, err := p.getKProc()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


Lomanic:
Simply naming this function Groups has a non-negligible chance of confusing the users with the Gids function also. Maybe rename this new Groups function to SupplementaryGroups. Your POV @shirou?

shirou:
I think `Groups` is ok. Because `id` command outputs `groups` which includes supplementary groups like this
```
%id
uid=1000(shirou) gid=1000(shirou) groups=1000(shirou),4(adm),24(cdrom),27(sudo),30(dip),132(libvirtd),998(docker)
```

and other reason, Groups in `/proc/status` includes not only supplementary groups, but also primary group.

TakayukiBGoto:
Sorry for my late reply.
I fixed requested changes.

Could you decide whether to change function name or not?

Lomanic:
Let's keep function name as it is.

TakayukiBGoto:
@Lomanic Thank you for your reply.
I think I fixed all your requested change.
Please review again.

process/process_openbsd.go
```diff
@@ -204,6 +204,18 @@ func (p *Process) GidsWithContext(ctx context.Context) ([]int32, error) {
 
 	return gids, nil
 }
+func (p *Process) Groups() ([]int32, error) {
```
Lomanic: this is now defined twice for openbsd

TakayukiBGoto: Sorry, I overlooked that.
I fixed.

shirou:
Sorry for late and thank you for your contribution. Change request is already fixed, so I merge this.

PR #778, opened 2019-10-15
Fix #777 cpu.Total should not add Guest and GuestNice

### LINKED ISSUE #777: cpu.Total should not add Guest and GuestNice
`User` time includes `Guest` and `Nice` includes `GuestNice`
https://unix.stackexchange.com/a/303224

alialaee:
Fix for #777


shirou:
I didn't notice that. Thank you so much!

PR #779, opened 2019-10-15
Validate call to ps returns results

lespea:
Sometimes this panics with `index out of range [0] with length 0`

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -134,7 +134,8 @@ func (p *Process) createTimeWithContext(ctx context.Context) (int64, error) {
 	r, err := callPsWithContext(ctx, "etime", p.Pid, false)
 	if err != nil {
 		return 0, err
-	} else if len(r) == 0 || len(r[0]) == 0 {
+	}
+	if len(r) == 0 || len(r[0]) == 0 {
 		return 0, fmt.Errorf("unexpected return from ps")
 	}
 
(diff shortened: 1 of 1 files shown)


process/process_darwin.go
```diff
@@ -134,6 +134,8 @@ func (p *Process) createTimeWithContext(ctx context.Context) (int64, error) {
 	r, err := callPsWithContext(ctx, "etime", p.Pid, false)
 	if err != nil {
 		return 0, err
+	} else if len(r) == 0 || len(r[0]) == 0 {
```
Lomanic: Useless `else` as the `if` just above exits the function with a `return`.

lespea: Right but this puts the error validation checks in the same if-else block.  I think it makes it easier to read then two if blocks right after another but I can change it if you want.

Lomanic: I would prefer if you changed it, thanks :)

Lomanic:
Having bound checks like this is always good so it will be merged, do you have some context when it happened?

process/process_darwin.go
lespea: Okay this should work

lespea:
Sorry I didn't really have time to dig into exactly how this happened I just fixed the issue quick.  I might have more time later this week to try and recreate this.  Basically I'm calling this function for every processes running on a system.

Lomanic:
In fact I think this error should be handled directly in `callPsWithContext` definition (returning a non-nil err `no data returned by ps`) when no data is returned by ps, even if it returns successfully (which is surprising).

PS: you should amend your commits instead of appending new ones to have cleaner git history

Lomanic:
This is not needed anymoore after https://github.com/shirou/gopsutil/pull/790 (see also #799).

PR #782, opened 2019-10-23
Speed up ConnectionStat on Linux by omitting Uids

### LINKED ISSUE #771: `statsFromInodes` pairs fast operation(s) with slow Uid search
If you want to query ConnectionPids with a func like `ConnectionsPidWithContext` you always search all of the `ConnectionStat.Uids`.

That action to find the uids is more expensive than the rest of the call and may not be relevant to the caller.

On a server with many connections commenting out the [uid lines](https://github.com/shirou/gopsutil/blob/2f74cb51781d173ae257a61e955594064ab6f7b0/net/net_linux.go#L508-L509) halved the impact of the application, which was reaching this code in a loop.

Have an option to skip the expensive `getUids` `fillFromStatus` calls. It's worth noting that I don't consider this a significant departure from what we currently do because we already return an [empty slice](https://github.com/shirou/gopsutil/blob/2f74cb51781d173ae257a61e955594064ab6f7b0/net/net_linux.go#L611) in the case of an [error](https://github.com/shirou/gopsutil/blob/2f74cb51781d173ae257a61e955594064ab6f7b0/net/net_linux.go#L509).

I suggest a new public function that explicitly will not get Uids and a minor alteration to `statsFromInodes` to make that query optional for this new function.

**Additional**
There are other considerations for speeding up net like: #695. For this use case entirely bypassing the read will always be fastest if the caller won't use it.

Brian-Williams:
Fixes #771 

# Request

Querying Uids has a considerable cost and a caller may not 
require the information it provides. We also don't fail on not 
acquiring `Uids`, so it is an expensive operation that can 
silently fail.

This causes problems in many scenarios, where we may
spike usage if called in a cron.

I didn't get any feedback on my issue, so am happy to
close and modify if another alternative would be
considered.

### Real world usage
The following is a pprof flame graph from a real world
sever showing the relative cost:

flame graph

`fillFromStatus` calls `ReadFile` and is the majority of the
cost of a `Connections` call.
flame graph zoomed

Thanks to Chris Donaher for images.

I confirmed that modifying gopsutil to not query
`Uids` drastically improved performance with
pprof, htop, and dstat.

### Implementation
This is a way to enable altering of ConnectionStat querying w/o
having to release a new major version.

The ideal solution would likely have a `BareConnectionStat`
without a Uids field. Then we could create a public method, 
which doesn't include Uids and the existing methods would
enrich that struct by embedding in a `ConnectionStat` with
a Uids field.

### Alternative considered; future proof of implementation

I considered having an environmental variable, but
that could also cause problems because a caller
may want to query without Uids often and rarely
make the full call.

This methodology also allows us to make changes
without releasing a new version for other quirks.
For example if the `Uids` field isn't found we don't
fail, but we could add the following with this
methodology:
net/net.go...
```go
func FailWithoutUids() ConnectionStatConfigurer {
	return func(config *connectionStatConfig) {
		config.failWithoutUids = true
	}
}
```
net/net_linux.go...
```
conn.Uids, uidErr = proc.getUids()
if failOnUid {
    return nil, err
}
```

If gopsutil would accept a new public method
specifically for this use case I would be happy to
do that instead.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -46,6 +46,29 @@ type ConnectionStat struct {
 	Pid    int32   `json:"pid"`
 }
 
+// connectionStatConfig configures what to search for with ConnectionStat
+type connectionStatConfig struct {
+	_               struct{}
+	skipUids        bool
+}
+
+type ConnectionStatConfigurer func(*connectionStatConfig)
+
+func configConnectionStatConfig(config []ConnectionStatConfigurer) connectionStatConfig {
+	var c connectionStatConfig
+	for _, opt := range config {
+		opt(&c)
+	}
+	return c
+}
+
+// SkipUids doesn't collect ConnectionStat.Uids if it is significantly faster to ignore them
+func SkipUids() ConnectionStatConfigurer {
+	return func(config *connectionStatConfig) {
+		config.skipUids = true
+	}
+}
+
 // System wide stats about different network protocols
 type ProtoCountersStat struct {
 	Protocol string           `json:"protocol"`
--- a/net/net_fallback.go
+++ b/net/net_fallback.go
@@ -40,18 +40,18 @@ func ProtoCountersWithContext(ctx context.Context, protocols []string) ([]ProtoC
 	return []ProtoCountersStat{}, common.ErrNotImplementedError
 }
 
-func Connections(kind string) ([]ConnectionStat, error) {
-	return ConnectionsWithContext(context.Background(), kind)
+func Connections(kind string, config ...ConnectionStatConfigurer) ([]ConnectionStat, error) {
+	return ConnectionsWithContext(context.Background(), kind, config...)
 }
 
-func ConnectionsWithContext(ctx context.Context, kind string) ([]ConnectionStat, error) {
+func ConnectionsWithContext(ctx context.Context, kind string, config ...ConnectionStatConfigurer) ([]ConnectionStat, error) {
 	return []ConnectionStat{}, common.ErrNotImplementedError
 }
 
-func ConnectionsMax(kind string, max int) ([]ConnectionStat, error) {
-	return ConnectionsMaxWithContext(context.Background(), kind, max)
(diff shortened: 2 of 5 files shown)


shirou:
Thank you for very interesting PR. I can understand getting Uid make API slower and Uid is not required for everyone. So, introduce config is understandable.

However, adding `config` as variadic style is not same current gopsutil method invocation style. gopsutil is very old go project (more than 6years) and at first there are no such a style...
I know adding a varadic argument does not break compatibility. So it is ok in functionally. However once adding such a type of method definition, I am afraid of a chaos such as method A has config but method B doesn't have.

Adding `WihoutUid` functions are more conservative. But adding config argument is more extendable. What do you think ? > @Lomanic 

(By the way, I am planning changing to NetLink API on linux by using [gosigar](https://github.com/elastic/gosigar/tree/master/sys/linux), in a few months)

Brian-Williams:
Hey thanks for reviewing so quickly. I'm excited for the NetLink change!

I chose configuration over paired methods only because the current implementation of fetching `Uid` ignores errors and I could see another request to config that. However, if you are planning an implementation change, that may not be relevant.

I understand the fear of config creep. In the wider scope of gopsutil it could become a headache to maintain. As you suggested pairing `<func>WithoutUid` for `Connections` and it's downstream functions may be best.

I have some questions about how we would implement that:

- Would we want to have a new struct w/o a `Uid` field?
- Would we accept a `goto` loop if it would allow us to inline the private funcs?
- My NetLink knowledge is lacking, would these funcs become obsolete?

Lomanic:
> Adding WihoutUid functions are more conservative. But adding config argument is more extendable. What do you think ? > @Lomanic

Yes I agree, let's go for `WithoutUids` function(s).

> Would we want to have a new struct w/o a Uids field?

We can keep the same type/struct, only its Uids field won't be populated. 

> Would we accept a goto loop if it would allow us to inline the private funcs?

I think I'm OK with this if it's not too complicated, would have to see the code to judge.

PR #783, opened 2019-10-28
Add WithoutUids funcs for Connections

### LINKED ISSUE #771: `statsFromInodes` pairs fast operation(s) with slow Uid search

Brian-Williams:
Fixes #771

See previous discussion in alternative PR #782

# Request

Add `WithoutUids` functions for `Connections` and downstream funcs
for performance.

### Considerations
This PR adds a lot of public functions because of the combination of
max, pid, and withoutUids. I'd be happy to strip these down to expose
without only in specific combinations.

`ConnectionsPidWithContext` and `ConnectionsPidMaxWithContext` had
duplicate logic.

The only difference I saw was their error messages for not finding
pid(s).

`ConnectionsPidWithContext` was:
`return nil, fmt.Errorf("cound not get pid(s), %d: %s", pid, err)`

`ConnectionsPidMaxWithContext` was:
`return nil, fmt.Errorf("cound not get pid(s), %d", pid)`

I chose to use the one that propagated the error for
`connectionsPidMaxWithoutUidsWithContext`.

If this duplication was for optimization reasons this change does
make a handful of funcs no longer inline. We can maintain a shorter
call stack by directly calling
`connectionsPidMaxWithoutUidsWithContext`. I'm not certain if that
was the reason. I chose not to do so in this PR to keep with
convention over compiler optimizations.

Inlining comparison used:
```
docker run --rm -v "$PWD":/go/src/github.com/shirou/gopsutil -it -w /go/src/github.com/shirou/gopsutil golang:1.13 go build -gcflags "-m=2" net/net.go net/net_linux.go | grep inline
```
Left: master@0a33b6d8d2995226d785bdb0f0f066edbd71ff16
Right: this branch
https://www.diffchecker.com/UK3dX45h

### DIFF
--- a/net/net_linux.go
+++ b/net/net_linux.go
@@ -401,32 +401,36 @@ func ConnectionsMaxWithContext(ctx context.Context, kind string, max int) ([]Con
 	return ConnectionsPidMax(kind, 0, max)
 }
 
+// Return a list of network connections opened, omitting `Uids`.
+// WithoutUids functions are reliant on implementation details. They may be altered to be an alias for Connections or be
+// removed from the API in the future.
+func ConnectionsWithoutUids(kind string) ([]ConnectionStat, error) {
+	return ConnectionsWithoutUidsWithContext(context.Background(), kind)
+}
+
+func ConnectionsWithoutUidsWithContext(ctx context.Context, kind string) ([]ConnectionStat, error) {
+	return ConnectionsMaxWithoutUidsWithContext(ctx, kind, 0)
+}
+
+func ConnectionsMaxWithoutUidsWithContext(ctx context.Context, kind string, max int) ([]ConnectionStat, error) {
+	return ConnectionsPidMaxWithoutUidsWithContext(ctx, kind, 0, max)
+}
+
 // Return a list of network connections opened by a process.
 func ConnectionsPid(kind string, pid int32) ([]ConnectionStat, error) {
 	return ConnectionsPidWithContext(context.Background(), kind, pid)
 }
 
+func ConnectionsPidWithoutUids(kind string, pid int32) ([]ConnectionStat, error) {
+	return ConnectionsPidWithoutUidsWithContext(context.Background(), kind, pid)
+}
+
 func ConnectionsPidWithContext(ctx context.Context, kind string, pid int32) ([]ConnectionStat, error) {
-	tmap, ok := netConnectionKindMap[kind]
-	if !ok {
-		return nil, fmt.Errorf("invalid kind, %s", kind)
-	}
-	root := common.HostProc()
-	var err error
-	var inodes map[string][]inodeMap
-	if pid == 0 {
-		inodes, err = getProcInodesAll(root, 0)
-	} else {
-		inodes, err = getProcInodes(root, pid, 0)
-		if len(inodes) == 0 {
-			// no connection for the pid
-			return []ConnectionStat{}, nil
-		}
-	}
-	if err != nil {
-		return nil, fmt.Errorf("cound not get pid(s), %d: %s", pid, err)
-	}
-	return statsFromInodes(root, pid, tmap, inodes)
+	return ConnectionsPidMaxWithContext(ctx, kind, pid, 0)
+}
+
+func ConnectionsPidWithoutUidsWithContext(ctx context.Context, kind string, pid int32) ([]ConnectionStat, error) {
+	return ConnectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, 0)
 }
 
 // Return up to `max` network connections opened by a process.
@@ -434,7 +438,19 @@ func ConnectionsPidMax(kind string, pid int32, max int) ([]ConnectionStat, error
 	return ConnectionsPidMaxWithContext(context.Background(), kind, pid, max)
 }
 
+func ConnectionsPidMaxWithoutUids(kind string, pid int32, max int) ([]ConnectionStat, error) {
+	return ConnectionsPidMaxWithoutUidsWithContext(context.Background(), kind, pid, max)
+}
+
 func ConnectionsPidMaxWithContext(ctx context.Context, kind string, pid int32, max int) ([]ConnectionStat, error) {
+	return connectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, max, false)
+}
+
+func ConnectionsPidMaxWithoutUidsWithContext(ctx context.Context, kind string, pid int32, max int) ([]ConnectionStat, error) {
+	return connectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, max, true)
+}
+
+func connectionsPidMaxWithoutUidsWithContext(ctx context.Context, kind string, pid int32, max int, skipUids bool) ([]ConnectionStat, error) {
 	tmap, ok := netConnectionKindMap[kind]
 	if !ok {
 		return nil, fmt.Errorf("invalid kind, %s", kind)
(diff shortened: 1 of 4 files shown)


net/net_linux.go
```diff
@@ -401,40 +401,54 @@ func ConnectionsMaxWithContext(ctx context.Context, kind string, max int) ([]Con
...
+func ConnectionsPidMaxWithoutUids(kind string, pid int32, max int) ([]ConnectionStat, error) {
+	return ConnectionsPidMaxWithoutUidsWithContext(context.Background(), kind, pid, max)
+}
+
 func ConnectionsPidMaxWithContext(ctx context.Context, kind string, pid int32, max int) ([]ConnectionStat, error) {
+	return connectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, max, false)
+}
+
+func ConnectionsPidMaxWithoutUidsWithContext(ctx context.Context, kind string, pid int32, max int) ([]ConnectionStat, error) {
+	return connectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, max, true)
```
shirou: This invoked *without* Uids. So this becomes incompatible from current. Current returns with uids.
```suggestion
	return connectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, max, false)
```

shirou:
Thank you for a PR. I will merge this after fixing an issue which is compatibility. 
But please forgive me to add a warning to comments like `"These methods is not stable. After implementing more faster way such as NetLink, these methods are removed from API."`.

And I am trying to start NetLink API (but sorry, I can work with this OSS only weekend, it takes a months...)

net/net_linux.go
Brian-Williams: I chose a negative `skipUids` for the 5th parameter of `connectionsPidMaxWithoutUidsWithContext`, so that `false` would be the original implementation of not skipping Uids. `true` causes the `Uids` fields not to be populated. When `false` the `Uids` field will be populated or an empty slice. When `true` the field will always be a nil slice.

I had a mistake where `ConnectionsPidWithContext` was calling a `ConnectionsPidMaxWithoutUidsWithContext`. I have corrected that to call `ConnectionsPidMaxWithContext`.

Here is a small proof of work:
```shell
# cat test/main.go 
package main

import (
	"github.com/shirou/gopsutil/net"
	"fmt"
)

func main() {
	v, _ := net.Connections("inet")
	fmt.Printf("Connections: %s\n", v)
	v2, _ := net.ConnectionsWithoutUids("inet")
	fmt.Printf("ConnectionsWithoutUids: %s\n", v2)
}
# go run test/main.go 
Connections: [{"fd":0,"family":2,"type":1,"localaddr":{"ip":"0.0.0.0","port":58841},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"LISTEN","uids":[],"pid":0} {"fd":0,"family":2,"type":1,"localaddr":{"ip":"0.0.0.0","port":111},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"LISTEN","uids":[],"pid":0} {"fd":0,"family":10,"type":1,"localaddr":{"ip":"::","port":42509},"remoteaddr":{"ip":"::","port":0},"status":"LISTEN","uids":[],"pid":0} {"fd":0,"family":10,"type":1,"localaddr":{"ip":"::","port":111},"remoteaddr":{"ip":"::","port":0},"status":"LISTEN","uids":[],"pid":0} {"fd":0,"family":10,"type":1,"localaddr":{"ip":"::","port":8080},"remoteaddr":{"ip":"::","port":0},"status":"LISTEN","uids":[],"pid":0} {"fd":0,"family":2,"type":2,"localaddr":{"ip":"127.0.0.1","port":996},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"NONE","uids":[],"pid":0} {"fd":0,"family":2,"type":2,"localaddr":{"ip":"0.0.0.0","port":111},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"NONE","uids":[],"pid":0} {"fd":0,"family":2,"type":2,"localaddr":{"ip":"0.0.0.0","port":40170},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"NONE","uids":[],"pid":0} {"fd":0,"family":10,"type":2,"localaddr":{"ip":"::","port":40020},"remoteaddr":{"ip":"::","port":0},"status":"NONE","uids":[],"pid":0} {"fd":0,"family":10,"type":2,"localaddr":{"ip":"::","port":111},"remoteaddr":{"ip":"::","port":0},"status":"NONE","uids":[],"pid":0}]
ConnectionsWithoutUids: [{"fd":0,"family":2,"type":1,"localaddr":{"ip":"0.0.0.0","port":58841},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"LISTEN","uids":null,"pid":0} {"fd":0,"family":2,"type":1,"localaddr":{"ip":"0.0.0.0","port":111},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"LISTEN","uids":null,"pid":0} {"fd":0,"family":10,"type":1,"localaddr":{"ip":"::","port":42509},"remoteaddr":{"ip":"::","port":0},"status":"LISTEN","uids":null,"pid":0} {"fd":0,"family":10,"type":1,"localaddr":{"ip":"::","port":111},"remoteaddr":{"ip":"::","port":0},"status":"LISTEN","uids":null,"pid":0} {"fd":0,"family":10,"type":1,"localaddr":{"ip":"::","port":8080},"remoteaddr":{"ip":"::","port":0},"status":"LISTEN","uids":null,"pid":0} {"fd":0,"family":2,"type":2,"localaddr":{"ip":"127.0.0.1","port":996},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"NONE","uids":null,"pid":0} {"fd":0,"family":2,"type":2,"localaddr":{"ip":"0.0.0.0","port":111},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"NONE","uids":null,"pid":0} {"fd":0,"family":2,"type":2,"localaddr":{"ip":"0.0.0.0","port":40170},"remoteaddr":{"ip":"0.0.0.0","port":0},"status":"NONE","uids":null,"pid":0} {"fd":0,"family":10,"type":2,"localaddr":{"ip":"::","port":40020},"remoteaddr":{"ip":"::","port":0},"status":"NONE","uids":null,"pid":0} {"fd":0,"family":10,"type":2,"localaddr":{"ip":"::","port":111},"remoteaddr":{"ip":"::","port":0},"status":"NONE","uids":null,"pid":0}]
```

Brian-Williams:
> I will merge this after fixing an issue which is compatibility.
Yes, I apologize for the mistake. The problem was in `ConnectionsPidWithContext`:
```
-       return ConnectionsPidMaxWithoutUidsWithContext(ctx, kind, pid, 0)
+       return ConnectionsPidMaxWithContext(ctx, kind, pid, 0)
 }
```

> But please forgive me to add a warning to comments like `"These methods is not stable. After implementing more faster way such as NetLink, these methods are removed from API."`.

Great point!  I've added the following to the top level `ConnectionsWithoutUids` funcs:
```
// WithoutUids functions are reliant on implementation details. They may be altered to be an alias for Connections or be
// removed from the API in the future.
```

net/net_linux.go
Brian-Williams: For context the original name for `connectionsPidMaxWithoutUidsWithContext` was `connectionsPidMaxWithContextUidsOpt`. I can change it back if that is a better name.

shirou:
Thank you for updating! I confirmed it does not break a value compatibility. And I appreciate that understanding about future API remove. Great contributing!

PR #785, opened 2019-11-04
Use golang.org/x/windows functions instead of github.com/shirou/w32

tklauser:
All functions used from github.com/shirou/w32 are also available from
golang.org/x/sys/windows which is already used in other places. Convert
the remaining usages to use the functions from x/sys/windows.


tklauser:
CircleCI failure is due to commit golang/sys@ef33b2fb2c414dca4bfcf6ca59cbd1d9ad00703d. Go 1.9 used for CircleCI container doesn't seem to support go:linkname properly yet.

shirou:
Awesome suggestion, thank you! I have updated cirleci version to 1.11 #786 . Could you try with latest commit?

tklauser:
Sure, rebased onto latest master. Thanks.

tklauser:
This is golang/go#23311 which still exists with go 1.11, but was fixed with go 1.12 and newer. Not sure if upgrading to circleci/golang:1.12 is an option. Otherwise we might try to add an `empty.s` file in `golang.org/x/sys/windows`.

shirou:
Currently go 1.11 is supported version. We can not change to 1.12 until 1.14 is released. Could you try to add the `empty.s` in PR? Please add "this files is copied from .... and will be removed when 1.12 supported" at top of the file.

tklauser:
Unfortunately, since the build failure is in `golang.org/x/sys/windows` we cannot just add the file in this PR. Thus, I've sent https://go-review.googlesource.com/c/sys/+/206357 and hope it will get accepted. Once it is accepted, I can update this PR vendoring `golang.org/x/sys/windows` containing said fix and all should build fine.

shirou:
Great!! Thank you so much!

tklauser:
The fix for `x/sys/windows` was merged as golang/sys@d32e6e3b99 and I updated this PR to include it. The build looks fine now.

shirou:
So quick! This is great contribution not only this library but also Go ecosystem!

After Go's next release, I will merge this PR. Thank you so much!

tklauser:
Thanks.

FWIW, I don't think you even have to wait for Go's next release to get the fix in `golang.org/x/sys/windows`. It's developed independently and integrated in its new version as a module in this PR. It also builds with Go 1.9, which seems to be the minimum version `gopsutil` requires according to the README. Tested using:

```
go get golang.org/dl/go1.9
go1.9 download
cd $GOPATH/github.com/shirou/gopsutil
sed -ie 's/go /go1.9 /g' Makefile
make build_test
```

shirou:
sorry to late. but Thank you!

Issue #737, opened 2019-08-02
gopsutil missing a way to retrieve CPU architecture like in 'uname -m'


tmm1:
I was surprised to see that the equivalent of `uname -m` is not available anywhere.

Lomanic:
There is `runtime.GOARCH` though it only returns the arch the binary has been compiled for, which can be different from real CPU running it on OSes able to run both 32 and 64 bit binaries

You may look at the [`cpu.Info()`](https://godoc.org/github.com/shirou/gopsutil/cpu#Info) function, though it doesn't result the same thing as `uname -m`. You may check [github.com/klauspost/cpuid](https://github.com/klauspost/cpuid) for CPUs with the CPUID instruction (ARM not supported).

Maybe I overlooked it, but indeed it doesn't seem there is a way to get hardware architecture like `uname -m` (or runtime.GOARCH) in gopsutil. Not sure if it should be obtained in `cpu.Info()` or [`host.Info()`](https://godoc.org/github.com/shirou/gopsutil/host#Info).

tmm1:
The cpu package provides some hints, but I'm more interested in the OS architecture. For instance on aarch64 process its still possible to run kernel/OS in armv8l mode.

I think we should add the equivalent of `uname -m` to `host.Info()`. This may be same as `runtime.GOARCH` in most cases, but not all.

Lomanic:
Implementations in python
https://github.com/pydata/numexpr/blob/master/numexpr/cpuinfo.py
https://github.com/workhorsy/py-cpuinfo/blob/master/cpuinfo/cpuinfo.py

The objective to me is to return the same result as runtime.GOARCH, while really querying at runtime instead of relying on a constant set at build-time.

Possible values (obtained with `go tool dist list | cut -d/ -f2 | sort -u`)

```
386
amd64
amd64p32
arm
arm64
mips
mips64
mips64le
mipsle
ppc64
ppc64le
s390x
wasm
```

Lomanic:
## Notes for implementing

On Unix, use uname syscall like below and compare returned value to [this list](https://stackoverflow.com/a/45125525) to do the mapping with the possible GOARCH values above

```go
package main

import (
	"fmt"

	"golang.org/x/sys/unix"
)

func main() {
	var utsname unix.Utsname
	unix.Uname(&utsname)
	fmt.Println(string(utsname.Machine[:]))
}
```

[...]

PR #794, opened 2019-11-27
Add CPU architecture to InfoStat

albertvaka:


Lomanic:
runtime.GOARCH is not host CPU arch but the architecture the binary was compiled for.

See the discussion at #737 and in particular https://github.com/shirou/gopsutil/issues/737#issuecomment-517806622 and https://github.com/shirou/gopsutil/issues/737#issuecomment-517946949 for a more detailed explanation and how it should be implemented.

PR #798, opened 2019-12-03
add HOST_RUN config description

like-inspur:
there are five system environment variable can be config in common/common.go, and HOST_RUN is not documented


Lomanic:
Good catch, thanks for this!

Issue #797, opened 2019-12-03
Restore temperature info for macOS


adamsmith:
macOS used to support temperature readings. Support for this was added in https://github.com/shirou/gopsutil/pull/334 but later reverted in https://github.com/shirou/gopsutil/commit/c95755e4bcd7a62bb8bd33f3a597a7c7f35e2cf3 because the original PR included GPL-covered code ([issue](https://github.com/shirou/gopsutil/issues/518)).

**Describe the solution you'd like**
I believe this functionality can be added back using open source code. Specifically, there is C code that reads the same information from the Apple SMC [here](https://github.com/zengxs/apple-smc/blob/master/module/smc.c).

If in the future you want to add support for fan speeds, there's a C implementation of that [here](https://github.com/poporul/stats/blob/master/smc.c).

(I found these by searching Github-wide for `SMCParamStruct NOT MERCHANTABILITY`.)

Thanks!

PR #800, opened 2019-12-13
Restored temperature info for macOS

Girbons:
reintroduced temperature info for macOS following what was proposed in #797.

Tested on my macOS seems to work
Screenshot 2019-12-16 at 11 24 18

### DIFF
--- a/host/host_darwin.go
+++ b/host/host_darwin.go
@@ -223,11 +223,3 @@ func KernelVersionWithContext(ctx context.Context) (string, error) {
 	version, err := unix.Sysctl("kern.osrelease")
 	return strings.ToLower(version), err
 }
-
-func SensorsTemperatures() ([]TemperatureStat, error) {
-	return SensorsTemperaturesWithContext(context.Background())
-}
-
-func SensorsTemperaturesWithContext(ctx context.Context) ([]TemperatureStat, error) {
-	return []TemperatureStat{}, common.ErrNotImplementedError
-}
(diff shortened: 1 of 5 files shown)


Lomanic:
Thanks for this great PR, I only have a hackintosh VM to test it so all values are set to 0 but they are all the same as the last GPL version from 63047d7. Could you fix these compilation warnings, seen for example when i ran the tests with `go test github.com/shirou/gopsutil/host`?

```
In file included from host/host_darwin_cgo.go:7:
host/include/smc.c:77:5: warning: implicitly declaring library function 'printf' with type 'int (const char *, ...)' [-Wimplicit-function-declaration]
host/include/smc.c:77:5: note: include the header <stdio.h> or explicitly provide a declaration for 'printf'
host/include/smc.c:93:7: warning: implicitly declaring library function 'strlen' with type 'unsigned long (const char *)' [-Wimplicit-function-declaration]
host/include/smc.c:93:7: note: include the header <string.h> or explicitly provide a declaration for 'strlen'
host/include/smc.c:124:3: warning: implicitly declaring library function 'memset' with type 'void *(void *, int, unsigned long)' [-Wimplicit-function-declaration]
host/include/smc.c:124:3: note: include the header <string.h> or explicitly provide a declaration for 'memset'
host/include/smc.c:151:3: warning: implicitly declaring library function 'memcpy' with type 'void *(void *, const void *, unsigned long)' [-Wimplicit-function-declaration]
host/include/smc.c:151:3: note: include the header <string.h> or explicitly provide a declaration for 'memcpy'
```

Lomanic:
Thanks again @Girbons

PR #802, opened 2019-12-17
Add support for freebsd/arm64

dmgk:
Also update golang.org/x/sys to the latest master to bring in freebsd/arm64 bits.


Lomanic:
Thanks a lot for this PR!

As I don't have any arm64 device with freebsd installable on it I will just assume it works and is tested. I will merge as-is unless @shirou has a device he can test on.

shirou:
Me neither. But I a user found any issues, they will open an issue. So I merge this. Thank you so much for this awesome PR!

PR #803, opened 2019-12-20
rename InActive(file) to Inactive(file) in mem_linux.go

lanixzcj:
FIX:availlable mem is caculated incorrectly in Linux 2.6.32.


lanixzcj:
The detail info is in #804 
Please review ,Thank you~

shirou:
Ah! Good catch! `Inactive(file)` not `InActive(file)`also in my Linux 4.15! Thank you!

PR #808, opened 2019-12-25
Change access right for GetProcessTimes call

MashaSamoylova:
GetProcessTimes requires lower rights.  Now you can get times from the system process, for example from the process with PID 4.

See https://devblogs.microsoft.com/oldnewthing/20151210-00/?p=92451


Lomanic:
LGTM, let me double check.

PR #810, opened 2020-01-03
Use Getstatfs from golang.org/x/sys/unix for 64-bit inode support on FreeBSD 12

tklauser:
Use unix.Getstatfs and its associated Statfs_t type instead of
implementing them locally in this package. This allows to use 64-bit
inode fields on FreeBSD 12 while still keeping backwards compatibility
for old FreeBSD versions, as unix.Getfsstat will use the correct syscall
number and data structure version and convert its result
correspondingly.

Also see https://golang.org/cl/136816 for details.

### DIFF
--- a/disk/disk_freebsd.go
+++ b/disk/disk_freebsd.go
@@ -8,7 +8,6 @@ import (
 	"encoding/binary"
 	"path"
 	"strconv"
-	"unsafe"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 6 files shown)


Lomanic:
Tested OK, thanks a lot, do you plan to create the same MR for the other BSDs darwin and openbsd?

tklauser:
I'll check darwin and openbsd and will submit PRs for them too.

Lomanic:
Thanks a lot, in the mean time I started doing the same for darwin and openbsd but your MRs are more complete indeed.

PR #812, opened 2020-01-07
Use Getfsstat from golang.org/x/sys/unix on Darwin

tklauser:
Starting with Go 1.12, direct syscalls on darwin are no longer
supported. Instead, libSystem is used when making syscalls. See
https://golang.org/doc/go1.12#darwin

In order to still support Getfsstat, use the syscall wrapper and types
from golang.org/x/sys/unix which uses the correct syscall method
depending on the Go version.

Also use the correct MNT_* consts and their respective strings according
to the mount(8) manpage.

Follow-up for #810


shirou:
I haven't checked it is work on actual Darwin, but source seems good and this is a cool job. it must be worked (I hope will someone open an issue if does not works)

PR #814, opened 2020-01-08
Use Getfsstat from golang.org/x/sys/unix on OpenBSD

tklauser:
Use the syscall wrapper and types from golang.org/x/sys/unix instead of
implementing them locally.

Also remove unused generated types and consts.

Follow-up for #810 and #812


shirou:
Great!! Thank you so much!

Issue #817, opened 2020-01-14
[disk], resolving symlinks on /dev/mapper not respecting paths when running in container


i-prudnikov:
**Describe the bug**
This line is trying to evaluate link not respecting the case when running in container.
https://github.com/shirou/gopsutil/blob/master/disk/disk_linux.go#L304

**Expected behavior**
Disk statistics can be gathered when running in container and host fs is mounted inside container.

**To Reproduce**
https://github.com/influxdata/telegraf/issues/6898

**Environment (please complete the following information):**
 - [X] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
```
Linux <HOSTNAME> 4.4.0-154-generic #181-Ubuntu SMP Tue Jun 25 05:29:03 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

```
root@logs-02:~# cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

**Possible solution:**
 Either to remove the entire block https://github.com/shirou/gopsutil/blob/master/disk/disk_linux.go#L303-L308 (this is preferable as symlink contains the name of LVM and it is not making sense to eval it to the name of device) or to change the above-mentioned line line to:
```go
devpath, err := filepath.EvalSymlinks(common.GetEnv("HOST_MOUNT_PREFIX","/",d.Device))
```

Lomanic:
#819 was merged, closing (FYI,  @i-prudnikov, you can [set your commit message with the right keywords](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) to automatically close the referenced issue)

PR #686, opened 2019-05-13
Tl/check for logical volumes

tonylambiris:
Adds logic when dealing with device-mapped volumes (ie: /dev/dm-0). Also because the device label is present, this can be mapped back to `/dev/mapper/[label]` for more processing if needed.

Here is a snippet of code that demonstrates with this fix you are now able to get devices that may have been setup using LVM: https://gist.github.com/tonylambiris/86c39693e214a926757c47fc0c9d03ba

Output:
```
➜ go run disk.go
INFO[0000] ([]disk.PartitionStat) (len=6 cap=41) {
 (disk.PartitionStat) {"device":"/dev/dm-4","mountpoint":"/","fstype":"btrfs","opts":"rw,noatime"},
 (disk.PartitionStat) {"device":"/dev/dm-3","mountpoint":"/boot","fstype":"ext4","opts":"rw,noatime"},
 (disk.PartitionStat) {"device":"/dev/nvme0n1p1","mountpoint":"/boot/efi","fstype":"vfat","opts":"rw,noatime"},
 (disk.PartitionStat) {"device":"/dev/dm-0","mountpoint":"/home","fstype":"btrfs","opts":"rw,noatime"},
 (disk.PartitionStat) {"device":"/dev/sdd1","mountpoint":"/run/media/tlambiris/T5","fstype":"ext4","opts":"rw,nosuid,nodev,relatime"},
 (disk.PartitionStat) {"device":"/dev/sdc2","mountpoint":"/run/media/tlambiris/FANTOM","fstype":"hfsplus","opts":"rw,nosuid,nodev,relatime"}
} 
INFO[0000] (map[string]disk.IOCountersStat) (len=1) {
 (string) (len=4) "dm-4": (disk.IOCountersStat) {"readCount":79327,"mergedReadCount":0,"writeCount":226891,"mergedWriteCount":0,"readBytes":6419386880,"writeBytes":3975909376,"readTime":32746,"writeTime":841434,"iopsInProgress":0,"ioTime":125154,"weightedIO":874260,"name":"dm-4","serialNumber":"","label":"arch-root"}
} 
INFO[0000] (map[string]disk.IOCountersStat) (len=1) {
[...]

PR #819, opened 2020-01-17
Fix for track disk stats when running in container

i-prudnikov:
The commented block has 2 problems, so should be disabled:
1. Will not work, in case we track disk info on the host from container, as /dev will be mapped into other mount point
2. Useful info about vg name removed if we follow symlink, here the example:
```
~# ls -la /dev/mapper/dm--01--vg-root
lrwxrwxrwx 1 root root 7 Oct  3 10:47 /dev/mapper/dm--01--vg-root -> ../dm-0
```
Issue is opened:
https://github.com/shirou/gopsutil/issues/817

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -300,12 +300,18 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 				}
 			}
 
-			if strings.HasPrefix(d.Device, "/dev/mapper/") {
-				devpath, err := filepath.EvalSymlinks(d.Device)
-				if err == nil {
-					d.Device = devpath
-				}
-			}
+			//The block below has 2 problems, so should be disabled:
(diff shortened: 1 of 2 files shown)


shirou:
related #686

Thanks for the issue! your are right. we should add `HOST_MOUNT_PREFIX` to every device mounted path. 

so could you change to add `common.GetEnv("HOST_MOUNT_PREFIX"` as you suggested, instead of deleting ?

```
devpath, err := filepath.EvalSymlinks(common.GetEnv("HOST_MOUNT_PREFIX","/",d.Device))
```

Lomanic:
What about a `HOST_DEV` mapping `/dev` variable like the other ones?

i-prudnikov:
Implemented with HOST_DEV

shirou:
It looks nice for me! Thank you for your contribution!

(I will add `HOST_DEV` to README)

Issue #816, opened 2020-01-14
[process] ineffective detection of process executable name in linux


i-prudnikov:
**Describe the bug**
In linux, process executable is detected by evaluating `/proc/<PID>/exe`
https://github.com/shirou/gopsutil/blob/bcf28f0c372c3cdfcfda49420468d1584dfe8f2c/process/process_linux.go#L812
In some cases, this symbolic link is not in a proper state, that leads to error when resolving it, while process is still running

**To Reproduce**
```go
import "github.com/shirou/gopsutil/process"
proc,err := process.NewProcess(5)
name, err := proc.Exe()
```
```
root@logs-02:~# ls /proc/5
ls: cannot read symbolic link '/proc/5/exe': No such file or directory
attr       clear_refs       cpuset   fd       limits     mem         net        oom_score      projid_map  sessionid  stat     task
autogroup  cmdline          cwd      fdinfo   loginuid   mountinfo   ns         oom_score_adj  root        setgroups  statm    timers
auxv       comm             environ  gid_map  map_files  mounts      numa_maps  pagemap        sched       smaps      status   uid_map
cgroup     coredump_filter  exe      io       maps       mountstats  oom_adj    personality    schedstat   stack      syscall  wchan

root@logs-02:~# ls -la /proc/5/exe
ls: cannot read symbolic link '/proc/5/exe': No such file or directory
lrwxrwxrwx 1 root root 0 Jan 14 13:05 /proc/5/exe
```

**Expected behavior**
Get valid executable name

**Environment (please complete the following information):**
 - [X] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]

```
Linux <HOSTNAME> 4.4.0-154-generic #181-Ubuntu SMP Tue Jun 25 05:29:03 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

```
root@logs-02:~# cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

**Possible solution:**
[...]

PR #820, opened 2020-01-17
add process name parsing form /proc/<PID>/stat, when /proc/<PID>/exe has broken symlink

i-prudnikov:
To fix this issue:
https://github.com/shirou/gopsutil/issues/816

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -85,15 +85,7 @@ func (p *Process) Name() (string, error) {
 func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 	if p.name == "" {
 		if err := p.fillFromStatusWithContext(ctx); err != nil {
-			statPath := common.HostProc(strconv.Itoa(int(p.Pid)), "stat")
-			contents, err := ioutil.ReadFile(statPath)
-			if err != nil {
-				return "", err
-			}
-			data := string(contents)
-			binStart := strings.IndexRune(data, '(') + 1
-			binEnd := strings.IndexRune(data[binStart:], ')')
-			p.name = data[binStart : binStart+binEnd]
+			return "", err
 		}
 	}
 	return p.name, nil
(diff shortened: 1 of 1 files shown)


Lomanic:
As I've told in the original issue, this trims long process names. Do we want really want this @shirou?

i-prudnikov:
Well, from my perspective, it uses stat only in case there is a problem with /proc/<PID>/exe, it is better than fail.

i-prudnikov:
@shirou , @Lomanic:  it seems I mixed some stuff and put changes in the wrong place, currently moved changes into `fillFromExeWithContext` function in `process_linux.go`. Sorry for confusion.

The proposed changes allow to correctly provide the name of executable in the following cases:
1. When symlink is broken in `/HOST_FS/proc/PID/exe`
2. When it is not enough access rights to follow symlink `/HOST_FS/proc/PID/exe`. This can be a case when running in non-privileged container. In this case the contents of /proc filesystem is accessible, but following symlinks is not allowed:
```
#From container (in non-privileged mode):
root@logs-02:/# ls -la /hostfs/proc/1/exe
ls: cannot read symbolic link '/hostfs/proc/1/exe': Permission denied
lrwxrwxrwx 1 root root 0 Jan 21 10:48 /hostfs/proc/1/exe

#From host:
root@logs-02:~# ls -la /proc/1/exe
lrwxrwxrwx 1 root root 0 Jan 21 10:48 /proc/1/exe -> /lib/systemd/systemd
```

Lomanic:
Sorry but this is worse than before your last commit as your fallback is now returning a process _name_ in the `Exe()` function, as I wrote in https://github.com/shirou/gopsutil/issues/816#issuecomment-575400949. If we were to merge this, your first commit 2640a014c1953878f0c014bf5928ebce67ea54f4 was the right path, not your last one.

Please also squash your commits as to keep history clean.

i-prudnikov:
@Lomanic, I think that there is a misunderstanding here, could you please clarify?
1. The initial commit, in my point of view, is not a right path because I was trying to fix this function:
```go
	func (p *Process) NameWithContext(ctx context.Context) (string, error) {
		if p.name == "" {
			if err := p.fillFromStatusWithContext(ctx); err != nil {
				return "", err
			}
		}
		return p.name, nil
	}
```

This function calls `fillFromStatusWithContext(ctx)` that have no connection with the issue that this MR supposed to fix - #816. The problem comes from broken link in `/proc/(pid)/exe`, and `fillFromStatusWithContext(ctx)` is actually not touching it at all, it works with `/proc/(pid)/status` instead. And, there is no issue with `/proc/(pid)/status` at all, even if `/proc/(pid)/exe` is a broken symlink.

2. AFAIU, the issue #816 comes from `func (p *Process) fillFromExeWithContext(ctx context.Context) (string, error)` as it reads `/proc/(pid)/exe`. The proposed fix is a fall back, it won't give you the full executable path, but at least executable name. To my understanding, this is better than fail.

Are we on the same page or I missed something?

Lomanic:
Sorry if I wasn't clear, but I understand your point of view now with your last post.

My point is indeed that we shouldn't return a process _name_ (trimmed for long ones even) as a fallback to a function that's expected to return full path to the process _executable_. This breaks API contract, could certainly break downstream programs.

> P.S. Maybe it is also a client responsibility to fall back to process name (that is properly detected) when getting executable returns error. In this case we can cancel this PR.

Yes indeed. I was confused by your proposed fix in #816 because to me you were conflating Exe() and Name() and I didn't know which function is used telegraf. Sorry that my intent wasn't clear.

As far as I know, there is no other way of getting a process full executable path using procfs than checking `/proc/PID/exe` symlink, so I will close this PR (and maybe the linked issue).

Issue #734, opened 2019-08-01
Implement better CI checks


Lomanic:
Linked to  #733

We currently use CircleCI for gopsutil CI. As far as I can see it doesn't run for our PRs, it looks like we should enable this option which is disabled by default
https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories

We should maybe also contact CircleCI to have macOS builders ([see here](https://circleci.com/open-source/)) for free. 

As I'm not even admin of the github repo I won't be able to do this, could you @shirou?

I'm also thinking about using [Gitlab CI](https://about.gitlab.com/solutions/github/) as I'm very familiar with it at work (as I administrate our own self-hosted Gitlab and runners), using their hosted instance at gitlab.com. We could even use our own runners on more platforms than what is proposed for free by CircleCI.

shirou:
sorry for not say that but yesterday I have enabled that setting.

![image](https://user-images.githubusercontent.com/177213/62295492-25037600-b4a8-11e9-94f0-6b5523cdb6a5.png)

or, if you are familiar with Gitlab CI, it is fine to migrate to it for me.

Lomanic:
Thanks for setting this. We will see next time a PR is created.

We could have both Gitlab and Circle CIs by the way. I'm leaving this issue open for the GitlabCI part.

Lomanic:
I confirm CircleCI ran tests in #735, excellent! :clap:.

For the GitlabCI part, I now remember I won't be able to add it if I'm not Owner of the repo unfortunately.

Lomanic:
I think these checks are sufficient for now, it already checks that it compiles on all platforms. I will probably implement a go-fmt check though and will close this afterwards.

Lomanic:
Test https://staticcheck.io/docs/ also maybe

Lomanic:
https://cirrus-ci.org/features/

> To support Open Source community Cirrus CI provides Linux, Windows, macOS and FreeBSD services free of charge.

To be tested.

I'm also curious about builds.sr.ht, which has [1st class support for various BSDs](https://sourcehut.org/blog/2019-09-12-sourcehut-makes-bsd-software-better/) https://man.sr.ht/tutorials/builds.sr.ht/github-integration.md

shirou:
I have tested Github actions which has Linux/Darwin/Windows support on #805 but no luck...

So I have installed cirrus-ci and start building #806 but not successes yet. I will continue on next time...

PR #821, opened 2020-01-17
Fix check pid existence when running in different process namespace (container)

i-prudnikov:
This fix is to address issue mentioned in https://github.com/shirou/gopsutil/pull/713, originally comes from https://github.com/influxdata/telegraf/issues/6813

### DIFF
--- a/process/process_posix.go
+++ b/process/process_posix.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/shirou/gopsutil/internal/common"
 	"golang.org/x/sys/unix"
 )
 
@@ -78,6 +79,16 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+
+	if _, err := os.Stat(common.HostProc()); err == nil { //Means that proc filesystem exist
+		// Checking PID existence based on existence of /<HOST_PROC>/proc/<PID> folder
+		// This covers the case when running inside container with a different process namespace (by default)
+		_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))
+		return err == nil, err
+	}
+
+	//'/proc' filesystem is not exist, checking of PID existence is done via signalling the process
+	//Make sense only if we run in the same process namespace
 	err = proc.Signal(syscall.Signal(0))
 	if err == nil {
 		return true, nil
@@ -95,6 +106,7 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
 	case syscall.EPERM:
 		return true, nil
 	}
+
 	return false, err
 }
 


shirou:
Indeed. It could break cgroup independency, but if a user does not want to allow a process can watch a process existance out side of a container, simply does not add environment variable. So I agree to this PR.

 so basically I agree to this great PR, but could you `go fmt` please?

process/process_posix.go
```diff
@@ -5,6 +5,7 @@ package process
 import (
 	"context"
 	"fmt"
+	"github.com/shirou/gopsutil/internal/common"
```
Lomanic: Please follow our convention to separate stdlib and 3rd-party packages in import statements.

process/process_posix.go
```diff
@@ -78,23 +79,35 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
-	case syscall.ESRCH:
-		return false, nil
-	case syscall.EPERM:
-		return true, nil
+
+	if ! strings.HasPrefix(common.HostProc(),"/proc") { //Means that proc is mounted not to /proc, and we highly
+		// likely running inside container with a different process namespace (by default), so we check pid existence
+		// based on /<HOST_PROC>/proc/<PID> folder
+		file, err := os.Open(common.HostProc(strconv.Itoa(int(pid))))
+		defer file.Close()
```
Lomanic: This will panic if file is `nil`, if we encounter an error. This should be after the check for `err != nil`.

process/process_posix.go
```diff
@@ -78,23 +79,35 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
+		if err!=nil{return false, err} else{return true, nil}
+	}else{ //Assume that we are in the same process namespace, try to signal process
```
Lomanic: Useless `else` as we always return in the `if`.

process/process_posix.go
```diff
@@ -78,23 +79,35 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
-		return false, nil
-	case syscall.EPERM:
-		return true, nil
+
+	if ! strings.HasPrefix(common.HostProc(),"/proc") { //Means that proc is mounted not to /proc, and we highly
+		// likely running inside container with a different process namespace (by default), so we check pid existence
+		// based on /<HOST_PROC>/proc/<PID> folder
+		file, err := os.Open(common.HostProc(strconv.Itoa(int(pid))))
+		defer file.Close()
+		if err!=nil{return false, err} else{return true, nil}
```
Lomanic: `return err == nil, err`

process/process_posix.go
```diff
@@ -78,23 +79,35 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
-	if !ok {
-		return false, err
-	}
-	switch errno {
-	case syscall.ESRCH:
-		return false, nil
-	case syscall.EPERM:
-		return true, nil
+
+	if ! strings.HasPrefix(common.HostProc(),"/proc") { //Means that proc is mounted not to /proc, and we highly
```
Lomanic: This code is called on freebsd openbsd darwin where there is no procfs, could lead to downstream breakage if users blindly set the `HOST_PROC` variable on every OS. Probably heck for `common.HostProc()` existence instead.

process/process_posix.go
```diff
@@ -78,23 +79,35 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
+		// likely running inside container with a different process namespace (by default), so we check pid existence
+		// based on /<HOST_PROC>/proc/<PID> folder
+		file, err := os.Open(common.HostProc(strconv.Itoa(int(pid))))
```
Lomanic: Use `os.Stat` instead as in this commit https://github.com/shirou/gopsutil/commit/258343806a4d3002c09ae9632715b5364df840e0

process/process_posix.go
```diff
@@ -78,23 +79,35 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
-		return false, err
-	}
-	switch errno {
-	case syscall.ESRCH:
-		return false, nil
-	case syscall.EPERM:
-		return true, nil
+
+	if ! strings.HasPrefix(common.HostProc(),"/proc") { //Means that proc is mounted not to /proc, and we highly
+		// likely running inside container with a different process namespace (by default), so we check pid existence
```
Lomanic: Likewise, all your comments assume we're running on Linux, which is confusing because this code is common to all POSIX platforms.

process/process_posix.go
i-prudnikov: Done, gofmt was surprisingly switched off :)

process/process_posix.go
i-prudnikov: This and below, restructured this part taking into consideration this and all of the comments below.

i-prudnikov:
Overall, I followed this logic:
1. If /<HOST_PROC>/proc or /proc exist, then PID existence is checked via existence of folder in proc filesystem
2. Otherwise PID existence is checked by signalling.

process/process_posix.go
```diff
@@ -80,32 +80,31 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
...
 	}
 
-	if ! strings.HasPrefix(common.HostProc(),"/proc") { //Means that proc is mounted not to /proc, and we highly
-		// likely running inside container with a different process namespace (by default), so we check pid existence
-		// based on /<HOST_PROC>/proc/<PID> folder
-		file, err := os.Open(common.HostProc(strconv.Itoa(int(pid))))
-		defer file.Close()
-		if err!=nil{return false, err} else{return true, nil}
-	}else{ //Assume that we are in the same process namespace, try to signal process
+	if _, err := os.Stat(common.HostProc()); err == nil { //Means that proc filesystem exist
```
i-prudnikov: Currently it is 2 calls to os.Stat, can be reduced to 1 if we cache somewhere in global var the result of this line.
@Lomanic, @shirou could you please advise the right place for that?

shirou: These `os.Stat` target are different and second one is a fallback. so we can not merge.

shirou:
Thank you @Lomanic and @i-prudnikov . LGTM

Lomanic:
@i-prudnikov please squash your commits as to preserve lines history. Or @shirou please enable this option for merges. Thanks.

shirou:
Oh, roger. I have enabled squash merge!

process/process_posix.go
i-prudnikov: Well, they are. But I feel that  my point is not clear. The os.Stat(common.HostProc()) is useless to call every time when we check pid existence, this is something that could be checked once at startup or init. Possible use case here is when you have an array of PIDs and you check in cycle if the PID still exist. In this case you have n=len(pidArray) -1 useless os.Stat calls.

i-prudnikov:
@shirou, @Lomanic  commits are squashed, thank you for review.

shirou:
LGTM now. Thank you for your contribution!

Lomanic:
This shouldn't have been merged as now the following test is failing:
```
--- FAIL: Test_IsRunning (1.00s)
    process_test.go:620: IsRunning error: open /proc/11200/stat: no such file or directory
```
A non-running process should not return an error.

We really need a better CI (#734) to catch this kind of regressions.

i-prudnikov:
@Lomanic , could you please provide the commit SHA from which you run the failed test. I'll try to investigate what goes wrong. I see that I have mismatched process_test.go file in my source branch.

i-prudnikov:
@Lomanic, I've run test against master branch,  the root case is the returned error from `PidExistsWithContext` function, that should be `nil`, in case process is not found. The fix is a simple as:
```
_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))
if e, ok := err.(*os.PathError); ok && e.Err.(syscall.Errno) == syscall.ENOENT {
	return err == nil, nil
}
return err == nil, err
```
I've opened a new merge request for that fix: https://github.com/shirou/gopsutil/pull/840

PR #826, opened 2020-01-26
[process][windows] Use win32 API in process.IOCounters() instead of slow WMI call #250

Lomanic:
#250


shirou:
Super great!! thank you so much!

PR #827, opened 2020-01-27
net: move netstat parsing functions to net.go

Helflym:
As AIX is also using "netstat -an" in order to get connections,
move netstat functions to net.go to make it available for both
OpenBSD and AIX.
Also rename lsof parsing functions to have Lsof in their name.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


shirou:
`netstat` output is different between OSes. so we can not move to the generic `net.go` file.

Helflym:
Some parts are similars, especially when dealing with net connections. 
But I understand your point of view, so I'll close this PR, and directly submit the AIX one with its own netstat parsing functions.

PR #829, opened 2020-01-31
mem: collect Active(anon), Inactive(anon) and Unevictable

xueweiz:
These fields helps people understand anonymous memory usage pattern.

We plan to use them in the [node-problem-detector](https://github.com/kubernetes/node-problem-detector/tree/master/pkg/systemstatsmonitor) project.


shirou:
Active(anon) and Inactive(anon) are exist only Linux. And `VirtualMemoryExStat` is for only linux(currently), so PR is acceptable. Thank you for your contribution!

PR #834, opened 2020-02-10
net: add AIX support

Helflym:


Helflym:
Any news about this CL ? I need it for other go packages.

Lomanic:
Sorry, I wanted to check it out but it was really hard to come by a way to get AIX on Virtualbox (for free). I finally stumbled upon [this promising blog article](https://astr0baby.wordpress.com/2018/09/14/running-aix-1-3-inside-virtual-box-5-2-16/) but didn't get the chance to test it out. With some exceptions I'm only able to work on gopsutil on the weekend, and even then, not all weekends. 

Also, I wasn't sure if it would have been a good idea to merge this if nobody in the team would be able to test and fix things for this OS in the future.

Helflym:
Yes, I understand your point of view. You're not the only Go package maintainers wondering how to check on AIX. Sadly, I don't have any solutions for you (AIX 1.3 won't work at all with Go, trust me !). 
The only solution right now is that AIX community has to provide patches when you'll be introducing new features/bugs. As your package is included in wider projects, at least beats software, we're likely to detect bug or missing features will compiling their new versions.

shirou:
Thank you for this PR. And as @Lomanic said, we can not check gopsutil is working on all platforms. So I agree to @Helflym, I trust that platform community. If gopsutil can not work on the platforms, someone will open an issue or PR like this PR. This is a OSS. Therefore, I merge this PR. Thank you for your contribution!


(Sorry for late response. I can work only weekend same as @Lomanic. I am very appreciate and respecting @Lomanic about his contribution)

PR #838, opened 2020-02-13
host: fix TestHostInfoStat_String

tklauser:
TestHostInfoStat_String has been broken since kernelArch has been
introduced. Fix it.

Also adjust the error message to show the wanted string, aligned with
the gotten string so it's easier to spot differences.


Lomanic:
Good catch thank you. The improved error message is also a nice addition.

PR #839, opened 2020-02-14
net: move parseNetLine and parseNetAddr to net_unix.go

tklauser:
These funcs are only used in net_unix.go which is only compiled for
darwin and freebsd.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -3,11 +3,7 @@ package net
 import (
 	"context"
 	"encoding/json"
-	"fmt"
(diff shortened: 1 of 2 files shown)


tklauser:
@shirou @Lomanic if you find time, could either one of you please have a look? Thanks!

shirou:
sorry for late reply. Yes, you are right. Thank you for your contribution!

tklauser:
Thanks @shirou!

PR #840, opened 2020-02-14
Correct returned error (set to nil) in case process is not exist

i-prudnikov:
To address test failure:
```
--- FAIL: Test_IsRunning (1.00s)
    process_test.go:620: IsRunning error: open /proc/11200/stat: no such file or directory
```
 mentioned in https://github.com/shirou/gopsutil/pull/821

### DIFF
--- a/process/process_posix.go
+++ b/process/process_posix.go
@@ -83,7 +83,11 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
 	if _, err := os.Stat(common.HostProc()); err == nil { //Means that proc filesystem exist
 		// Checking PID existence based on existence of /<HOST_PROC>/proc/<PID> folder
 		// This covers the case when running inside container with a different process namespace (by default)
+
 		_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))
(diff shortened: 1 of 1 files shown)


process/process_posix.go
```diff
@@ -83,7 +83,11 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
 	if _, err := os.Stat(common.HostProc()); err == nil { //Means that proc filesystem exist
 		// Checking PID existence based on existence of /<HOST_PROC>/proc/<PID> folder
 		// This covers the case when running inside container with a different process namespace (by default)
+
 		_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))
+		if e, ok := err.(*os.PathError); ok && e.Err.(syscall.Errno) == syscall.ENOENT {
```
shirou: To check the file is exists or not, how about to use `os.IsNotExist`? and if not exists, should always return false, no need to check `err == nil`.
```suggestion
		if os.IsNotExist(err) {
		    return false, nil
		}
```

i-prudnikov: Indeed, looks better, MR updated and commits are squashed

PR #843, opened 2020-02-20
add macos test on CircleCI.

shirou:
now we can use mac build with circleci by using OSS project

https://circleci.com/open-source/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
-version: 2
+version: 2.1
+
+
 jobs:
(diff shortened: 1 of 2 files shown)


shirou:
```
Orb cci-orb/golang@0.1.4 not loaded. To use this orb, an organization admin must opt-in to using third party orbs in Organization Security settings.
```
error happend even I enabled the opt-in to use [this orb](https://circleci.com/orbs/registry/orb/cci-orb/golang#executors-macos)

PR #845, opened 2020-02-21
Add memory.ItemizedPercent

yudai:
Hi, thank you for the handy library.

---
This new method returns a new struct PercentStat that holds CPU usage
percentages for different work loads (e.g. user, system and so on).
This commit also exposes CalculatePercent and CaluculateItemizedPercent
that returns usage percentages from two TimesStats so that users
can calculate percentages themselves as well.


shirou:
Thank you for your PR. But I think we should keep this library as simple as possible. And it seems this PR provide utility function, so how about to create another library which uses gopsutil?

yudai:
Thank you for the reply. I understand keeping a library simple is always the way to go.

However, I'm afraid I assume this is a common use case. Actually I just found `psutil` has the same functionality.

https://psutil.readthedocs.io/en/latest/#psutil.cpu_times_percent
https://github.com/giampaolo/psutil/issues/362

What if we simply rename `ItemizedPercent()` added in this PR to `CPUTimesPercent()` and make it return `TimesStat` instead of new `PercentStat` (IMO, it's better/safer to separate `TimeStat` and `PercentStat` as they store different types of values, but it's not critical to me)?

`CalculatePercent()` and `CaluculateItemizedPercent()` are just to expose existing logic in `cpu.go` for uses. We can drop these from the PR if you'd like.

If the idea above works for you, I'll update my code.

PR #850, opened 2020-03-03
[process][windows] Fix #846 use win32 API in disk.IOCounters() instead of slow WMI call returning wrong values

### LINKED ISSUE #846: IOCountersStat ReadCount & WriteCount are always 0 on Windows 10
**Describe the bug**
`IOCountersStat.ReadCount` and `IOCountersStat.WriteCount` are always 0 on Windows 10.
https://github.com/shirou/gopsutil/blob/master/disk/disk.go#L32
On Mac OS X and Linux(Ubuntu) this works fine. Do I do something wrong or is it simply a bug?

**To Reproduce**
```go
diskCounters, _ := disk.IOCounters()
fmt.Println(diskCounters)
```

```shell
map{C::{"readCount":0, ....., "writeCount":0,...
```

When I use Python `psutil` on the same machine from the same terminal I get all the info correctly:
```python
import psutil
print(psutil.disk_io_counters())
```

```shell
sdiskio(read_count=1385946, write_count=5768525.....)
```

**Expected behavior**
I expect disk.IOCounters() to return non-zero `readCount` and `writeCount` on Windows.

 - [x] Windows: [Version 10.0.18362.365]

**Additional context**
Also tried cross-compile on Mac using:
```shell
GOOS=windows GOARCH=amd64 go build main.go
```
as well as running go file straight inside the Windows machine:
```shell
go run main.go
```
Both with the same result.

Would appreciate any help. (I can rewrite my whole program to Python if need be though I'd preffer using Go)

Lomanic:
[process][windows] Fix #846 use win32 API in disk.IOCounters() instead of slow WMI call returning wrong values

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -5,6 +5,8 @@ package disk
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"syscall"
 	"unsafe"
 
 	"github.com/shirou/gopsutil/internal/common"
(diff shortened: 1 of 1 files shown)


Lomanic:
WIP as this is failing with GOARCH 386. Don't merge.

Lomanic:
Fixed on GOARCH 386.

Note that previously the keys of the map returned were letters like "C:", "D:" etc, we now return "PhysicalDrive0" etc. instead.

shirou:
sorry for late response. This PR is works for my Windows 10 and very good as usual. Thank you so much!

Lomanic:
No problem for the delay, it gave me time to adjust this PR.

PR #856, opened 2020-04-14
Provide an estimation of Iowait metric per process

juan-leon:
The trick to get a pretty accurate metric is using `delayacct_blkio_ticks` (see man proc)


shirou:
Thank you so much! yes, I confirm your PR is working and [psutil also have same code](https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L1578). Thank you for your great contribution!

PR #857, opened 2020-04-15
Do not cache boot time for linux. Fix #837

### LINKED ISSUE #837: Uptime can jump when date is adjusted for devices without RTC
**Describe the bug**

Related to https://github.com/influxdata/telegraf/issues/7018

When a device starts at a wrong date (say epoch, or another more recent date reported by fake_hwclock for instance), and the date is later adjusted to the real one with NTP, gopsutil starts reporting wrong uptime.

**To Reproduce**
```go
Uptime() // reports 30s uptime
/* 1 second elapses but NTP adjusts the date 24h in the future */
Uptime() // reports 1 day an 31s uptime
```

**Expected behavior**

Do not use a differential computation for linux devices, maybe prefer parsing the /proc/uptime ?
Or at least try to detect large time slippage and reset the boot time ?

 - Linux: PRETTY_NAME="Raspbian GNU/Linux 10 (buster) / Linux eanode5 4.19.75-v7+ #1270 SMP Tue Sep 24 18:45:11 BST 2019 armv7l GNU/Linux

**Additional context**
Device has no RTC (Raspberry Pi 3 A+)

Gui13:
This should fix the issue #837: we now query the value of /proc/stat or /proc/uptime at each invocation.

Not doing so will not return the correct boot time when NTP adjusts the time of the system in linux.

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -10,7 +10,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync/atomic"
 	"time"
 )
 
@@ -54,14 +53,7 @@ func NumProcs() (uint64, error) {
 	return cnt, nil
 }
 
-// cachedBootTime must be accessed via atomic.Load/StoreUint64
-var cachedBootTime uint64
-
 func BootTimeWithContext(ctx context.Context) (uint64, error) {
-	t := atomic.LoadUint64(&cachedBootTime)
-	if t != 0 {
-		return t, nil
-	}
 
 	system, role, err := Virtualization()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Lomanic:
Could you test this with and without your patch applied please?

1. stop the NTP service
2. Reboot so that time is wrong on the Raspberry Pi
3. Launch this program below, before ntpd kicks in
4. Launch ntpd, time jumps
5. Observe what's reported by this test program (is it still reported as running after time jumps?)

```go
package main

import (
	"fmt"
	"os"
	"time"

)

func main() {
	p, _ := process.NewProcess(int32(os.Getpid()))

	c := time.Tick(5 * time.Second)
	for range c {
		fmt.Println(time.Now(), p.IsRunning())
	}
}
```

Lomanic:
I was able to test my program above on my Rapsberry Pi and IsRunning() is returning false for the current process after time jumps, with or without this patch. I didn't test host.Uptime() though, but it's probably good. Will test it soon and merge if OK. Thanks for your patience.

Lomanic:
We should use [unix.Sysinfo](https://godoc.org/golang.org/x/sys/unix#Sysinfo) in the future for all Unix-like OSes, that would reduce platform-dependent code.

PR #869, opened 2020-05-12
Added check for `CLK_TCK` for darwin

renaynay:
This PR adds a check for the clock tick of darwin OS as is done for freebsd, linux, and openbsd when calculating CPU stats instead of assuming a static `ClocksPerSec` of 128.

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -4,6 +4,7 @@ package cpu
 
 import (
 	"context"
+	"os/exec"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


Lomanic:
Shelling out to an external process is to be avoided as it's expensive (relatively slow) and relatively unreliable (output could change over time or between different hosts).

In this case and reading code from https://github.com/tklauser/go-sysconf and https://stackoverflow.com/a/10455939, CLK_TCK (same thing as CLOCKS_PER_SEC, at least on Darwin) is defined at build-time and from POSIX to a constant, so there is no need to call getconf at runtime, just define it as a constant, from C headers. Or we should just use go-sysconf.

gballet:
@Lomanic that's a fair point, however what if the machine that compiles the code has a different value than the machine it runs on?

The default value was 128 which is different from the 100 we notice on the machine we tried it with. Presumably, the person who set this value noticed something different on their machines.

Lomanic:
I think the 128 constant is a mistake and should be set to 100 as per my previous comment, not sure this value can change on Darwin (it's defined by POSIX and Darwin is even [certified to be compliant](https://stackoverflow.com/a/33839387)). Either way, it's better to obtain the proper value now.

PR #870, opened 2020-05-12
Renamed variables storing `CLK_TCK` value for consistency across OSs

renaynay:
This PR renames the variables holding the value of `CLK_TCK` across all OSs for consistency. `ClocksPerSec` is used for darwin, openbsd, freebsd and solaris, but other names are used for linux and windows. It would be helpful to rename the variable so that it is consistent across all supported platforms.

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -13,7 +13,7 @@ import (
 	"github.com/shirou/gopsutil/internal/common"
 )
 
-var CPUTick = float64(100)
+var ClocksPerSec = float64(100)
 
 func init() {
 	getconf, err := exec.LookPath("getconf")
@@ -25,7 +25,7 @@ func init() {
 	if err == nil {
 		i, err := strconv.ParseFloat(strings.TrimSpace(string(out)), 64)
 		if err == nil {
-			CPUTick = i
+			ClocksPerSec = i
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


Lomanic:
These variables shouldn't have been exported in the first place, they are only needed internally by the package for stats calculation. Do you need them? I read in https://github.com/ethereum/go-ethereum/pull/21041 that our docs may be wrong, optimally we should align on [psutil's](https://psutil.readthedocs.io/en/latest/#psutil.cpu_times) and return the same values, if that's not the case I would prefer fixing this this way (fix docs and returned values).

renaynay:
@Lomanic I've created a separate [PR](https://github.com/shirou/gopsutil/pull/872) to update the docs for `TimesStat`. Hopefully this mitigates some of the misunderstanding. 

The current implementation of TimesStat in gopsutil is aligned with psutil's, as the documentation for psutil points out that every attribute (of `cpu_times`) represents the seconds the CPU has spent in the given mode.

Lomanic:
Thanks for your quick reply, it clears things up and thanks for opening this PR to improve the docs. Still remains the fact that these variables shouldn't be exported to end-users because it's only a lib internal, do you have usage for these variables?

renaynay:
Yes, we still have a usage for these variables as we'd like to continue returning the cpu ticks, so it would be nice to be able to pull the ClocksPerSec value from this lib. We feel it makes sense to make it part of the public psutil interface, since it is a system-wide parameter that we would otherwise need to get through other means.

renaynay:
Hey @Lomanic , do you have any other comment on this? Just following up and checking in on the status.

Issue #620, opened 2018-12-19
process.MemoryInfo returns MemoryInfoStat with only two out of five fields filled on linux


tonimctoni:
I do not know if this is intended, but on linux *process.MemoryInfo* returns a *MemoryInfoStat* struct with only the *RSS* and *VMS* fields set.

They get set [here](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L942).

I find it weird because [here](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L981) the other fields get set too, so it does not seem to be a limitation of the platform.

It got me confused because the procstat plugin for telegraf seems to ignore this [here](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L217).

shirou:
Sorry for late response. Could you show us your platform and if linux, `/proc/<some pid>/status` ?

vazhem:
https://github.com/influxdata/telegraf/issues/3606
Telegraf uses MemoryInfo() to get memory info for a process.
The problem is that gopsutil for this call returns data from /proc/pid/statm.
/proc/pid/statm does not have info about swap. But /proc/pid/status does.
So gopsutil should be fixed to return swap info in the MemoryInfo() call.

example of /proc/pid/status

```
cat /proc/80425/status
Name:   telegraf
Umask:  0022
State:  S (sleeping)
Tgid:   80425
Ngid:   80425
Pid:    80425
PPid:   80150
TracerPid:      0
Uid:    0       0       0       0
Gid:    0       0       0       0
FDSize: 256
Groups:
NStgid: 80425   136
NSpid:  80425   136
NSpgid: 80150   1
NSsid:  80150   1
VmPeak:  3269424 kB
VmSize:  3220792 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:     40704 kB
VmRSS:     32588 kB
RssAnon:           32588 kB
RssFile:               0 kB
RssShmem:              0 kB
VmData:   728780 kB
VmStk:       132 kB
VmExe:     22988 kB
VmLib:      2080 kB
VmPTE:       696 kB
VmPMD:        28 kB
VmSwap:     2800 kB
HugetlbPages:          0 kB
Threads:        76
SigQ:   0/251420
SigPnd: 0000000000000000
ShdPnd: 0000000000000000
SigBlk: 0000000000000000
SigIgn: 0000000000000000
SigCgt: ffffffffffc1feff
[...]

PR #871, opened 2020-05-12
Add more info for MemoryInfoWithContext on linux

Euqaidoz:
This should solve the following issue:
https://github.com/shirou/gopsutil/issues/620

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -84,7 +84,7 @@ func (p *Process) Name() (string, error) {
 
 func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 	if p.name == "" {
-		if err := p.fillFromStatusWithContext(ctx); err != nil {
+		if _, err := p.fillFromStatusWithContext(ctx); err != nil {
 			return "", err
 		}
 	}
@@ -94,7 +94,7 @@ func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 // Tgid returns tgid, a Linux-synonym for user-space Pid
 func (p *Process) Tgid() (int32, error) {
 	if p.tgid == 0 {
-		if err := p.fillFromStatusWithContext(context.Background()); err != nil {
+		if _, err := p.fillFromStatusWithContext(context.Background()); err != nil {
 			return 0, err
 		}
 	}
@@ -153,7 +153,7 @@ func (p *Process) Parent() (*Process, error) {
 }
 
 func (p *Process) ParentWithContext(ctx context.Context) (*Process, error) {
-	err := p.fillFromStatusWithContext(ctx)
+	_, err := p.fillFromStatusWithContext(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -173,7 +173,7 @@ func (p *Process) Status() (string, error) {
 }
 
 func (p *Process) StatusWithContext(ctx context.Context) (string, error) {
-	err := p.fillFromStatusWithContext(ctx)
+	_, err := p.fillFromStatusWithContext(ctx)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


Lomanic:
I'm not sure what you're trying to achieve here, p.memInfo is being populated in fillFromStatusWithContext currently, so couldn't your change simply be limited to only redefining MemoryInfoWithContext like this?

```
func (p *Process) MemoryInfoWithContext(ctx context.Context) (*MemoryInfoStat, error) {
	_, err := p.fillFromStatusWithContext(ctx)
	if err != nil {
		return nil, err	}
	}
	return p.memInfo, nil
}
```
Note also that process_linux.go is not go-fmt-ed.

Euqaidoz:
Hi, thanks for the feedback, you are absolutely right, I will redo the changes

Euqaidoz:
Hi, could someone merge this, or tell me what I should do to have it merged?
Thanks!

reimda:
@shirou I think this will fix an issue we have in telegraf with inaccurate swap measurements. Could you take a look and let us know what you think? Thanks!

TiagoMHCSantana:
@shirou The fix that @reimda mentioned is really important. Could you please take a look at this?

shirou:
Sorry, I totally forget about this issue.

From my understanding, there are two functions.
- `fillFromStatusWithContext` to get memory information from `status`.
- `fillFromStatmWithContext` to get memory information from `statm`.

`fillFromStatmWithContext` returns `Meminfo` in addition to MeminfoEx, which is used by `MemoryInfoWithContext`. But the `Meminfo` from `fillFromStatmWithContext` was no Swap, etc.

And after this PR, the `meminfo` return value in `fillFromStatmWithContext` is not required anymore. I will update this PR.

shirou:
@TiagoMHCSantana I have updated this PR (thank you for @Euqaidoz as an original author). I think memory info from `status` and `statm` should be equal, but could you double-check about the update? 

Thank you!

TiagoMHCSantana:
Thank you @shirou. I am going to double-check it indirectly via telegraf and then get back to you!

CappiJonas:
Hello @shirou, how are you?
I work with @TiagoMHCSantana and he asked me to test this change, but I'm having a little difficulty of doing that. I tried to reference the dependency by using de go get command with the git hash of the commit that you did and with the name of the branch, but I didn't have success. I will send the commands that I ran to you take a look. Could you help me with that showing me if the commands are right and, if not, what the right commands. 
Thank you very much in advance.

Commands:
- go get -u github.com/shirou/gopsutil/v3@d500d42a346d175f0f22a883dfccca3645edae3b
- go get -u github.com/shirou/gopsutil@Euqaidoz:improve-linux-process-MemoryInfo

shirou:
```
git clone https://github.com/shirou/gopsutil.git
cd gopsutil
git checkout d500d42a346d175f0f22a883dfccca3645edae3b

cd process
vi process_test.go 
(add `t.Log(v)` after `p.MemoryInfo()` in `func Test_Process_MemoryInfo`)

go test -run Test_Process_MemoryInfo -v
```

If you get a result like this that has a `hwm` and `data` value, it is what we expected.

```
=== RUN   Test_Process_MemoryInfo
    process_test.go:124: {"rss":5922816,"vms":1638748160,"hwm":5922816,"data":84295680,"stack":135168,"locked":0,"swap":0}
```

CappiJonas:
Hello shirou. How are you?
So, thank you for your answer. I did the test like you guided me, but the information that we need is the swap memory. I stress the memory of one of the our vm's and I ran the this test and the result is showed below:
image
As you can see, the used swap memory is 1.5GB and the test returned zero. So, how could we resolve this problem of returning zero in the swap memory?

shirou:
That you are using the `free` command, Perhaps what you want to get is not the swap of one single process, but the swap of the whole system? If so, please try `mem.SwapMemory()` ([godoc](https://pkg.go.dev/github.com/shirou/gopsutil/v3@v3.24.4/mem#SwapMemory)).

CappiJonas:
Actually, we need the swap of the proccess. Do you know how I can test the return of the swap usage?

PR #876, opened 2020-05-17
Don't ignore context for io counters

gtosh4:
`IOCountersWithContext` would drop the given `ctx` and call `IOCountersByFile` which would create a background context then call `IOCountersByFileWithContext`.


Lomanic:
Thanks, good catch

PR #878, opened 2020-05-19
[load][android] implement on top of sysinfo syscall

### LINKED ISSUE #877: load.Avg() doesn't work on Android
Android is no longer allowing access to /proc:

```
$ cat /proc/loadavg
cat: /proc/loadavg: Permission denied
```

However busybox uptime still shows loadavg on Android, and it seems to use `sys/sysinfo.h`

```
$ uptime
 22:29:23 up  2:28,  load average: 2.38, 2.42, 2.40
```

https://github.com/mirror/busybox/blob/b097a84d62dcfcf201a695b54bcfafae70cac8a6/procps/uptime.c#L67

tmm1:
fixes #877

### DIFF
--- a/load/load_linux.go
+++ b/load/load_linux.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/shirou/gopsutil/internal/common"
 )
@@ -16,6 +17,29 @@ func Avg() (*AvgStat, error) {
 }
 
 func AvgWithContext(ctx context.Context) (*AvgStat, error) {
+	stat, err := fileAvgWithContext(ctx)
+	if err != nil {
+		stat, err = sysinfoAvgWithContext(ctx)
+	}
+	return stat, err
+}
+
+func sysinfoAvgWithContext(ctx context.Context) (*AvgStat, error) {
+	var info syscall.Sysinfo_t
+	err := syscall.Sysinfo(&info)
+	if err != nil {
+		return nil, err
+	}
+
+	const si_load_shift = 16
+	return &AvgStat{
+		Load1:  float64(info.Loads[0]) / float64(1<<si_load_shift),
+		Load5:  float64(info.Loads[1]) / float64(1<<si_load_shift),
+		Load15: float64(info.Loads[2]) / float64(1<<si_load_shift),
+	}, nil
+}
+
+func fileAvgWithContext(ctx context.Context) (*AvgStat, error) {
 	values, err := readLoadAvgFromFile()
 	if err != nil {
 		return nil, err


load/load_linux.go
```diff
@@ -16,6 +18,29 @@ func Avg() (*AvgStat, error) {
 }
 
 func AvgWithContext(ctx context.Context) (*AvgStat, error) {
+	if runtime.GOOS == "android" {
```
Lomanic, line 21: You should just redefine the whole function body to your version I think, I don't see any point in keeping the old definition and doing a special case just for Android. It will be probably more optimized also.

tmm1: I found conflicting information online about how to translate the values into load floats. I think maybe it has changed over time, so it seems safer to only enable this for android right now (where load is broken altogether).

tmm1: The version in this PR I tested on modern linux kernels and confirmed the values are correct. I think it won't work on older kernels (whereas android always keeps relatively up to date kernels).

Lomanic: Do you have specific kernel versions where it was said to be different? Go requires [2.6.23](https://github.com/golang/go/wiki/MinimumRequirements#linux), so maybe that's then always been calculated the same way starting from before this version.

tmm1: No I wasn't able to find the history. I only saw different stackoverflow solutions with different  advice.

I ended up doing the same thing as https://stackoverflow.com/a/4993273/332798 because that produced the correct results on kernels 4.9.113 and 5.4.15

But in [busybox](https://github.com/mirror/busybox/blob/b097a84d62dcfcf201a695b54bcfafae70cac8a6/procps/uptime.c#L67) and [here](https://stackoverflow.com/a/12303849/332798) they use `FSHIFT` also. When I tried that, the numbers produced were not correct.

load/load_linux.go
```diff
@@ -16,6 +18,29 @@ func Avg() (*AvgStat, error) {
...
+		return sysinfoAvgWithContext(ctx)
+	} else {
+		return fileAvgWithContext(ctx)
+	}
+}
+
+func sysinfoAvgWithContext(ctx context.Context) (*AvgStat, error) {
+	var info syscall.Sysinfo_t
+	err := syscall.Sysinfo(&info)
```
tmm1, line 30: I used `syscall` package here but the same method and struct are also in the `unix` package. Not sure which one is better to use.. The implementation looks the same so I guess it doesn't matter.

load/load_linux.go
Lomanic: I'm not a fan about switching against runtime.GOOS, would prefer to call fileAvgWithContext, then sysinfoAvgWithContext in case of error returned by fileAvgWithContext.

tmm1: I don't think it is a good idea because it will break load average values on some systems without any error shown. But if you prefer I can change it.

tmm1: I misunderstood.. call fileAvg first. Okay, good idea!

tmm1: Done!

Lomanic:
Was curious about benchmark

```go
package main

import (
        "fmt"
        "os"
        "time"

        "github.com/shirou/gopsutil/load"
)

func main() {
        start := time.Now()
        fmt.Println(load.Avg())
        elapsed := time.Now().Sub(start)
        fmt.Println("Elapsed time", elapsed)

        os.Setenv("HOST_PROC", "/nonexisting")
        start = time.Now()
        fmt.Println(load.Avg())
        elapsed = time.Now().Sub(start)
        fmt.Println("Elapsed time", elapsed)
}
```

Result:
```
{"load1":2.07,"load5":1.84,"load15":1.76} <nil>
Elapsed time 138.252µs
{"load1":2.072265625,"load5":1.84228515625,"load15":1.7587890625} <nil>
Elapsed time 13.316µs
```

PR #879, opened 2020-05-20
Fix smaps parser

guillemj:
The smaps parser has multiple problems:

1. It is checking the length of the blocks, but we are allocating initialized entries so `len(blocks)` will never be 0, so we end up calling `getBlocks` with empty lines. Instead use the capacity form of `make()`.
2. It is passing the name of the next memory region to the previously parsed field blocks. Instead keep the previous `blockname` and pass that to `getBlocks`.
3. It is not handling any other field separator than spaces (new kernel versions include such fields). Instead use `strings.Fields()`, taking care not to pass to it empty strings.

And handle the new `THPeligible` smaps field, which does not contain a memory value and is also separated by both spaces and tabs, which should now be handled by the fixed parser.


shirou:
Sorry for late response and thank you for your PR.

But I have a question about 3. `strings.Fields()` can handle empty string ("") which returns empty slice and[ handle tab `\t` separated string.](https://play.golang.org/p/gDXc_EPNUGe).  From #890, `THPeligible` have tab separated line but I think current `strings.Fields()` handle it. So I think those change is not required, What do you think?

guillemj:
About the old 3) right that's what I would expect, I cannot recall why I did this, but from what I recall from testing it was misbehaving in some way. I've updated it to avoid that now.

PR #882, opened 2020-05-25
Add pgmajfault to data returned by vmstat on Linux

james-bebbington:
Minor change to add `pgmajfault` to swap data returned by `vmstat` on Linux

### DIFF
--- a/mem/mem.go
+++ b/mem/mem.go
@@ -88,6 +88,9 @@ type SwapMemoryStat struct {
 	PgIn        uint64  `json:"pgin"`
 	PgOut       uint64  `json:"pgout"`
 	PgFault     uint64  `json:"pgfault"`
+
+	// Linux specific numbers
+	// https://www.kernel.org/doc/Documentation/cgroup-v2.txt
 	PgMajFault  uint64  `json:"pgmajfault"`
 }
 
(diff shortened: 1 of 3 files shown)


shirou:
sorry for late response. Maybe I'm wrong, but I think `pgmajfault` is only in a Linux. if so, please add a comment ` // Linux specific numbers` like `VirtualMemoryStat`.

mem/mem.go
```diff
@@ -88,6 +88,7 @@ type SwapMemoryStat struct {
 	PgIn        uint64  `json:"pgin"`
 	PgOut       uint64  `json:"pgout"`
 	PgFault     uint64  `json:"pgfault"`
+	PgMajFault  uint64  `json:"pgmajfault"`
```
james-bebbington, line 91:
```suggestion
	// Linux specific numbers
	// https://www.kernel.org/doc/Documentation/cgroup-v2.txt
	PgMajFault  uint64  `json:"pgmajfault"`
```

james-bebbington:
> sorry for late response. Maybe I'm wrong, but I think `pgmajfault` is only in a Linux. if so, please add a comment ` // Linux specific numbers` like `VirtualMemoryStat`.

No worries - done

shirou:
Thank you so much!

PR #885, opened 2020-05-28
Move C files to be included by `go mod`

### LINKED ISSUE #832: macOS: Unable to build using Go Modules
**Describe the bug**
- https://github.com/shirou/gopsutil/pull/800
- This PR added `host/include/smc.h` and `host/include/smc.c`.  These files are in the `include` sub-directory.  This directory includes no go files, so Go modules does not include them when `go mod vendor` is used.  Either the h & c files should be moved into the host directory, or a dummy go file should be added to `include`

**To Reproduce**
Go mod vendor gopsutil, then try to build:
```
vendor/github.com/shirou/gopsutil/host/host_darwin_cgo.go:9:11: fatal error: 'include/smc.c' file not found
 #include "include/smc.c"
          ^~~~~~~~~~~~~~~
1 error generated.
```

 - [x] Mac OS: [paste the result of `sw_vers` and `uname -a`

afontaine:
Fixes #832 by moving the C files into a go module directory, which means they should be included by `go mod vendor`.

Not sure how I should name them, let me know if you want it changed.

### DIFF
--- a/host/host_darwin_cgo.go
+++ b/host/host_darwin_cgo.go
@@ -6,7 +6,7 @@ package host
 // #cgo LDFLAGS: -framework IOKit
 // #include <stdio.h>
 // #include <string.h>
-// #include "include/smc.c"
+// #include "smc_darwin.h"
 import "C"
 import "context"
 
--- a/host/smc_darwin.c
+++ b/host/smc_darwin.c
@@ -1,4 +1,4 @@
-#include "smc.h"
+#include "smc_darwin.h"
 
 #define IOSERVICE_SMC "AppleSMC"
 #define IOSERVICE_MODEL "IOPlatformExpertDevice"
--- a/host/smc_darwin.h
+++ b/host/smc_darwin.h
@@ -1,5 +1,5 @@
-#ifndef __SMC_H__
-#define __SMC_H__ 1
+#ifndef __SMC_DARWIN_H__
+#define __SMC_DARWIN_H__ 1
 
 #include <IOKit/IOKitLib.h>
 
@@ -29,4 +29,4 @@ kern_return_t open_smc(void);
 kern_return_t close_smc(void);
 double get_temperature(char *);
 
-#endif // __SMC_H__
+#endif // __SMC_DARWIN_H__


afontaine:
well, so much for this being easy 😞 

I don't really understand the error here either, do you have any insight @shirou?

afontaine:
oh hey that did the trick. I am a touch concerned that this would somehow still not work though 🤔 but my C seems to think that this should work

shoenig:
I believe this breaks compatibility with non-darwin targets

```bash
$ uname -a
Linux nzxt 5.4.0-33-generic #37-Ubuntu SMP Thu May 21 12:53:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

```bash
$ cat go.mod
module foo

go 1.14

replace github.com/shirou/gopsutil => github.com/afontaine/gopsutil v2.20.5-0.20200528220053-51e143da77b3+incompatible

require (
	github.com/shirou/gopsutil v0.0.0-00010101000000-000000000000
	golang.org/x/sys v0.0.0-20200523222454-059865788121 // indirect
)
```

```bash
$ cat foo.go
package main

import _ "github.com/shirou/gopsutil/host"
```

```bash
$ go build
foo.go:3:8: C source files not allowed when not using cgo or SWIG: smc.c

$ CGO_ENABLED=1 go build
foo.go:3:8: C source files not allowed when not using cgo or SWIG: smc.c
```

afontaine:
Ah... yes, of _course_ it's not that simple :thinking:

Lomanic:
You have to rename `smc.c` to `smc_darwin.c` and `smc.h` to `smc_darwin.h` and change the references in the source and it's then possible to `go build github.com/shirou/gopsutil/host`on Linux successfully. This is what I've done in https://github.com/shirou/gopsutil/issues/832#issuecomment-627818332. Though I prefer your names (smc_darwin.{c,h} compared to my less descriptive host_darwin.{c,h}).

Though there were the same compilation warnings with these changes, I fixed them in my own branch  (I'm quite sure I wouldn't be able to push to your branch) and took the occasion to rename darwin c files in the disk package.

As you spent some time on it and want some attribution maybe, I can let you fix that in your branch and merge this PR afterwards, I will rebase my PR after merging yours if you do so.

afontaine:
Ah the magic of go! Thanks @Lomanic 

I'm not super concerned about the attribution, more about the fix :grin: I updated the file names, and I don't really know what I'm doing with `go` :sweat_smile:

afontaine:
@Lomanic by all means please merge #889 instead of this, as it looks like you've captured more files that need to be moved/renamed 👍

Lomanic:
OK, no problem.

PR #890, opened 2020-06-15
WIP: start adding NewProcessWithFields

PierreF:
WIP on #286. I've created it to get feed-back and be sure the path I've took match @Lomanic expectation.
I'll continue in this way unless I got request to change the design.

The idea is to add a cache which will be around most inner function, the function that actually parse file. All those function will first check the cache and return value from cache is available.

This solution does not modify too much the exiting code, it mostly just add wrapper around inner function to read/write from the cache if needed.

I've already done change for MemoryMaps() and Ppid().

I've created a benchmark, to see the gain when calling 3 method (CreateTime(), Ppid() and Times()):
```
goos: linux
goarch: amd64
Benchmark_NewProcessWithFields/NewProcess-8                  876           1327201 ns/op
Benchmark_NewProcessWithFields/WithFields-8                 2434            439737 ns/op
PASS
ok      github.com/shirou/gopsutil/process      4.470s
```

Few remark on the current state:
* If we call a method for a field not requested at `NewProcessWithFields` creation, we got an `ErrorFieldNotRequested`. Currently if another Field parsed the file required, it won't return the error, but I will fix that (e.g. you may call `Ppid()`, `Times()` and `CreateTime()` even if you only request `CreateTime` because on linux all the three fields read `/proc/PID/stat`)
* There is a quick fix for #879, because Linux kernel from Ubuntu 20.04 added `THPeligible` in smaps which use tab instead of space :( I only need to change to run test on my laptop and will remove it from final PR.
* I'm not sure the benchmark worse being kept. Unless you ask me to keep it, I will remove it from final PR.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -17,6 +17,52 @@ var (
 	invoke                 common.Invoker = common.Invoke{}
 	ErrorNoChildren                       = errors.New("process does not have children")
 	ErrorProcessNotRunning                = errors.New("process does not exist")
+	ErrorFieldNotRequested                = errors.New("this fields was not requested")
+	ErrorRequireUpdate                    = errors.New("this method require updates and cannot be used with NewProcessWithFields")
+)
+
+type Field uint
+
+const (
+	AllFields Field = iota + 1
+	Background
+	Cmdline
+	CmdlineSlice
+	Connections
+	CPUPercent
+	CreateTime
+	Cwd
+	Exe
+	Foreground
+	Gids
+	IOCounters
+	IOnice
+	IsRunning
+	MemoryInfo
+	MemoryInfoEx
+	MemoryMaps
+	MemoryMapsGrouped
+	MemoryPercent
+	Name
+	NetIOCounters
+	NetIOCountersPerNic
+	Nice
+	NumCtxSwitches
+	NumFDs
+	NumThreads
+	OpenFiles
+	PageFaults
+	Parent
+	Ppid
+	Rlimit
+	RlimitUsage
+	Status
+	Terminal
+	Tgid
+	Threads
+	Times
+	Uids
+	Username
 )
 
 type Process struct {
@@ -32,10 +78,18 @@ type Process struct {
 	sigInfo        *SignalInfoStat
 	createTime     int64
 
+	cache             map[string]interface{}
+	prefetchCompleted bool
+
 	lastCPUTimes *cpu.TimesStat
 	lastCPUTime  time.Time
 
 	tgid int32
+}
+
+type valueOrError struct {
+	value interface{}
+	err   error
 }
 
 type OpenFilesStat struct {
@@ -167,6 +221,31 @@ func NewProcess(pid int32) (*Process, error) {
 	return p, nil
 }
 
+// NewProcessWithFields creates a new Process instance, and pre-fetch all requested
+// fields. Those fields and only those fields will be available and won't be updated
+// after creation.
+// This method may be faster if you retrive multiple fields, because unlike NewProcess where
+// files are parsed for each method call, here files are parsed once at creation.
+func NewProcessWithFields(pid int32, fields ...Field) (*Process, error) {
+	p := &Process{
+		Pid:   pid,
+		cache: make(map[string]interface{}),
+	}
+
+	exists, err := PidExists(pid)
+	if err != nil {
+		return p, err
+	}
+	if !exists {
+		return p, ErrorProcessNotRunning
+	}
+
+	p.prefetchFields(fields)
+	p.prefetchCompleted = true
+
+	return p, nil
+}
+
 func PidExists(pid int32) (bool, error) {
 	return PidExistsWithContext(context.Background(), pid)
 }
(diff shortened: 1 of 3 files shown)


PierreF:
Continued works. Linux should be completed.

Children field have the same issue as Parent field: if we prefetch this field and create children's Process with same pre-fetched fields, it may retrieve *all* processes.
I'm not sure what we want here, so currently both Parent and Children is not pre-fetched, and when called, the PIDs will be fetched and parent/children Process are created using same pre-fetched fields.

I've use a [basic "ps"](https://gist.github.com/PierreF/a3d93944890cfa522fd6eb46dcc65b57) that use gopsutils to validate the this PR doesn't impact performance of NewProcess. Performance is unchanged:
```
$ \time ./ps.master
10.01user 13.03system 0:18.44elapsed 124%CPU (0avgtext+0avgdata 13636maxresident)k
0inputs+88outputs (0major+7431minor)pagefaults 0swaps

$ \time ./ps.new
9.63user 12.79system 0:17.79elapsed 126%CPU (0avgtext+0avgdata 13584maxresident)k
0inputs+96outputs (0major+7419minor)pagefaults 0swaps
```

Note: this is when not using NewProcessWithFields, i.e. existing user when pay any performance penalty for this PR.

PierreF:
Add support for Windows and run test on Windows 10.

Performance are still not impacted for NewProcess (still using [basic "ps"](https://gist.github.com/PierreF/a3d93944890cfa522fd6eb46dcc65b57)):
```
C:\Users\IEUser\Downloads>ps-master.exe -count 100
errCount = 4300. Spent 435.0854ms

C:\Users\IEUser\Downloads>ps-new.exe -count 100
errCount = 4300. Spent 433.0671ms
```

(the errCount is due to trying to call not implemented function, like `Terminal` or `Status`)

PierreF:
Added ProcessesWithFields on Linux & Windows.

I've make ProcessesWithFields to take a context as argument. I think both ProcessesWithFields and NewProcessWithFields should both take a context or none. But I'm unsure which way we are moving. Are we adding context to all function (and ThingWithContext is the way to go) or are we removing context from all function.

I've updated my ["basic" ps](https://gist.github.com/PierreF/9c87f49001fbbdccc6dc52e1b824d612) to support WithFields. It *much* better on Linux. Windows is not that better:
```
$ ./ps 
errCount = 0. Spent 9.417733979s
$ ./ps -fields
errCount = 0. Spent 2.993156593s


```

PierreF:
This PR should be completed now. Waiting for review.

AtakanColak:
Thank you for this PR, it seems like it would be very useful to my for my project in which procs take most of the compute time. Could we cache [internal/common.Virtualization()](https://github.com/shirou/gopsutil/blob/master/internal/common/common_linux.go#L113) function after first ever call to it as well? It seems that there isn't much reason to call it again as the OS and Virtualization information won't change.

I will reduce number of NewProcess calls in my code but here is the profiling tree
image

Lomanic:
@AtakanColak caching common.Virtualization() may be for another PR/issue.

@PierreF sorry dropped the ball on this, life got in the way and then forgot about this specific PR. That's quite a big change but also an improvement as highlighted also by @AtakanColak. I don't know when I will be able to review it (I had a smaller change in mind, but yours is more optimized probably).

AtakanColak:
@PierreF I also would like to make a suggestion regarding this PR. While calling NewProcessWithFields() we are spending some time on fetching requested information from the OS. However in that meantime, the process can exit, and even its PID can get replaced by a newer process. For this, I believe there should be a simple CreateTime() check at the end of the NewProcessWithFields() function. 


`start := time.now() // before fetching `
`// fetch requested fields`
`crt, err := p.CreateTime() // after fetching`
`if err != nil  // if we can't get create time at this point maybe we should return this error`
`if crt > (start.UnixNano() / 1000000) // because CreateTime() return in milliseconds and UnixNano is nanoseconds`
`// if crt is after we started, then return an error saying process was changed in the meantime`

Purushothama-VS:
when will be this pull request getting merged?

Purushothama-VS:
> Continued works. Linux should be completed.
> 
> Children field have the same issue as Parent field: if we prefetch this field and create children's Process with same pre-fetched fields, it may retrieve _all_ processes. I'm not sure what we want here, so currently both Parent and Children is not pre-fetched, and when called, the PIDs will be fetched and parent/children Process are created using same pre-fetched fields.
> 
> I've use a [basic "ps"](https://gist.github.com/PierreF/a3d93944890cfa522fd6eb46dcc65b57) that use gopsutils to validate the this PR doesn't impact performance of NewProcess. Performance is unchanged:
> 
> ```
> $ \time ./ps.master
> 10.01user 13.03system 0:18.44elapsed 124%CPU (0avgtext+0avgdata 13636maxresident)k
> 0inputs+88outputs (0major+7431minor)pagefaults 0swaps
> 
> $ \time ./ps.new
> 9.63user 12.79system 0:17.79elapsed 126%CPU (0avgtext+0avgdata 13584maxresident)k
> 0inputs+96outputs (0major+7419minor)pagefaults 0swaps
> ```
> 
> Note: this is when not using NewProcessWithFields, i.e. existing user when pay any performance penalty for this PR.

when do you plan to merge this pull request?

PR #893, opened 2020-06-18
cpu: support for DragonflyBSD

gballet:


cpu/cpu_dragonfly.go
```diff
@@ -0,0 +1,173 @@
...
+	"golang.org/x/sys/unix"
+)
+
+var ClocksPerSec = float64(128)
+var cpuMatch = regexp.MustCompile(`^CPU:`)
+var originMatch = regexp.MustCompile(`Origin\s*=\s*"(.+)"\s+Id\s*=\s*(.+)\s+Family\s*=\s*(.+)\s+Model\s*=\s*(.+)\s+Stepping\s*=\s*(.+)`)
+var featuresMatch = regexp.MustCompile(`Features=.+<(.+)>`)
+var featuresMatch2 = regexp.MustCompile(`Features2=[a-f\dx]+<(.+)>`)
+var cpuEnd = regexp.MustCompile(`^Trying to mount root`)
+var cpuCores = regexp.MustCompile(`FreeBSD/SMP: (\d*) package\(s\) x (\d*) core\(s\)`)
```
shirou, line 24: sorry, I don't have a Dragonfly, so I can not check but is this `FreeBSD` correct ?

cpu/cpu_dragonfly.go
```diff
@@ -0,0 +1,173 @@
...
+	buf, err := unix.SysctlRaw("kern.cp_time")
+	if err != nil {
+		return nil, err
+	}
+
+	times := (*cpuTimes)(unsafe.Pointer(&buf[0]))
+	return []TimesStat{*timeStat("cpu-total", times)}, nil
+}
+
+// Returns only one InfoStat on FreeBSD.  The information regarding core
```
shirou, line 92: Could you change this comment?
```suggestion
// Returns only one InfoStat on DragonflyBSD.  The information regarding core
```

cpu/cpu_dragonfly.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, line := range lines {
+		if matches := cpuEnd.FindStringSubmatch(line); matches != nil {
+			break
+		} else if matches := originMatch.FindStringSubmatch(line); matches != nil {
+			c.VendorID = matches[1]
+			c.Family = matches[3]
+			c.Model = matches[4]
+			t, err := strconv.ParseInt(matches[5], 10, 32)
+			if err != nil {
+				return c, 0, fmt.Errorf("unable to parse FreeBSD CPU stepping information from %q: %v", line, err)
```
shirou, line 143: ditto

cpu/cpu_dragonfly.go
```diff
@@ -0,0 +1,173 @@
...
+				c.Flags = append(c.Flags, strings.ToLower(v))
+			}
+		} else if matches := featuresMatch2.FindStringSubmatch(line); matches != nil {
+			for _, v := range strings.Split(matches[1], ",") {
+				c.Flags = append(c.Flags, strings.ToLower(v))
+			}
+		} else if matches := cpuCores.FindStringSubmatch(line); matches != nil {
+			t, err := strconv.ParseInt(matches[1], 10, 32)
+			if err != nil {
+				return c, 0, fmt.Errorf("unable to parse FreeBSD CPU Nums from %q: %v", line, err)
```
shirou, line 157: ditto

cpu/cpu_dragonfly.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+			cpuNum = int(t)
+			t2, err := strconv.ParseInt(matches[2], 10, 32)
+			if err != nil {
+				return c, 0, fmt.Errorf("unable to parse FreeBSD CPU cores from %q: %v", line, err)
```
shirou, line 162: ditto

shirou:
Thank you for your contribution!. However, this file seems exactly same as `cpu_freebsd.go`. I don't know the DragonflyBSD but I'm curious to see if this PR really works...

If this works, it's ok to add, but could you change the `FreeBSD` word in a error? Thank you!

cpu/cpu_dragonfly.go
gballet: Good catch. I just checked, and parsing the `dmesg` output is unlikely to work on FreeBSD, too, as regular users do not have access to `dmesg`. I'll remove this regexp because it will never match, and the correct value is eventually extracted in `InfoWithContext`.

gballet: ok, I see why you parse the output of `dmesg`: to differentiate cpu vs. cores.

gballet:
The PR works, at least on my machines ;) There are indeed some issues with parsing the number of cores, I'm going to check if it's possible to extract that information without parsing the output of `dmesg` because this is unlikely to remain stable over the releases.

cpu/cpu_dragonfly.go
shirou: Thank you for the information. I found it is not worked correctly in FreeBSD... it should be use `kern.sched.topology_spec` to getting how many psychical core it has. I will fix it, so we can left this at this time.

gballet: I have rewritten the PR to use the equivalent `hw.cpu_topology.tree` in DragonflyBSD. Looking forward to seeing what you come up with on the FreeBSD side.

shirou:
I can not confirm this can works, but I merge it. Thank you for your great contribution!

Issue #679, opened 2019-05-02
host.Virtualization does not detect lxc guest


danielnelson:
**Describe the bug**

Calling host.Virtualization() from inside a LXC container does not detect that it is being ran under LXC.  This prevents the fix in #441 from working.

**To Reproduce**
```go
package main

import (
        "fmt"

        "github.com/shirou/gopsutil/host"
)

func main() {
        system, role, err := host.Virtualization()
        fmt.Println("system:", system)
        fmt.Println("role:", role)
        fmt.Println("err:", err)
}
```

**Actual behavior**
```
system: 
role: 
err: <nil>
```

**Expected behavior**
```
system: lxc
role: guest
err: <nil>
```

**Environment:**
 - [x] Linux

LXC Host: Debian Buster `Linux buster 4.13.0-1-amd64 #1 SMP Debian 4.13.13-1 (2017-11-16) x86_64 GNU/Linux`

```
$ dpkg -l | grep lxc
ii  liblxc1                       1:3.1.0+really3.0.3-8          amd64        Linux Containers userspace tools (library)
ii  lxc                           1:3.1.0+really3.0.3-8          amd64        Linux Containers userspace tools
ii  lxc-templates                 3.0.3-1                        amd64        Linux Containers userspace tools (templates)
ii  lxcfs                         3.0.3-2                        amd64        FUSE based filesystem for LXC
```
LXC Guest: Debian Buster

**Additional context**
This detection technique appears to work:

`cat /proc/1/environ | tr '\0' '\n' | grep ^container`

https://lists.linuxcontainers.org/pipermail/lxc-users/2013-December/005998.html

shirou:
Thank you for reporting! 

memo: [ohai implementation ](https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/virtualization.rb#L190) can help.

PR #894, opened 2020-06-20
LXC Guest Detection from Environ Procfile

citruspi:
This PR adds an additional check for LXC guests using `/proc/1/environ` and resolves 
- https://github.com/shirou/gopsutil/issues/679
- https://github.com/influxdata/telegraf/issues/2968

Implementation notes:
- `ReadFile` was added because `/proc/1/environ` doesn't appear to contain any newlines and `ReadLines` was returning an empty slice
- I left the existing test for LXC guests in place in case it remains valid on some systems, I'd be comfortable removing that if that would be preferred

Tested on a Centos 7 LXC guest.

```sh
[root@centos7-test ~]# ./linux-master 
Virt - system:kvm role:host
Uptime - 327h10m35s
[root@centos7-test ~]# ./linux-branch 
Virt - system:lxc role:guest
Uptime - 45m6s
```

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -194,6 +194,17 @@ func VirtualizationWithContext(ctx context.Context) (string, string, error) {
 		}
 	}
 
+	if PathExists(filepath.Join(filename, "1", "environ")) {
+		contents, err := ReadFile(filepath.Join(filename, "1", "environ"))
+
+		if err == nil {
+			if strings.Contains(contents, "container=lxc") {
+				system = "lxc"
+				role = "guest"
+			}
+		}
+	}
+
 	if PathExists(filepath.Join(filename, "self", "cgroup")) {
 		contents, err := ReadLines(filepath.Join(filename, "self", "cgroup"))
 		if err == nil {
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -94,6 +94,17 @@ func (i FakeInvoke) CommandWithContext(ctx context.Context, name string, arg ...
 
 var ErrNotImplementedError = errors.New("not implemented yet")
 
+// ReadFile reads contents from a file
+func ReadFile(filename string) (string, error) {
+	content, err := ioutil.ReadFile(filename)
+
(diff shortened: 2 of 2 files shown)


internal/common/common_linux.go
```diff
@@ -194,6 +194,17 @@ func VirtualizationWithContext(ctx context.Context) (string, string, error) {
 		}
 	}
 
+	if PathExists(filepath.Join(filename, "1", "environ")) {
+		contents, err := ReadFile(filepath.Join(filename, "1", "environ"))
+
+		if err == nil {
+			if StringsContains([]string{contents}, "container=lxc") {
```
shirou, line 201: `contents` is just a string, so how about to use `strings.Contains` in a standard library?

shirou:
I have confirmed `container=lxc` is in the `/proc/1/environ`.  I commented one thing. If you can change the code, it seems LGTM!

citruspi:
@shirou Thanks for taking a look! I've made the recommended changes.

shirou:
LGTM. Thank you so much!

citruspi:
Happy to contribute :)

Would you be willing to tag a release containing the fix at some point in the near future? I'd like to get LXC uptime monitoring in `telegraf` fixed.

Lomanic:
Releases are done once every month.

Astaoth:
Hi, on a LXC running on a Proxmox 7.1 (debian 11, kernel 5.13.19-6-pve), this fix doesn't seem to work anymore.

Here is what I have in my /proc/1/environ : 
`# cat /proc/1/environ
TERM=linuxcontainer=lxccontainer_ttys= `

Would it be possible to make it works again ?

shirou:
That contents looks containing `container=lxc`, so this [condition](https://github.com/shirou/gopsutil/pull/894/files#diff-2cf69bb8e08ae00fd883bec92baeb4ab3570913ebf2b6fa3f1f077f05368ad7aR201) should work. Can you think of any other reasons?

Issue #853, opened 2020-03-19
FreeBSD build failing


tmm1:
```
[freebsd-x86_64] ../gopath/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/disk/disk_freebsd.go:87:51: cannot use stat.Mntfromname[:] (type []byte) as type []int8 in argument to common.IntToString
[freebsd-x86_64] ../gopath/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/disk/disk_freebsd.go:88:49: cannot use stat.Mntonname[:] (type []byte) as type []int8 in argument to common.IntToString
[freebsd-x86_64] ../gopath/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/disk/disk_freebsd.go:89:50: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
[freebsd-x86_64] ../gopath/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/disk/disk_freebsd.go:173:43: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
```

This just started out of nowhere. Not sure what changed. I think I upgraded x/sys

tmm1:
Looks like there were some changes in x/sys/unix for freebsd in the last few days: https://github.com/golang/sys/commit/5766fd39f98ddd8d769ad4aedcee557dd28de90f#diff-2b37904226feaa1ac0c049523015d0bcL126-R128

ianlancetaylor:
Please open an issue at https://golang.org/issue or discuss at https://golang.org/cl/223698.  Thanks.

i0rek:
We just ran into this. It is not fixed in sys just yet. It could be fixed in gopsutil by casting, would you accept a PR for that?

shirou:
Thanks. PR is always welcome!

xxxserxxx:
@ncabatoff, @samsalisbury, @shirou, and @lomanic: pinning x/sys to v0.0.0-20200316230553-a7d97aace0b0 also works.  Specifically, Hashicorp, it lets you use the current version of `gopsutil`: you don't have to pin to a 2-y/o version.

For anyone finding this, in your project:
```
[...]

PR #895, opened 2020-06-22
Fix FreeBSD build with latest x/sys/unix

tklauser:
There was a breaking change in upstream golang.org/x/sys/unix which
changed the type of some Statfs_t members, see #853. As it looks like
this change won't be reverted, adjust gopsutil to work with it and allow
to build against the latest version of x/sys/unix.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -48,16 +48,18 @@
   version = "v1.6.0"
 
 [[projects]]
-  digest = "1:63d8236278940f9c63816950fb4ed7fa389307c0b6367484b748f5b08dee9299"
+  branch = "master"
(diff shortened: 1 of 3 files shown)


tklauser:
@shirou @Lomanic could one of you please have a look at this if you find time? This would allow users of this package to use the latest version of `golang.org/x/sys/unix` again. Thank you for maintaining this useful package!

Lomanic:
I don't have access to my computer before Sunday. Thanks for the PR and the kind word, appreciated.

shirou:
I have confirmed this patch works on latest golang.org/x/sys. Thank you for your contribute!

PR #896, opened 2020-06-25
Add linux/riscv64 support

tklauser:
The first commit adds the necessary const and type definitions to be able to build on linux/riscv64.

The second commit adds a build test for linux/riscv64


shirou:
sorry, could you revert the first commit which remove unused consts? These are unused now but it will be helpful when we change.

tklauser:
> sorry, could you revert the first commit which remove unused consts? These are unused now but it will be helpful when we change.

FWIW, in case they're ever needed they could be used from `golang.org/x/sys/unix` instead, which exports them for all `GOOS/GOARCH` combinations supported by Go.

In any case, I have dropped the commit from this PR since it is not directly related to `linux/riscv64` support. Thanks.

shirou:
Thank you for quick update.  I can not check it works on linux/riscv64 since I don't have that, but I merge this PR. Thank you for your great work!

PR #897, opened 2020-07-02
Add processes created to misc stats

james-bebbington:
This PR adds the "processes" result from `proc/stat`. Not 100% sure if "ProcsCreated is the best name for this.

https://man7.org/linux/man-pages/man5/proc.5.html lists this as "Number of forks since boot" which I believe is equivalent to the number of processes created since boot. A similar concept should exist for other OSs, though may not be able to be easily retrieved.

### DIFF
--- a/load/load.go
+++ b/load/load.go
@@ -21,6 +21,7 @@ func (l AvgStat) String() string {
 
 type MiscStat struct {
 	ProcsTotal   int `json:"procsTotal"`
+	ProcsCreated int `json:"procsCreated"`
 	ProcsRunning int `json:"procsRunning"`
 	ProcsBlocked int `json:"procsBlocked"`
 	Ctxt         int `json:"ctxt"`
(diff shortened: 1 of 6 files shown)


Lomanic:
It's possible to get this info on openbsd (to be checked on other *bsds) with `sysctl -n kern.forkstat.forks` (found via `vmstat -f` source). Could get this value with `unix.Sysctl("kern.forkstat.forks")` (or `unix.SysctlUint32("kern.forkstat.forks")`) though.

windows kernel doesn't track this as far as I could find.

james-bebbington:
> It's possible to get this info on openbsd (to be checked on other *bsds) with `sysctl -n kern.forkstat.forks

Added your suggested code for openbsd. It doesn't look like this is available on freebsd according to https://www.freebsd.org/cgi/man.cgi?query=sysctl&sektion=8

I'm unable to test this however as I don't have access to a BSD machine.

james-bebbington:
@Lomanic any chance you could take another look at this one?

Lomanic:
Sorry for the delay, thanks for your PR @james-bebbington

PR #898, opened 2020-07-07
Fix MemoryMaps on ARM

tyldavis:
ARM has some tab characters in smaps instead of spaces, hence switching to strings.Fields instead of strings.Split which handles splitting on all whitespace instead of just spaces.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -576,7 +576,8 @@ func (p *Process) MemoryMapsWithContext(ctx context.Context, grouped bool) (*[]M
 			if len(field) < 2 {
 				continue
 			}
-			v := strings.Trim(field[1], " kB") // remove last "kB"
+			v := strings.Trim(field[1], "kB") // remove last "kB"
+			v = strings.TrimSpace(v)
 			t, err := strconv.ParseUint(v, 10, 64)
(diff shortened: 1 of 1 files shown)


tyldavis:
Really looking forward to being able to run tests on ARM via CircleCI: https://ideas.circleci.com/ideas/CCI-I-900

process/process_linux.go
```diff
@@ -610,11 +611,11 @@ func (p *Process) MemoryMapsWithContext(ctx context.Context, grouped bool) (*[]M
 
 	blocks := make([]string, 16)
 	for _, line := range lines {
-		field := strings.Split(line, " ")
-		if strings.HasSuffix(field[0], ":") == false {
+		fields := strings.Fields(line)
+		if len(fields) > 0 && strings.HasSuffix(fields[0], ":") == false {
```
Lomanic, line 615: `&& !strings.HasSuffix(fields[0], ":")` would be cleaner.

tyldavis: Updated, thanks.

shirou:
Thank you @tyldavis  and @Lomanic ! I also not tested yet, but merge anyway.

Issue #902, opened 2020-07-10
In linux cpu iowait should be treated like idle time. 


msays2000:
https://github.com/shirou/gopsutil/blob/6d64cd6d43e7927e70891507b795c176a60caf37/cpu/cpu.go#L99

If iowait implies cpu is technically free to do any other ready work, then we should treat it as idle time. 

In python psutil and in htop it is treated as idle time. 

https://github.com/giampaolo/psutil/blob/550c825c95b537e9c3fee8452e5e1cd4fe5c704a/psutil/__init__.py#L1664-L1677

https://github.com/torvalds/linux/blob/447976ef4fd09b1be88b316d1a81553f1aa7cd07/kernel/sched/cputime.c#L244-L253

msays2000:
sample PR: https://github.com/shirou/gopsutil/pull/903

Grandmother:
Fixed in this commit: https://github.com/shirou/gopsutil/commit/d1cc7d2125cdccc309c104917b221f08ee0fa731

Grandmother:
@shirou, I can't close the issue, could you please?

PR #903, opened 2020-07-10
use iowait time as idle time

msays2000:
https://github.com/shirou/gopsutil/issues/902

If iowait implies cpu is technically free to do any other ready work, then we should treat it as idle time.

In python psutil and in htop it is treated as idle time.

https://github.com/giampaolo/psutil/blob/550c825c95b537e9c3fee8452e5e1cd4fe5c704a/psutil/__init__.py#L1664-L1677

https://github.com/torvalds/linux/blob/447976ef4fd09b1be88b316d1a81553f1aa7cd07/kernel/sched/cputime.c#L244-L253

### DIFF
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -97,9 +97,9 @@ func (c InfoStat) String() string {
 }
 
 func getAllBusy(t TimesStat) (float64, float64) {
-	busy := t.User + t.System + t.Nice + t.Iowait + t.Irq +
-		t.Softirq + t.Steal
-	return busy + t.Idle, busy
+	busy := t.User + t.System + t.Nice + t.Irq +
+		t.Softirq + t.Steal + t.Guest + t.GuestNice
+	return busy + t.Idle  + t.Iowait, busy
 }
 
 func calculateBusy(t1, t2 TimesStat) float64 {


msays2000:
@shirou please approve or deny this PR

shirou:
Sorry for late response. I have checked psutil code and [_cpu_tot_time](https://github.com/giampaolo/psutil/blob/550c825c95b537e9c3fee8452e5e1cd4fe5c704a/psutil/__init__.py#L1645) change behavior by Linux or not. So we should add this check.

msays2000:
> Sorry for late response. I have checked psutil code and [_cpu_tot_time](https://github.com/giampaolo/psutil/blob/550c825c95b537e9c3fee8452e5e1cd4fe5c704a/psutil/__init__.py#L1645) change behavior by Linux or not. So we should add this check.

@shirou the busy time should exclude iowait is the change in this PR:

[_cpu_busy_time](https://github.com/giampaolo/psutil/blob/550c825c95b537e9c3fee8452e5e1cd4fe5c704a/psutil/__init__.py#L1664-L1676)

msays2000:
@shirou please merge this PR on reviewing and it being correct. Thank you.

shirou:
Exclude from iowait is ok. but I think adding guest and guest nice for all platform seems different from psutil implementation.

msays2000:
@shirou ok I can revert the guest and guest nice from busy. will send a updated PR.

Grandmother:
Fixed in this commit: https://github.com/shirou/gopsutil/commit/d1cc7d2125cdccc309c104917b221f08ee0fa731
The check of linux/not linux also added there.

PR #904, opened 2020-07-11
Added support for reading hardware fan speeds in Linux.

tymonx:
* Added the host.SensorsFans() function
* Added the host.SensorsFansWithContext() function
* Added the FanStat struct
* Added support for Linux and not implemented error in other OSs

Example output:

[{"sensorKey":"nct6776","sensorSpeed":2768} {"sensorKey":"nct6776","sensorSpeed":3161}]

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -38,6 +38,12 @@ type TemperatureStat struct {
 	Temperature float64 `json:"sensorTemperature"`
 }
 
+// FanStat describes a fan status.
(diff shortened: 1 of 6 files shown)


shirou:
Sorry for late response. But I don't want to add Linux specific metrics anymore.

shirou:
Thank you for raising PR, but closing for now.

PR #905, opened 2020-07-11
Fixed temperature sensors detection in Linux

tymonx:
Previous implementation returns all values from the /sys/class/hwmon/hwmon*/temp* location.

It was incorrect and very misleading. The SensorsTemperatures() should return only list of sensors.

Current temperature values are only in the /sys/class/hwmon/hwmon*/temp*_input location.

Reference: https://www.kernel.org/doc/Documentation/hwmon/sysfs-interface

Other files represent thresholds, alarms, setups.

* Added new optional fields, High (*_max) and Critical (*_crit)
* Removed redundant _input suffix. name + label is sufficient
* Now temperatures, _ := host.SensorsTemperatures(); len(temperatures) returns exacts number of temperature sensors
* Fixed label strings transformations. Replaced ' ' with '_'. Previous implementation transformed complex labels to
  unreadable and hard to parse strings

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -36,6 +36,8 @@ type UserStat struct {
 type TemperatureStat struct {
 	SensorKey   string  `json:"sensorKey"`
 	Temperature float64 `json:"sensorTemperature"`
+	High        float64 `json:"sensorHigh"`
+	Critical    float64 `json:"sensorCritical"`
 }
 
 func (h InfoStat) String() string {
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -29,7 +29,11 @@ type LSB struct {
 }
 
 // from utmp.h
-const USER_PROCESS = 7
+const (
+	USER_PROCESS = 7
+
+	hostTemperatureScale = 1000.0
+)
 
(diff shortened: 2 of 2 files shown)


PedroLamas:
I'm experiencing this same issue on my Armbian based linux, as it contains the "_label" but not the "_input" required for this

image

This code change would fix my problem!

Lomanic:
This makes gopsutil more aligned with psutil and with more sensible results returned, which is always good. Though, it breaks backward compatibility for existing users, hence why I've been reluctant merging this.

Also, [this test](https://github.com/shirou/gopsutil/blob/d9f9a85e75b41b9555783b2eee705daf7f3a9b43/host/host_test.go#L120) should be fixed with the new fields.

PedroLamas:
It seems Armbian fixed this from their side, as I now have the "temp*_input" after updating to the latest firmware:

image

Telegraf now correctly reports the temperature!

shirou:
I will merge this under v3 directory after v3 released (end of this month, perhaps).

shirou:
Merged by #938. Thank you for your contribution!

PR #909, opened 2020-07-24
add go.mod and add "v2" to import path.

shirou:
add `gomod` and `v2` to import paths.

This PR should be merged 
- after golang 1.15 release and
- before releasing v2.20.7.

Because go module becomes default enable after 1.13, and 1.12 will be EoL at 1.15 release. So we should add go.mod after that.
and adding go.mod is not enough. To add latest go mod supported version to https://proxy.golang.org, new git-tag version is required. so we should merge this PR at certain timing.

If releasing 1.15 becomes delayed, merge will be postponed to next monthly release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,15 @@
+module github.com/shirou/gopsutil/v2
+
+go 1.14
+
+require (
+	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d
+	github.com/davecgh/go-spew v1.1.1
+	github.com/go-ole/go-ole v1.2.4
+	github.com/pmezard/go-difflib v1.0.0
+	github.com/stretchr/testify v1.6.0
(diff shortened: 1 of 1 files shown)


Lomanic:
Not a fan at all as it forces to use modules. Would it be possible to use `replace` in go.mod so module users can solve the v2 paths while still omitting it on disk for users not using modules?

shirou:
replace! yes, you are right. I will try to use replace. Thank you!

imsodin:
This will only have an effect when working on gopsutil directly (I think, not sure about that), but it will have no effect on other packages importing this package (gopsutil):

> replace directives only apply in the main module's go.mod file and are ignored in other modules. See Minimal version selection for details.

https://golang.org/ref/mod#go

shirou:
Ummmmm... thank you for the information. so it's better to publish v3 instead of v2 to adapt Go Module. I also not a fan of it, but it now becomes a standard, unfortunately.

PR #910, opened 2020-07-28
Update net_windows.go

Apoorvashaan:


Lomanic:
This is Windows-specific while gopsutil is meant to be platform-agnostic, this is an implementation detail for this platform not meant to be exported.

PR #920, opened 2020-08-17
Add arch buildconstraints on disk for openbsd (fixes #912)

### LINKED ISSUE #912: Does not build on openbsd-arm
As of the current versions of relevant packages (I think), this does not build on freebsd-arm. (It used to be it didn't build on freebsd-arm64, I updated everything to fix that, which indeed works now.)

```
jb@kvin:../syncthing/syncthing % egrep gopsutil\|org/x/ go.mod
	github.com/shirou/gopsutil v2.20.6+incompatible
	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
	golang.org/x/net v0.0.0-20200707034311-ab3426394381
	golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd
	golang.org/x/text v0.3.3
	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4
jb@kvin:../syncthing/syncthing % go version
go version go1.14.5 darwin/amd64
jb@kvin:../syncthing/syncthing % GOOS=freebsd GOARCH=arm go build ./cmd/...
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.6+incompatible/disk/disk_freebsd.go:87:52: cannot use stat.Mntfromname[:] (type []int8) as type []byte in argument to common.ByteToString
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.6+incompatible/disk/disk_freebsd.go:88:50: cannot use stat.Mntonname[:] (type []int8) as type []byte in argument to common.ByteToString
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.6+incompatible/disk/disk_freebsd.go:89:51: cannot use stat.Fstypename[:] (type []int8) as type []byte in argument to common.ByteToString
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.6+incompatible/disk/disk_freebsd.go:173:44: cannot use stat.Fstypename[:] (type []int8) as type []byte in argument to common.ByteToString
```

(I don't envy anyone trying to keep up with the type changes in sys around this...)

imsodin:
There's definitions for amd64 and 386, but nothing else on openbsd. In the version we used previously the build constraint was `openbsd,amd64`, now it's just `openbsd`. I now added build constraints for either `amd64` or `386`, thus "fixing" the build on arm (no functionality, but also no build failure).

### DIFF
--- a/disk/disk_openbsd.go
+++ b/disk/disk_openbsd.go
@@ -1,4 +1,5 @@
 // +build openbsd
+// +build amd64 386
 
 package disk
 


Lomanic:
👎 from me, there is nothing specific to these two archs in this file. 

To fix #912, `disk_openbsd_arm.go` should be generated with `GOOS=openbsd GOARCH=arm go tool cgo -godefs disk/types_openbsd.go > disk/disk_openbsd_arm.go` (or preferably directly on an arm host).

imsodin:
Shouldn't there still be constraints? It uses variables defined in these disk_openbsd_\*arch\*.go files, thus builds will fail on all archs for which no such file exists. It's much nicer to have no functionality instead of a broken build (at least for Syncthing, which does work without gopsutil).

Lomanic:
You will still have broken functionality (broken build) if you want to call an undefined function (because of build constraints) though?

I don't have any of the [armv7 boards supported by OpenBSD](https://www.openbsd.org/armv7.html#hardware) to generate or test everything works, it's plausible that copying the 386 definition for arm would work (`cp gopsutil/disk/disk_openbsd_{386,arm}.go`).

imsodin:
That makes sense, I guess I am used to fallback stubs that just return an error when invoked. What doesn't make sense is that with the constraint on the previous gopsutil version we used (`+build openbsd,amd64`) we did cross-compile and publish releases for lots of arches other than amd64 on openbsd - something's fishy there. But anyway, without fallback constraints are indeed pointless - sorry for the noise and thanks for explaining ;)

PR #923, opened 2020-08-20
net_linux.go: decode port as uint16 instead if int64

disconnect3d:
Changes the port parsing from `/proc/net/*` files records from parsing them as 64-bit integers to parse them as 16-bit unsigned integers.

While this is mostly a cosmetic change, it will also make so that the code fails faster in case the entry is malformed (for whatever reason).

Note that the returned value is still casted to uint32 when an  `Addr` object is created. 
It seems to me that the `Addr.port` field should be changed to `uint16` but maybe some other APIs/systems wants it to be `uint32` and also changing it there may require changes in users code if they update. This being said I am not changing that field's type.

EDIT: Additionally, I am also changing the `base` used for parsing to 16 and removing the addition of `"0x"` string prefix. The `base=16` is the appropriate base here and the `"0x"` prefix is redundant after we explicitly specify the base.

### DIFF
--- a/net/net_linux.go
+++ b/net/net_linux.go
@@ -696,7 +696,7 @@ func decodeAddress(family uint32, src string) (Addr, error) {
 		return Addr{}, fmt.Errorf("does not contain port, %s", src)
 	}
 	addr := t[0]
-	port, err := strconv.ParseInt("0x"+t[1], 0, 64)
+	port, err := strconv.ParseUint(t[1], 16, 16)
 	if err != nil {
 		return Addr{}, fmt.Errorf("invalid port, %s", src)
 	}


net/net_linux.go
```diff
@@ -696,7 +696,7 @@ func decodeAddress(family uint32, src string) (Addr, error) {
 		return Addr{}, fmt.Errorf("does not contain port, %s", src)
 	}
 	addr := t[0]
-	port, err := strconv.ParseInt("0x"+t[1], 0, 64)
+	port, err := strconv.ParseUint([1], 16, 16)
```
shirou, line 699: Test fails due to 
```
./net_linux.go:699:36: syntax error: unexpected comma, expecting type
```

Is this what you want?
```suggestion
	port, err := strconv.ParseUint(t[1], 16, 16)
```

Lomanic:
Please squash your commits also.

disconnect3d:
> Please squash your commits also.

Done

Lomanic:
Why do you specify a bit size of 16 (uint16) and not 32 (uint32) as in fact uint32 is the final type this port is converted to?

disconnect3d:
> Why do you specify a bit size of 16 (uint16) and not 32 (uint32) as in fact uint32 is the final type this port is converted to?


Because any reasonable system has a port range of 0-~65k and doesn't exceed the uint16 range.

As I said previously, I think the `Addr.port` should be changed to `uint16` but that's probably a bigger change which I think should not belong here.

We can change this parsing to `uint32` here, but that may just delay some issues like a too-big-port-values being overflowed into uint16 eventually or (properly) validated (in case of possible port range) and rejected.

Lomanic:
OK, that's fine by me.

> As I said previously, I think the Addr.port should be changed to uint16 but that's probably a bigger change which I think should not belong in this PR.

[You're indeed right](https://stackoverflow.com/a/113228) (and this limit is platform-independent), but it's too late for this as it would break downstream gopsutil users. Though, the fix for users is really simple (and could even fix some flawed code?).

PR #924, opened 2020-08-25
fix an invalid shift in windows processes name UTF16 conversion

nightmared:
Hello there,

this trivial PR should fix the conversion of UTF16 byte slices to strings, which was invalid in the presence of "high" UCS-2 or unicode codepoints.

Have a nice day !


Lomanic:
I didn't encounter this but the change is sensible. Thanks!

PR #926, opened 2020-08-28
use new syscall on darwin

### LINKED ISSUE #925: Darwin unix.SYS___SYSCTL has been removed and causes gopsutil to break
**Describe the bug**
Three days ago a change was made to drop these old syscalls https://github.com/golang/sys/commit/6fcdbc0bbc04dcd9e7dc145879ceaf9bf1c6ff03 and as a result gopsutil breaks as it was relying on this old syscall version

https://github.com/shirou/gopsutil/blob/7e94bb8bcde053b6d6c98bda5145e9742c913c39/internal/common/common_darwin.go#L39

**To Reproduce**
Use gopsutil with the latest version of golang.org/x/sys

Can fix by using unix.SYS_SYSCTL instead

audiolion:
Go has officially removed support for the old style of syscalls.
See https://github.com/golang/sys/commit/6fcdbc0bbc04dcd9e7dc145879ceaf9bf1c6ff03

fixes #925

### DIFF
--- a/internal/common/common_darwin.go
+++ b/internal/common/common_darwin.go
@@ -36,7 +36,7 @@ func CallSyscall(mib []int32) ([]byte, uint64, error) {
 	// get required buffer size
 	length := uint64(0)
 	_, _, err := unix.Syscall6(
-		unix.SYS___SYSCTL,
+		unix.SYS_SYSCTL,
 		uintptr(unsafe.Pointer(&mib[0])),
 		uintptr(miblen),
 		0,
@@ -54,7 +54,7 @@ func CallSyscall(mib []int32) ([]byte, uint64, error) {
 	// get proc info itself
 	buf := make([]byte, length)
 	_, _, err = unix.Syscall6(
-		unix.SYS___SYSCTL,
+		unix.SYS_SYSCTL,
 		uintptr(unsafe.Pointer(&mib[0])),
 		uintptr(miblen),
 		uintptr(unsafe.Pointer(&buf[0])),
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -611,7 +611,7 @@ func (p *Process) getKProcWithContext(ctx context.Context) (*KinfoProc, error) {
 	length := uint64(unsafe.Sizeof(procK))
 	buf := make([]byte, length)
 	_, _, syserr := unix.Syscall6(
-		unix.SYS___SYSCTL,
+		unix.SYS_SYSCTL,
 		uintptr(unsafe.Pointer(&mib[0])),
 		uintptr(len(mib)),
 		uintptr(unsafe.Pointer(&buf[0])),


Lomanic:
This breaks building the `internal/common` package on darwin with `GOOS=darwin go build github.com/shirou/gopsutil/internal/common` when using a previous version of golang.org/x/sys (for example, [1151b9dac4a98d49ef7f80f07ddd826ff51e0b36](https://github.com/golang/sys/tree/1151b9dac4a98d49ef7f80f07ddd826ff51e0b36)).

What about using `unix.SYS_SYSCTL`? The deleted variable `unix.SYS___SYSCTL` was set to this value in fact https://github.com/golang/sys/commit/6fcdbc0bbc04dcd9e7dc145879ceaf9bf1c6ff03#diff-6453904623ddb4547da2ded4240ca065L49 and it doesn't break

audiolion:
@Lomanic thanks for the tip!

Lomanic:
Hey, you're so fast to fix this :)

Could you fix it there too? https://github.com/shirou/gopsutil/blob/7e94bb8bcde053b6d6c98bda5145e9742c913c39/process/process_darwin.go#L614

Before merging your PR I would like to wait for https://go-review.googlesource.com/c/sys/+/251317/ to be merged, as indeed golang.org/x/sys/unix is broken on Go 1.12 on darwin https://github.com/golang/sys/commit/6fcdbc0bbc04dcd9e7dc145879ceaf9bf1c6ff03#commitcomment-41842771 

> I don't know  how to update gopsutil's gopkg.lock to use the latest master version of golang.org/x/sys

Gopkg.lock is generated using [dep](https://github.com/golang/dep), [you can update  it with](https://golang.github.io/dep/docs/daily-dep.html#updating-dependencies) `dep ensure -update golang.org/x/sys`. Here is the result on my machine

```diff
diff --git a/Gopkg.lock b/Gopkg.lock
index 04ca6c2..e01f4a9 100644
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -49,7 +49,7 @@

 [[projects]]
   branch = "master"
-  digest = "1:e7b553682850788c6b266faaf25218ce901558e541102d12b2998cad99f6b4fb"
+  digest = "1:a402b847cce9627ad9c06aeb22e0f7dbadec4a4edd3b88756da530ec94fa34fe"
   name = "golang.org/x/sys"
   packages = [
     "internal/unsafeheader",
@@ -59,7 +59,7 @@
     "windows/svc/mgr",
   ]
   pruneopts = "UT"
-  revision = "76b94024e4b621e672466e8db3d7f084e7ddcad2"
+  revision = "c663848e9a16c3f7dcfaea95e7ea787d9543cb30"

 [[projects]]
   branch = "v3"
```

But this will be obsolete after they merge https://go-review.googlesource.com/c/sys/+/251317/

audiolion:
Ok will update there as well. When I tried to use dep to update it told me:
image

Which I am assuming means I need to move it to `$GOPATH/src/github.com/audiolion/gopsutil` or something to get it to work? But I didn't want to mess around with something that I hadn't taken the time to understand.

Lomanic:
> Which I am assuming means I need to move it to `$GOPATH/src/github.com/audiolion/gopsutil`

Indeed. `$GOPATH/src/github.com/shirou/gopsutil` even maybe (that's what I do).

I will update your branch with the result if you don't want to mess with GOPATH in your environment.

audiolion:
yeah I wasn't sure where to put it because I forked the repo to be able to make the PR, so I didn't know if that was going to mess things up having it be github.com/audiolion instead of github.com/shirou

tklauser:
FWIW, https://golang.org/cl/251738 will re-add `SYS___SYSCTL` on darwin. However, I think the proper fix would be to use `unix.SysctlRaw` (or similar) which will use the libc-wrappers rather than the direct syscalls.

Also, from a quick `git grep` through the sources it looks like `internal/common.CallSyscall` isn't used on `GOOS=darwin` at all. Thus, the easiest fix might be to just remove the function given it is in the `internal` package and thus non-public API.

Lomanic:
You're totally right @tklauser, I was really confused after https://go-review.googlesource.com/c/sys/+/251317/ was merged to not find any exported `*_SYSCTL` (`GOOS=darwin go doc golang.org/x/sys/unix | grep _SYSCTL`) after upgrading x/sys/unix.

NateWilliams2:
Hey just want check on this, I have a dependency which uses gopsutil and now cannot build on MacOS. Does this have a simple fix which could be merged now, even if it's not the best way to proceed in the long-term? I'd appreciate a patch-fix if the proper fix will take a while :)

Lomanic:
Quick fix outside of gopsutil is to use the version of x/sys/unix specified in Gopkg.lock (ie use a version before the breaking change).

PR #930, opened 2020-08-31
Use unix.SysctlUvmexp on OpenBSD

tklauser:
Use SysctlUvmexp from golang.org/x/sys/unix to avoid having to define
the type Uvmexp and the sysctl consts. This will also allow to build on
GOOS=openbsd with e.g. GOARCH=arm64.


Lomanic:
Always nice to see `+6 −121` in a diff, thanks a lot @tklauser

PR #931, opened 2020-08-31
Fix TestVirtual_memory on OpenBSD

tklauser:
On OpenBSD, the total is used + free + cached + inactive like on macOS.


Lomanic:
Thanks @tklauser

PR #932, opened 2020-08-31
Use unix.SysctlUint32("hw.ncpuonline") on OpenBSD

tklauser:
Use SysctlUvmexp from golang.org/x/sys/unix to avoid having to simplify
the implementation of cpu.InfoWithContext

### DIFF
--- a/cpu/cpu_openbsd.go
+++ b/cpu/cpu_openbsd.go
@@ -32,7 +32,6 @@ const (
 	CTLKern     = 1  // "high kernel": proc, limits
 	CTLHw       = 6  // CTL_HW
 	SMT         = 24 // HW_SMT
-	NCpuOnline  = 25 // HW_NCPUONLINE
 	KernCptime  = 40 // KERN_CPTIME
 	KernCptime2 = 71 // KERN_CPTIME2
 )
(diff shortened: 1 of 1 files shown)


cpu/cpu_openbsd.go
```diff
@@ -169,19 +168,10 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
...
-	buf, _, err := common.CallSyscall(mib)
-	if err != nil {
-		return nil, err
-	}
-
-	var ncpu int32
-	br := bytes.NewReader(buf)
-	err = binary.Read(br, binary.LittleEndian, &ncpu)
-	if err != nil {
+	if u32, err = unix.SysctlUint32("hw.ncpuonline"); err != nil {
```
Lomanic, line 171: could you please rename `u32` to `ncpu`?

Lomanic (changes requested):

Lomanic:
Besides my small nit, this is a great modification, didn't know about this function in x/sys/unix, thanks for this PR @tklauser!

cpu/cpu_openbsd.go
tklauser: The idea was to reuse the `var` already declared and used to extract `hw.cpuspeed`, but I agree having a separate, properly named `var ncpu` makes it easier to read. FWIW, I also changed it above and renamed the existing `u32` to `mhz`.

tklauser:
Thanks @Lomanic, addressed the nit and also made the corresponding change when getting `unix.SysctlUint32("hw.cpuspeed")`. PTAL.

PR #933, opened 2020-08-31
[host][mem][process][openbsd] Add 386 const and types definitions

Lomanic:
Continuation of #721, this type using mktypes.sh that I didn't know at the time.

Weirdly, `go tool cgo -godefs` doesn't produce propery go-fmt-ed code, had to do run go-fmt manually afterwards.


Lomanic:
Tests are OK and I only executed mktypes.sh, merging as there's not much to review.

PR #934, opened 2020-09-01
Add HardwareCorrupted in meminfo

zrui65:


Lomanic:
@shirou do we need yet another linux-specific value? [github.com/prometheus/procfs](https://github.com/prometheus/procfs) is meant to be extensive because it's only for Linux, while gopsutil value is to be cross-platform. I think that people interested in linux-specific statistics should use github.com/prometheus/procfs instead.

shirou:
Sorry for late. I agree to @Lomanic. I don't want to add Linux specific value anymore. [psutil](https://psutil.readthedocs.io/en/latest/#memory) have only a common values.

PR #936, opened 2020-09-03
Use cancelable sleep

ninedraft:
Use cancelable sleep in cpu.PercentWithContext and process.Process.PercentWithContext


shirou:
Wow, great PR! Thank you so much!

Issue #545, opened 2018-07-02
mem.VirtualMemoryStat JSON fields capitalization and TestX_String tests


Lomanic:
There are quite a few JSON fields all in lower case at [mem/mem.go](https://github.com/shirou/gopsutil/blob/6a38ce20691a51c20533cf0bcb2258162e7d6dbd/mem/mem.go#L53) and following lines, whereas they only have their first capital letter in lower case in other packages and keep their initial capitalization. Shall we fix that in this package?

In the mean time, we should fix [the TestX_String tests like this one](https://github.com/shirou/gopsutil/blob/6a38ce20691a51c20533cf0bcb2258162e7d6dbd/mem/mem_test.go#L52) to use [reflection](https://play.golang.org/p/jo9S9wj3WWO) to dynamically get the fields instead of hard coding them inside the tests and to check their proper capitalization/name.

PR #938, opened 2020-09-07
v3 migration

shirou:
Thanks @Lomanic , PR count gradually becomes lower and golang mod becomes enables by default, I am thinking about migration to v3 (finally, sorry)

This PR includes migration scripts from current v2 to v3. to use this script, `bash tools/v3migration/v3migration.sh` with go build environment.

# issues

All issues should be also filed in #362.

- [x] RlimitStat change to int64 from uint64 (#346)
- [x] remove `process.NetIoCounter()` (#429)
- [x] fix typo: "memoryLimitInBbytes" -> memoryLimitInBytes (#464)
- [x] add `WithContext` methods (#469)
- [x] `process.GetWin32Proc` should not be exposed
- [x] mem.VirtualMemoryStat JSON fields proper capitalization (#545)
- [x] more broadly, have the same public API/types on all platforms (no more public API/types/constants defined only for some platforms)
- [x] return detailed state in `process.State` (as `[]string, err`), instead of just one state (currently, R: Running S: Sleep T: Stop I: Idle Z: Zombie W: Wait and L: Lock). Linked to #596 (determining background state, see https://github.com/shirou/gopsutil/issues/596#issuecomment-434207724)
- [x] get docker container cpu usage from cpuacct.usage (#590 and #581)
- [x]  disk.Opts should be string[] not string. (related to #955)
- [x] Fixed temperature sensors detection in Linux  (#905)
- [x] disk.GetDiskSerialNumber is only for linux. Rename SerialNumber, change return type to `(string, error)`, and returns NotImplementError for other platforms
- [x] disk.GetLabel is only for linux.  Rename to `Label`, change return type to `(string, error)`, and returns NotImplementError for other platforms and
- [x] Change net.InterfaceStat.Addrs to InterfaceAddrList #226

# schedule

* ～ 2020-09-30:Collect issues with breaking changes.  implement migration script. adding v3 to this PR is not permitted. so we can merge PR into current v2 if no breaking change.
* 2020-09-30: Add v3 directory to this PR
* ～ 2020-10-31: Start to implement complex change by hand.
* 2020-10-31: Release `v3.20.10`


shirou:
- [ ] disk.Opts should be `string[]` not string.  (related to #955)

Lomanic:
Maybe this PR should target a new `v3` branch, this way we could be iterating in this branch in smaller steps. What do you think @shirou?

shirou:
Thank you foe the comment. I think v3 directory in `master` branch is better to follow go module manner. so this PR targets master branch.

and after merge this PR, we can still develop v2 because different directory.

shirou:
- [] merge #905 to v3.

Issue #842, opened 2020-02-19
Processes() is extremely CPU intensive


xxxserxxx:
**Describe the bug**
`Processes()` is significantly more resource intensive than even forking `ps` with `exec.Command` and parsing the results.

**To Reproduce**
Download the [attached go source code](https://github.com/shirou/gopsutil/files/4227318/perf.txt) (and rename to `perf.go` it because github is dumb) and run it on the command line with:
```
go run ./perf.go 500
```
The argument is the number of iterations; you may need to start with fewer if it runs too long.

**Expected behavior**
I'd expect the native code to be more resource efficient than spawning a child `ps` process.  On my machine, I consistently see CPU use of between 2x and 4x CPU use, and much slower results:
```
➜  gopsutil go run perf.go 500
Forking ##########################################
....X....X....X....X....X....X....X....X....X....
Average CPU: 802
Time: 6.463249387s
GoPsUtil #########################################
....X....X....X....X....X....X....X....X....X....
Average CPU: 5563
Time: 13.694322005s
```

**Environment (please complete the following information):**
 - [X] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
HOME_URL="https://www.archlinux.org/"
➜  gopsutil uname -a
Linux glamdring 5.5.4-arch1-1 #1 SMP PREEMPT Sat, 15 Feb 2020 00:36:29 +0000 x86_64 GNU/Linux
```
(trimmed some attributes out)

```
github.com/shirou/gopsutil v2.20.1+incompatible
```

AtakanColak:
@xxxserxxx could you also try [go-sysinfo](https://github.com/elastic/go-sysinfo)? I've been wondering but not much time to try it atm.

xxxserxxx:
Yeah! Thanks; this gives me some impetus to build out proper benchmarks for devices.

AtakanColak:
Np, please feel free to share the results, it would be very helpful to me -and anyone needing a process data library in Go- as well.

xxxserxxx:
I wrote a benchmark, and am attaching it here; again, github has a restricted set of suffixes, so resuffix the file to `.go` first.

[...]

PR #942, opened 2020-09-11
Cache common/common_linux.Virtualization()

AtakanColak:
By assuming virtualization environment won't change during a the program's runtime, we can cache common/common_linux.Virtualization() with a simple map to reduce amount of system calls. I first mentioned this issue at [890](https://github.com/shirou/gopsutil/pull/890#issuecomment-690211919)

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -110,7 +110,14 @@ func Virtualization() (string, string, error) {
 	return VirtualizationWithContext(context.Background())
 }
 
+var virtualizationCache map[string]string = nil
+
 func VirtualizationWithContext(ctx context.Context) (string, string, error) {
+	// if cached already, return from cache
+	if virtualizationCache != nil {
+		return virtualizationCache["system"], virtualizationCache["role"], nil
+	}
+	
 	var system string
 	var role string
 
@@ -231,6 +238,13 @@ func VirtualizationWithContext(ctx context.Context) (string, string, error) {
 			role = "host"
 		}
 	}
+	
+	// before returning for the first time, cache the system and role
+	virtualizationCache = map[string]string{
+		"system":	system,
+		"role": 	role,
+	}
+	
 	return system, role, nil
 }
 
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -110,7 +110,7 @@ func Virtualization() (string, string, error) {
 	return VirtualizationWithContext(context.Background())
 }
 
-var virtualizationCache map[string]string = nil
+var virtualizationCache map[string]string
 
 func VirtualizationWithContext(ctx context.Context) (string, string, error) {
 	// if cached already, return from cache


AtakanColak:
I wanted to compare profiling results of this optimization. 

Before optimization of Virtualization(), there is a distinct compute weight on fillFromStatWithContext which calls Virtualization()

image

After optimization, we can clearly see that the compute weight is reduced and spread to the other functions. 

image

shirou:
I think these kind of optimization should be implemented in Application side, not library side.

AtakanColak:
@shirou could you elaborate on that? Most of the compute is spent with ReadLines on Virtualization which is in internal/common package. How could we optimize it from application side?

shirou:
```
package main

import (
        "fmt"
        "github.com/shirou/gopsutil/host"
)

var (
        h = ""
        g = ""
)

func main() {
        if h == "" {
                h, g, _ := host.Virtualization()
                fmt.Println(h, g)
        } else {
                fmt.Println(h, g)
        }
}
```
or whatever, in your application.

AtakanColak:
@shirou thank you for your reply. However your answer would be the case if it was my app who was calling Virtualization information, but it isn't. The function that calls common/common_linux.Virtualization() is fillFromStatWithContext() which is called by many process.Process methods in process_linux. We can't change private functions from the application side.

Lomanic:
In fact the call to `common.Virtualization()` (should be VirtualizationWithContext() probably also) comes from [common.BootTimeWithContext()](https://github.com/shirou/gopsutil/blob/8a625ec054ea912247e0372de6b2a02255a6c09b/internal/common/common_linux.go#L58) which is not cached anymore since #857. As I wrote in https://github.com/shirou/gopsutil/pull/857#issuecomment-629858607, moving to unix.Sysinfo would maybe remove this call.

I fear caching this will make it harder to run unit tests for this function (we would have to hook to this new variable `virtualizationCache` I think).

shirou:
ah, sorry for my misunderstanding. But as @Lomanic said, I also cache is not good for library. I will try to open a PR about using  `unix.Sysinfo`

AtakanColak:
@Lomanic #857 mention a device can start with wrong date but the question we must ask for this case is can a device start with wrong virtualization data? I don't think it can but of course we must be certain.

I'm not going to comment on unit tests as I'm not sure on potential effects of caching on unit tests.

Lomanic (changes requested):

Lomanic:
I think maybe we should cache this value in fact, for unit tests we can work around this by resetting this internal variable to nil if we want to call Virtualization() more than once. This would really improve performances for process.Processes() (see https://github.com/shirou/gopsutil/issues/842#issuecomment-697026725)

Lomanic:
Here's a proper benchmarck

```bash
git -C "$(go env GOPATH)/src/github.com/shirou/gopsutil" checkout master
git -C "$(go env GOPATH)/src/github.com/shirou/gopsutil" reset --hard origin/master
git -C "$(go env GOPATH)/src/github.com/shirou/gopsutil" clean -f -d
cat <<EOF >> "$(go env GOPATH)/src/github.com/shirou/gopsutil/process/process_test.go"
func BenchmarkNewProcess(b *testing.B) {
	checkPid := os.Getpid()
	for i := 0; i < b.N; i++ {
		NewProcess(int32(checkPid))
    }
}
EOF
printf "\nBefore change:\n"
go test -bench=. github.com/shirou/gopsutil/process
curl -sL https://github.com/shirou/gopsutil/pull/942.diff | git apply
printf "\nAfter change:\n"
go test -bench=. github.com/shirou/gopsutil/process
```

Results are telling:
```
Before change:
goos: linux
goarch: amd64
BenchmarkNewProcess-4               2000            542412 ns/op
PASS
ok      github.com/shirou/gopsutil/process      3.106s

After change:
goos: linux
goarch: amd64
BenchmarkNewProcess-4              20000             78478 ns/op
PASS
ok      github.com/shirou/gopsutil/process      4.158s
```

AtakanColak (commented):
removed explicit nil

AtakanColak:
@Lomanic removed explicit nil as requested. Thank you for benchmarking results. Looking forward for the merge.

shirou (approved):
As I said, I don't like cache in a library, but I don't have any other good solution, so I approve this PR. Thank you for your contribution!

reimda:
I tested this fix in telegraf and found that it makes performance similar to before #837 was merged.  Will this PR make it into a release soon?  We'd like to include it in telegraf 1.16.0 which is planned for next week.

shirou:
@Lomanic sorry for interrupting your comment, but I think your request is resolved. So I merge this PR. Thank you for your review and contribution!

shirou:
gopsutil is monthly release. After this PR merged, I have released `vv2.20.9`.

PR #943, opened 2020-09-11
Refactor "host" package

boogie-byte:
This has stared as an attempt to export a `HostID` exporter, but the resulting patch is mostly refactoring. Here are the main changes:

- `InfoWithContext()` function was unified across all platforms, and now relies on platform-specific function implementations.
- `InfoWithContext()` function now has the same behavior across all platforms (for example, some platfors ignored underlying errors, and some returned them).
- Call context is now passed to all underlying functions in `InfoWithContext()`.
- All context-less wrapping functions (the ones without `WithContext` suffix) were moved into `host.go` since they all are the same.
- `KernelArch()` function is now exported since there was no way to get its value without calling the `InfoWithContext()`, which is clearly an overkill.
- Platform-specific `HostID()` function was added for the same reasons as above.
- Missing function placeholders were added to the `host_fallback.go`.
- Some parts of `host_solaris.go` were rewritten. `PlatformInformationWithContext()` now returns sane values.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -1,7 +1,11 @@
 package host
 
 import (
+	"context"
 	"encoding/json"
+	"os"
+	"runtime"
+	"time"
 
 	"github.com/shirou/gopsutil/internal/common"
 )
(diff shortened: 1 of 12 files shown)


Lomanic:
That looks really good, thanks a lot.

This is a big step for #761

shirou:
Great job! it works on my linux and FreeBSD-11.3.

So newly exported functions are 

- HostID()
- KernelArch()

Adding API don't break compatibility. Thank you! Can I merge?

boogie-byte:
Yes, I think it could be merged. However, I'd like to stress that I've only tested it on Linux myself. `go vet` runs ok for all platforms except openbsd, where it fails with build error somwhere in the `process` package (same behavior
in the master branch, though).

shirou:
Thank you for your awesome contribution!!

boogie-byte:
Glad to help. BTW, any plans on next tagged release which will include this commit?

PR #944, opened 2020-09-15
[cpu][linux] Fix #849 implement giampaolo/psutil#1727 and test Counts against lscpu

### LINKED ISSUE #849: cpu count not correct
**Describe the bug**
[A clear and concise description of what the bug is.]

**To Reproduce**
```go
package main

import (
  "fmt"
  "github.com/shirou/gopsutil/cpu"
)

func main() {
  info, err := cpu.Counts(false)
  if err != nil {

  }
  fmt.Println(info)

}

root@master-189:/ssd/tools# go run main.go 
16
root@master-189:/ssd/tools# cat /proc/cpuinfo |grep "physical id"|sort |uniq |wc -l
2

```

**Expected behavior**
[A clear and concise description of what you expected to happen.]

NAME="Ubuntu"
VERSION="14.04.5 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.5 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"

**Additional context**
mac os the cpu physical num is correct,but the linux is not correct

Lomanic:
[cpu][linux] Fix #849 implement giampaolo/psutil#1727 and test Counts against lscpu

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


Lomanic:
This was only tested against amd64, if people watching the repo and having a linux box at hand could test it with `go test github.com/shirou/gopsutil/cpu` with this patch applied and report back, this would be extremely appreciated. Especially on multi-socketed machines. Thanks in advance!

shirou:
I confirmed it works on my amd64 Linux. Thank you so much!

Lomanic:
Did you run the tests on a multi-socket machine?

shirou:
Sorry, Actually, no. Should I revert this?

Lomanic:
No, but I wanted more samples for the test I wrote that compares values from lscpu and gopsutil.

Lomanic:
BTW today I could finally confirm that the tests run successfully on an armv7l host. This PR fixed this in fact (cpu.Counts(false) would return 0 physical cores on arm without error), as I think the initial bug report in #849 conflated cpu sockets/packages and cpu physical cores.

shirou:
Oh, awesome! Thank you!

PR #945, opened 2020-09-19
[host][linux] Change to use unix.SysInfo_t to get Uptime in Linux

shirou:
According to [this](https://github.com/shirou/gopsutil/pull/857#issuecomment-629858607) and [this](https://github.com/shirou/gopsutil/pull/942#issuecomment-691185427) comments, change to use `unix.Sysinfo_t` in Linux.

Unfortunately, other platform such as FreeBSD do not include "Uptime" in Sysinfo. So this only for Linux.

### DIFF
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -70,11 +70,11 @@ func BootTimeWithContext(ctx context.Context) (uint64, error) {
 }
 
 func UptimeWithContext(ctx context.Context) (uint64, error) {
-	boot, err := BootTime()
-	if err != nil {
+	sysinfo := &unix.Sysinfo_t{}
+	if err := unix.Sysinfo(sysinfo); err != nil {
 		return 0, err
 	}
-	return timeSince(boot), nil
+	return uint64(sysinfo.Uptime), nil
 }
 
 func UsersWithContext(ctx context.Context) ([]UserStat, error) {


Lomanic:
Couldn't we base BootTime against Uptime like on the other platforms? This would improve performance I think.

shirou:
Other platform uses `time.Now().Unix()` via `timeSince()` function. From my understanding, `time.Now()` has a monotonic clock but monotonic clock can work only against time object which has a monotonic clock. if not, fall back to wall clock.

Other platform get bootime as just a uint64 which does not have a monotonic clock, so ret value of `timeSince()` is also not a monotonic. This means, other platform has a problem same as #837 (but not confirmed....).

At least FreeBSD, perhaps we'd better to implement `clock_gettime(CLOCK_UPTIME` like [w source code](https://svnweb.freebsd.org/base/release/12.1.0/usr.bin/w/w.c?view=markup#l499).

Lomanic:
In fact we're already using the same thing as `timeSince()` in  `host.BootTime()` on docker/lxc guests  https://github.com/shirou/gopsutil/blob/6c9f3bdcae3cd68d3e80eacf19eafe0efe0a4ea8/internal/common/common_linux.go#L102

For the time being I'm merging this, it's already an improvement. :+1:

PR #947, opened 2020-09-29
[process][darwin] Fix #925 properly and improve some tests

### LINKED ISSUE #925: Darwin unix.SYS___SYSCTL has been removed and causes gopsutil to break

Lomanic:
[process][darwin] Fix #925 properly and improve some tests

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -12,7 +12,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-	"unsafe"
 
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/internal/common"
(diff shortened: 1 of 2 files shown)


shirou:
sorry, I don't have a Darwin machine but I think the code LGTM

shirou (approved):

Issue #818, opened 2020-01-15
Combine Process .Ppid(), .Name(), NumThreads() for Windows


AtakanColak:
**Is your feature request related to a problem? Please describe.**

While trying to all of the information of all my processes in my Windows 10 PC using gopsutil.process, I observed about %40 CPU usage and ~38s runtime. After introducing go's concurrency and removing Cmdline() and IOCounters() calls which used most of the runtime, I was able to reduce it to about ~1.5s. However %40 CPU usage stayed. I did CPU profiling with pprof and found getFromSnapProcess was using the most resources. Right now it discards 2 of the 3 values (ppid, name and num threads) returned by a  windows.ProcessEntry32 object in each call, which is found by iterating through all of the processess in the PC. This causes a O((3 * N) ^ N) complexity while it is possible to iterate for once and keep the data for further calls. 

**Describe the solution you'd like**

 I wrote an external replacement function for getFromSnapProcess which keeps its data as a map with referencing: 
 
https://golang.hotexamples.com/examples/syscall/-/Process32First/golang-process32first-function-examples.html

I observed a drop in CPU from %40 to %2. Runtime also dropped from 1.4s to about 0.3s. However I don't have the time to prepare this for gopsutil at the moment. Therefore I wanted to add this as a feature request.

Lomanic:
> This causes a O(3 * N * N) complexity while it is possible to iterate for once and keep the data for further calls with just a O(N) complexity with only a slight increase in memory usage.

Solution to this is probably #286, but then we would have to implement an API for cache invalidation, which is not straightforward. 

#186 would also apply.

AtakanColak:
> Solution to this is probably #286, but then we would have to implement an API for cache invalidation, which is not straightforward.

How about a simple CreateTime() check for it instead? If it is created at the same time with the same pid then it is the same process. At least for Ppid and Name have to stay the same, only one that could differ is NumThreads. Perhaps this kind of logic could also be implemented for #286 where it would be more meaningful?

PR #948, opened 2020-10-01
Cache Process Parent ID in Windows

AtakanColak:
There is a `parent` field in `process.Process`, that is used as a buffer in `process_linux.go` but unused in `process_windows.go` code. As per #818,  I wanted to make a suggestion to utilize it to cache ppid value which is returned in any `getFromSnapProcess` call. 

This change will work only if we assume parent process ID won't change in Windows (and should it change, new value shouldn't be trusted).

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -240,10 +240,19 @@ func (p *Process) Ppid() (int32, error) {
 }
 
 func (p *Process) PpidWithContext(ctx context.Context) (int32, error) {
+	// if cached already, return from cache
+	if p.parent != 0 {
+		return p.parent, nil
+	}
+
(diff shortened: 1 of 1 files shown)


shirou:
It works well, and I can not find any situation Parent ID changes (I am not familiar with Windows though).
And I think a cache is not good for library, application side should cache, but we already accepted a cache at #942, So I merge this.

shirou:
Note: I noticed #951 after merge this PR. so this PR perhaps same thread safe issue. Need to check.

AtakanColak:
@shirou yeah, totally forgot about race condition regarding #951 as it was write-once, I'll fix it tomorrow.

PR #949, opened 2020-10-02
Add "vet" target to the Makefile

boogie-byte:
Making the "vet" target will run "go vet ./..." for each supported GOOS. Also, some of the warnings produced by go 1.15 vet were mitigated.


boogie-byte:
There still are two issues unfixed:

- Some big compile-time issue with "process" package in OpenBSD, which I'd prefer to fix separately (probably while DRYing the whole package).
- Possible misuse of `unsafe.Pointer` in the Windows process code (not my area of expertise, but at the first glance it looks like the `unsafe.Pointer` conversion is not needed there since `windows.Handle` is a `uintptr` type alias):
```
process/process_windows.go:698:48: possible misuse of unsafe.Pointer
process/process_windows.go:718:47: possible misuse of unsafe.Pointer
```

Once both these issues are resolved, `make vet` could be included in the CI script.

boogie-byte:
I've looked into the OpenBSD `go vet` issue, and it looks like it fails due to the `vet` tool limitation which prevents CGO-enabled source files from being included. Given the fact it is also skipped in the `build_test` target, I've removed it from the `vet` target as well.
As for the `process_windows.go` fix, I've removed the `unsafe.Pointer` type cast and tested the process suspension and resumption on a Windows box; everything works as expected.
Since all `go vet` warnings are now fixed, I took the liberty to substitute the `go build` step with `make vet` in the CI script. This greatly improves the check coverage without compiling the code. As a side effect this change required to update the golang version in CI to 1.14 since there is a RISC-V specific code in the `host` package, and the RISC-V support was introduced only in go 1.14.

shirou:
Thank you! I can confirm it fix the go vet results in go 1.15.2. And also Go 1.11 is deprecated, so now good timing to change 1.14.

PR #953, opened 2020-10-07
process: Use SendSignalWithContext where context is available

zeisss:
This changes calls to `Process#SendSignal()` to `Process#SendSignalWithContext()` where a context is available.

Note that this doesn't really change anything, since `SendSignalWithContext` actually ignores the context - maybe that can be fixed at some point too.


shirou:
Thank you for this PR. I'm sorry, but #966 includes this PR and with other refactors. So I close this. But I very appreciate to your work!

zeisss:
No problem, thanks for working on this :)

Issue #954, opened 2020-10-07
disk.Partitions on linux cannot identify bind mounts


tmm1:
Bind mounts, i.e. `mount --bind /mnt/overlay/etc/hostname /etc/hostname` cannot be distinguished in `disk.PartitionStat` entries:

```
{
"device": "/dev/sda7",
"mountpoint": "/mnt/overlay",
"fstype": "ext4",
"opts": "rw,relatime"
},
{
"device": "/dev/sda7",
"mountpoint": "/etc/hostname",
"fstype": "ext4",
"opts": "rw,relatime"
}
```

```
# cat /etc/mtab  | grep hostname
/dev/sda7 /etc/hostname ext4 rw,relatime 0 0

# mount | grep hostname
/dev/sda7 on /etc/hostname type ext4 (rw,relatime)

# cat /proc/self/mountinfo | grep hostname
121 19 8:7 /etc/hostname /etc/hostname rw,relatime shared:50 - ext4 /dev/sda7 rw

# findmnt | grep hostname
|-/etc/hostname                          /dev/sda7[/etc/hostname]                          ext4       rw,relatime
```

I propose that we detect these mounts and add `opts:"bind"` to the generated entry. WDYT?

See more discussion on https://unix.stackexchange.com/a/290441/199880

cc #664

tmm1:
from procfs docs (https://man7.org/linux/man-pages/man5/proc.5.html):

```
36 35 98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 /dev/root rw,errors=continue
(1)(2)(3)   (4)   (5)      (6)      (7)   (8) (9)   (10)         (11)

              (4)  root: the pathname of the directory in the filesystem
                   which forms the root of this mount.

              (5)  mount point: the pathname of the mount point relative to
                   the process's root directory.

              (6)  mount options: per-mount options (see mount(2)).
```

Currently we ignore `(4)` altogether. So maybe a simpler solution is to expose a `Root` inside `disk.PartitionStat`

PR #955, opened 2020-10-07
[disk][linux] flag bind mounts

tmm1:
possible solution to #954

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -283,6 +283,14 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 			mountPoint := fields[4]
 			mountOpts := fields[5]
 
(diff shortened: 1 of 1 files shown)


disk/disk_linux.go
```diff
@@ -283,6 +283,14 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 			mountPoint := fields[4]
 			mountOpts := fields[5]
 
+			if rootDir := fields[3]; rootDir != "" && rootDir != "/" {
+				if len(mountOpts) == 0 {
```
shirou, line 287: 👍 mountOpts is just a string, so this is fastest way (mountOpts should be a list, not string, though).

shirou:
Good catch! I can confirm it can detect and add 'bind' to Opts in Ubuntu 20.04. Thank you.

PR #956, opened 2020-10-07
[disk][darwin] fix build with latest x/sys

tmm1:
fixes build on darwin with `golang.org/x/sys v0.0.0-20201007165808-a893ed343c85`

```
../../go/pkg/mod/github.com/fancybits/gopsutil@v2.16.11-0.20201007180101-23d63601d124+incompatible/disk/disk_darwin.go:67:51: cannot use stat.Mntfromname[:] (type []byte) as type []int8 in argument to common.IntToString
../../go/pkg/mod/github.com/fancybits/gopsutil@v2.16.11-0.20201007180101-23d63601d124+incompatible/disk/disk_darwin.go:68:49: cannot use stat.Mntonname[:] (type []byte) as type []int8 in argument to common.IntToString
../../go/pkg/mod/github.com/fancybits/gopsutil@v2.16.11-0.20201007180101-23d63601d124+incompatible/disk/disk_darwin.go:69:50: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
../../go/pkg/mod/github.com/fancybits/gopsutil@v2.16.11-0.20201007180101-23d63601d124+incompatible/disk/disk_darwin.go:85:43: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
```

### DIFF
--- a/disk/disk_darwin.go
+++ b/disk/disk_darwin.go
@@ -64,9 +64,9 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 			opts += ",nodev"
 		}
 		d := PartitionStat{
-			Device:     common.IntToString(stat.Mntfromname[:]),
-			Mountpoint: common.IntToString(stat.Mntonname[:]),
-			Fstype:     common.IntToString(stat.Fstypename[:]),
+			Device:     common.ByteToString(stat.Mntfromname[:]),
+			Mountpoint: common.ByteToString(stat.Mntonname[:]),
+			Fstype:     common.ByteToString(stat.Fstypename[:]),
 			Opts:       opts,
 		}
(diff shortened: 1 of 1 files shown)


audiolion:
Created an issue #958 to track this

Lomanic:
I confirmed that it now builds with `GOOS=darwin go build github.com/shirou/gopsutil/...` after updating x/sys. That's the same fix as freebsd in #895 so should be good (didn't test). I think we are probably going to see similar breaking changes from x/sys with openbsd.

Could you update Gopkg.lock with dep? See https://golang.github.io/dep/docs/daily-dep.html#updating-dependencies

Lomanic:
Superseded by #970.

PR #957, opened 2020-10-07
[disk][linux] Add `Source` field to PartitionStat for original symbolic link path

haojhcwa:
Currently in order to find the correct device name, anything mounted
from symbolic link under `/dev/mapper` will be evaluated to the real
device it points to. See [#686](https://github.com/shirou/gopsutil/pull/686)

However, the label under `/dev/mapper` is often much more infomative,
and is a very useful piece of informtion to keep. Though this info could
be mapped back, but it is much easier to just record it here.

In this change, I've add a new field named source and would only be
populated when the symlink evaluation had been carried out.

The json tag of the new source field is created with "omitempty" option,
so it won't create any behavior difference for code that does not make
use of it.

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -23,6 +23,7 @@ type UsageStat struct {
 
 type PartitionStat struct {
 	Device     string `json:"device"`
+	Source     string `json:"source,omitempty"`
 	Mountpoint string `json:"mountpoint"`
 	Fstype     string `json:"fstype"`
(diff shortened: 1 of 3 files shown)


Lomanic:
That's yet another linux-only feature when gopsutil aims to be the most platform-agnostic. :-1: from me.

shirou:
agree. sorry for closing but thank you for your contribution.

Issue #961, opened 2020-10-12
Race Condition for Ppid Cache in Windows


AtakanColak:
PR #942 merge had a race condition reported in PR #951. Similar race condition exists with PR #948. I'm working on a fix.

AtakanColak:
Closed as PR is merged.

PR #962, opened 2020-10-12
Fix Windows Ppid Cache Race Condition

AtakanColak:
As per #961, a RWMutex to Process struct and thread-safe ppid access functions are added which are now used for windows but virtually can be used for other OSes in the future.  

Another solution would be to not cache at all, but this solution reduces overall system calls and in windows it makes quite the difference. 

Please do review. @Lomanic @shirou

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"runtime"
 	"sort"
+	"sync"
 	"syscall"
 	"time"
 
@@ -26,6 +27,7 @@ type Process struct {
 	name           string
 	status         string
 	parent         int32
+	parentMutex    *sync.RWMutex // for windows ppid cache
 	numCtxSwitches *NumCtxSwitchesStat
 	uids           []int32
 	gids           []int32
@@ -167,7 +169,10 @@ func NewProcess(pid int32) (*Process, error) {
 }
 
 func NewProcessWithContext(ctx context.Context, pid int32) (*Process, error) {
-	p := &Process{Pid: pid}
+	p := &Process{
+		Pid:         pid,
+		parentMutex: new(sync.RWMutex),
+	}
 
 	exists, err := PidExistsWithContext(ctx, pid)
 	if err != nil {
--- a/process/process_race_test.go
+++ b/process/process_race_test.go
@@ -1,0 +1,30 @@
+// +build race
+
+package process
+
+import (
+	"sync"
+	"testing"
+)
+
+func Test_Process_Ppid_Race(t *testing.T) {
+	wg := sync.WaitGroup{}
+	testCount := 10
+	p := testGetProcess()
+	wg.Add(testCount)
+	for i := 0; i < testCount; i++ {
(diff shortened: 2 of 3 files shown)


AtakanColak:
go vet still fails, looking into it

AtakanColak:
go vet check passes after changing the mutex to a pointer that is initialized with NewProcess call

AtakanColak:
Squashed commits, waiting for review.

Lomanic (changes requested):
Thanks for PR, I didn't yet have a chance to test it. See my comment.

shirou:
Sorry, #966 makes conflicts. But If you follow [the comment](https://github.com/shirou/gopsutil/pull/962/files#r503256996) and make the changes, this conflicts may resolved,

AtakanColak:
@Lomanic thank you for the doc link, seeing a caching in windows ppid showed me that I was right to start this feature beforehand. Resolved it by carrying the methods to process_windows.go

@shirou resolved the conflicts.

Even though my tests passed, please do test before merge.

shirou:
Thank you so much! I can reproduce race condition by using this test and this PR can fix.

```
% go test --race -v -run Test_Process_Ppid
```

```
func Test_Process_Ppid(t *testing.T) {
	wg := sync.WaitGroup{}
	testCount := 10
	p := testGetProcess()
	wg.Add(testCount)
	for i := 0; i < testCount; i++ {
		go func(j int) {
			ppid, err := p.Ppid()
			wg.Done()
			skipIfNotImplementedErr(t, err)
			if err != nil {
				t.Errorf("Ppid() failed, %v", err)
			}

			if j == 9 {
				t.Logf("Ppid(): %d", ppid)
			}
		}(i)
	}
	wg.Wait()
}
```

Could you replace the current `Test_Process_Ppid()` by this code, confirm, and commit again? Thanks!

AtakanColak:
@shirou test passed on my windows. Added to the commit.

Lomanic:
Problem with your test @shirou is that it doesn't test the same thing anymore (getting a non-zero value, and the same value as `os.Getppid()`). I would restore the previous one and put that one new test to a dedicated test like in [https://github.com/shirou/gopsutil/blob/master/disk/disk_test.go#L70-L85](https://github.com/shirou/gopsutil/blob/3202231bcdbd3a6d236dbcc3ec92a96ddfa7168f/disk/disk_test.go#L70-L85).

And I would also suggest to be explicit that this test is made to detect race conditions (in its name). This test could even be put in a new `process/process_race_test.go` with `// +build race` flag on top so it only runs when the race detector is enabled.

shirou:
Ah, yes, totally agree to @Lomanic . 

I'm really sorry but could you revert that commit and create a new `process/process_race_test.go` file? > @AtakanColak .

AtakanColak:
@Lomanic @shirou committed with a process_race_test.go instead.

AtakanColak:
@Lomanic could you review and merge if deemed sufficient please.

Issue #913, opened 2020-08-03
Calling the new process.Groups() fails on macOS


kunzimariano:
**Describe the bug**
Calling the new method `process.Groups()` fails on macOS Catalina (10.15.6).
Instead I get: `runtime error: index out of range [16] with length 16`

**To Reproduce**
```go
p, err := process.NewProcess(pID)
if err != nil {
	panic(err)
}

sgs, err := p.Groups()
if err != nil {
	panic(err)
}

```

**Expected behavior**
To get a list of supplementary groups. 

**Environment (please complete the following information):**
 - [ ] Windows: [paste the result of `ver`]
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [x] Mac OS: [
ProductName:	Mac OS X
ProductVersion:	10.15.6
BuildVersion:	19G73 
Darwin mbp.local 19.6.0 Darwin Kernel Version 19.6.0: Sun Jul  5 00:43:10 PDT 2020; root:xnu-6153.141.1~9/RELEASE_X86_64 x86_64
]
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**
The problem seems to be the usage of `k.Eproc.Ucred.Ngroups` as the end of the iteration instead of the length of the array being iterated  `(k.Eproc.Ucred.Groups)`.

https://github.com/shirou/gopsutil/blob/master/process/process_darwin.go#L279

```
groups := make([]int32, k.Eproc.Ucred.Ngroups)
for i := int16(0); i < k.Eproc.Ucred.Ngroups; i++ {
	groups[i] = int32(k.Eproc.Ucred.Groups[i])
}
```

I noticed that `k.Eproc.Ucred.Ngroups` is also being use by `Gids()` as the effective group. 

This "dual meaning" confuses me a little, specially if it is based on this: https://unix.superglobalmegacorp.com/Net2/newsrc/sys/ucred.h.html

Leaving that aside, if iterating `k.Eproc.Ucred.Groups` directly is the proper fix I'm happy to submit a PR.

Lomanic:
Yes I could confirm that running tests on darwin, this was obviously never tested in  #775 on this platform as it fails systematically (openbsd is probably not affected byt freebsd yes surely, I have some issue on my VM though). Yet another argument for running tests on all platforms automatically (#734).

[...]

PR #964, opened 2020-10-12
Various fixes for OpenBSD

boogie-byte:
* Use Host.HostID() in TestHostGuid
* Use common.ErrNotImplementedError in net/net_openbsd.go
* Check if CPU metrics are not null before assertion in CPU.Times test
* Fix return type in Process.GroupsWithContext()

### DIFF
--- a/process/process_openbsd.go
+++ b/process/process_openbsd.go
@@ -227,7 +227,12 @@ func (p *Process) GroupsWithContext(ctx context.Context) ([]int32, error) {
 		return nil, err
 	}
 
-	return k.Groups, nil
+	groups := make([]int32, k.Ngroups)
+	for i := int16(0); i < k.Ngroups; i++ {
+		groups[i] = int32(k.Groups[i])
+	}
+
+	return groups, nil
 }
 func (p *Process) Terminal() (string, error) {
 	return p.TerminalWithContext(context.Background())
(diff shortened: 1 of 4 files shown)


process/process_openbsd.go
```diff
@@ -227,7 +227,12 @@ func (p *Process) GroupsWithContext(ctx context.Context) ([]int32, error) {
 		return nil, err
 	}
 
-	return k.Groups, nil
+	groups := make([]int32, k.Ngroups)
+	for i := int16(0); i < k.Ngroups; i++ {
```
Lomanic, line 231: Same concerns as in https://github.com/shirou/gopsutil/pull/965#discussion_r503489889

boogie-byte: I've just copied a FreeBSD implementation, TBH, and it looks right to me. As for #913 , the darwin code looks valid too, no idea why there is a indexing error. Using the `len(k.Groups)` won't help since it always has a fixed length (hence the `k.Ngroups`: it indicates the amount of meaningful `k.Groups` elements).

shirou:
Thank you for the contribution. But it is conflicted and I want to some checks, so I opened #980. It's all thanks to you.

(I found process.Processes seems not work in FreeBSD because other than k.Ngroups. So I will fix it in another PR)

PR #965, opened 2020-10-12
[WIP] Remove CGO bits from process/process_openbsd.go

boogie-byte:
Prior to this commit CGO was used in OpenBSD implementation of
Process.CmdlineSliceWithContext() for parsing the "kern.proc.args"
sysctl output. It requires some pointer arithmetics and raw pointer
dereferencing.

Having CGO in the "process" module prevents it from being go vet'ted
on any platform other than OpenBSD. In order to overcome this
limitation, the sysctl output parsing was reimplemented without raw
pointer deferencing. The resulting code might be slightly slower
than the original one, but it is cleaner and safer.

Since this fix allows go vet with GOOS=openbsd to run without any
issues on all platforms, openbsd entries were also added to the "vet"
Makefile target.


boogie-byte:
Actually, I'd like to try an alternative approach by using `kvm(3)` (not _that_ KVM, but BSD Kernel Data Access	Library) for all BSD variants we have, so I'd like to put this PR on hold for now.

Lomanic:
As-is this is really great work either way, thanks a lot for submitting this PR. Two weeks ago I was trying to parse kern.proc.args output on darwin and couldn't find how to do so, I'm really curious where you got this info. 

I think you should be able to populate `buf` with `buf, err := unix.SysctlRaw("kern.procargs", p.Pid)` as it was returning promising results on darwin during my tests (hoping there's no regression for elevated processes as I remember things like that during my numerous tests).

process/process_openbsd.go
```diff
@@ -227,7 +258,12 @@ func (p *Process) GroupsWithContext(ctx context.Context) ([]int32, error) {
 		return nil, err
 	}
 
-	return k.Groups, nil
+	groups := make([]int32, k.Ngroups)
+	for i := int16(0); i < k.Ngroups; i++ {
```
Lomanic, line 262: We already have issue #913 on darwin where Ngroups is bigger than the length of Groups. Wouldn't we also have this issue here?

boogie-byte:
> I'm really curious where you got this info.

Man pages and source code, nothing more :)

> I think you should be able to populate `buf` with `buf, err := unix.SysctlRaw("kern.procargs", p.Pid)` as it was returning promising results on darwin

Unfortunately it won't work properly for this exact CTL, since the OpenBSD implementation of `unix.SysctlRaw()` uses a [static map](https://github.com/golang/sys/blob/dfb3f7c4e6342ab64fe0e22bde473c74a5699721/unix/zsysctl_openbsd_amd64.go#L13) to convert CTL names to MIBs, which has a lot of stuff missing, including `kern.procargs`. Other implementations use more consistent ways to do that (e.g. FreeBSD has a "magic CTL" for that).
Also, please keep in mind that the result buffer format is not standardized across *BSDs. This one is exclusive for OpenBSD and won't work for Darwin.
`kvm_*` functions provide much cleaner API than `sysctl`, and I'm pretty sure we'll benefit from utilizing it for process info gathering. I'll come up with a PoC this week, fingers crossed.

process/process_openbsd.go
boogie-byte: This fix got here by accident from a neighboring branch. Removing it from here, since it belongs to #964

danieljakots:
Hi,

Any update on this? @boogie-byte are you still working on it?

Cheers,
Daniel

ivandeex:
https://github.com/rclone/rclone/pull/5402 is waiting for this PR

boogie-byte:
> Hi,
> 
> Any update on this? @boogie-byte are you still working on it?
> 
> Cheers,
> Daniel

Hey Daniel,

I had to put this aside for a while and it became abandoned eventually, *sigh*. I'll try to get back to it this weekend, so I hope I can provide you with an update in a couple of days.

shirou:
Dear @boogie-byte, thank you for your great work. But unfortunately, it seems this PR staled again. I am very sorry, but please allow me to close this PR and adopt a new one(#1087) . We are very grateful for your initial contribution. Thank you so much.

boogie-byte:
Hey @shirou,

You're being too kind. Unfortunately, I've had little to no time to work on open-source projects lately. I hope I'll be more useful in the future :)

PR #966, opened 2020-10-12
Refactor "process" package

boogie-byte:
* All context-less wrapping functions (the ones without WithContext
suffix) were moved into process.go since they all are the same.
* Call context is now passed to all underlying functions in
*WithContext() functions.
* All common *BSD bits were moved to process_bsd.go.
* Process.Tgid() method lacked a WithContext counterpart, so
Process.TgidWithContext() was added for uniformity.
* NewProcessWithContext() function was added since NewProcess() is
used a lot throughout the module, and there is no way to pass a
context to it.

This is a part of #761 effort.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -6,11 +6,13 @@ import (
 	"errors"
 	"runtime"
 	"sort"
+	"syscall"
 	"time"
 
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/internal/common"
 	"github.com/shirou/gopsutil/mem"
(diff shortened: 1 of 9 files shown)


shirou:
I found this PR changes to return nil instead of empty value when returning error on various functions like

```
func (p *Process) GidsWithContext(ctx context.Context) ([]int32, error) {
	return []int32{}, common.ErrNotImplementedError
}
```

This may break application code. but checking error value at first is a golang-way,
so I think this is not become a problem on application.

I merge this. Awesome work, Thank you!

PR #972, opened 2020-10-15
[cpu][linux] Return err if reading cpuinfo fails in Info()

### LINKED ISSUE #971: panic: runtime error: index out of range [0] with length 0 (cpu.Info, cpu_info_linux.go)
**Describe the bug**
panic: runtime error: index out of range [0] with length 0

func cpu.Info() ([]InfoStat, error) is returning nil, nil which doesn't allow the caller to guard access to the returned slice.

My program calling cpu.Info is a linux exe running via a systemd unit. Moreover, it starts automatically upon system boot. There appears to be a race condition where `/proc/cpuinfo` isn't yet populated (where it's assumed to always be). I could potentially delay the start of the daemon, but this seems like a issue related to error checking within this package.

For example:
```
$ vim cpu/cpu_linux.go
...
func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
	filename := common.HostProc("cpuinfo")
	lines, _ := common.ReadLines(filename) // Possible bug, error isn't checked here.
       ...
       return ret, err // Maybe before return check if ret slice is nil and return a non-nil error for caller.
```

**To Reproduce**
```go
infos, err := cpu.Info()
if err != nil {
    return
}
if infos == nil { // Sometimes this happens, meaning the returned error isn't getting set.
```

**Expected behavior**
func cpu.Info() ([]InfoStat, error) should return a non-nil error when []InfoStat is nil.

 - [x ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]

**Additional context**
NAME=Fedora
VERSION="32 (Workstation Edition)"
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
PRETTY_NAME="Fedora 32 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:32"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f32/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=32
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=32
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation

Linux shakti 5.7.9-200.fc32.x86_64 #1 SMP Fri Jul 17 16:23:37 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Lomanic:
Fixes #971

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -114,12 +114,14 @@ func Info() ([]InfoStat, error) {
 }
 
 func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
-	filename := common.HostProc("cpuinfo")
-	lines, _ := common.ReadLines(filename)
-
 	var ret []InfoStat
-	var processorName string
+	filename := common.HostProc("cpuinfo")
+	lines, err := common.ReadLines(filename)
+	if err != nil {
+		return ret, err
+	}
 
+	var processorName string
 	c := InfoStat{CPU: -1, Cores: 1}
 	for _, line := range lines {
 		fields := strings.Split(line, ":")


Lomanic:
@fugolmere would you please confirm this fixes your issue at boot time?

fugolmere:
First off thank you @Lomanic for addressing this so quickly. Secondly, I apologize but just realized I was running a very old version of the code. 
```
panic: runtime error: index out of range [0] with length 0
goroutine 1 [running]:
github.com/shirou/gopsutil/cpu.finishCPUInfo(0xc00029ecf8, 0x9, 0xda3fa2)
/home/vagrant/go/workspace/pkg/mod/github.com/shirou/gopsutil@v2.16.13-0.20170804030934-3aa2ffab12a1+incompatible/cpu/cpu_linux.go:88 +0x1c3
```

The stack trace show the panic caused by the following:
```
82     lines, err = common.ReadLines(sysCPUPath(c.CPU, "cpufreq/cpuinfo_max_freq"))
...
85     if err != nil {
86         return nil
87     }
88     value, err = strconv.ParseFloat(lines[0], 64)
```

when compared to the latest code:
```
88     lines, err = common.ReadLines(sysCPUPath(c.CPU, "cpufreq/cpuinfo_max_freq"))
...
91     if err != nil || len(lines) == 0 {
92         return nil
93     }
94     value, err = strconv.ParseFloat(lines[0], 64)

```
The old code didn't have the `len(lines) == 0` check and thus was likely causing a panic when accessing `lines[0]`.

Considering how a length check was added in a more recent commit, it makes me assume that there are situations where `common.ReadLines` could return `lines == nil && err == nil`. I suggest modifying the merge commit such that it also addressed the underlying problem in `common.ReadLinesOffsetN`. For example:

https://github.com/shirou/gopsutil/blob/master/internal/common/common.go#L119
```
...
if ret == nil {
    return ret, errors.New("nothing to read")
}
return ret, nil
```
Something similar to the above should make it such that checking the length (in addition to an error) would no longer be necessary for every call of `common.ReadLine`.

Lomanic:
I fear side effects as `common.ReadLines` and `common.ReadLinesOffsetN` are called in many places in gopsutil. This needs to be carefully thought out.

fugolmere:
Yeah that's understandable. The risk to reward ratio for such a change appears rather low and likely not worth it. In the least, this issue can service to document that there is a possible rare edge case scenario which may result in a slice access violation. If other's have similar encounters, then this issue can be referred back to.

PR #979, opened 2020-10-24
process: unbreak build on OpenBSD

ajacoutot:
Match FreeBSD code to prevent:
process_openbsd.go:230:10: cannot use k.Groups (type [16]uint32) as type []int32 in return argument


shirou:
Thank you so much!

PR #983, opened 2020-11-02
[cpu][windows] fix CPU struct name on windows.

### LINKED ISSUE #982: cpu.Info error on v3 for windows 10
**Describe the bug**
calling the cpu.Info function when the imported package is "github.com/shirou/gopsutil/v3/cpu" returns "Exception occurred. (Invalid class )" error while it worked with package "github.com/shirou/gopsutil/cpu".
I have looked into the source code (common_wimndows.go, cpu_windows.go) for both packages but cannot find any difference in those files that could cause the error on v3.

**To Reproduce**
```go
c, err := cpu.Info()
	if err != nil {
		fmt.Println("Could not retrieve CPU details.")
		log.Fatal(err)
	}
```

 Microsoft Windows [Version 10.0.18363.1139]  64bit
 CPU Model: Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz

shirou:
fixes #982 

The struct name requires `_`.

### DIFF
--- a/v3/cpu/cpu_windows.go
+++ b/v3/cpu/cpu_windows.go
@@ -17,7 +17,7 @@ var (
 	procGetNativeSystemInfo     = common.Modkernel32.NewProc("GetNativeSystemInfo")
 )
 
-type win32Processor struct {
+type win32_Processor struct {
 	LoadPercentage            *uint16
 	Family                    uint16
 	Manufacturer              string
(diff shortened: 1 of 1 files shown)


Lomanic:
Verified with `go test -v github.com/shirou/gopsutil/v3/cpu` (it was failing on windows), thanks @shirou

PR #984, opened 2020-11-03
[net][windows] Fix getting udp6 ports

MashaSamoylova:
It's currently impossible to list UDP6 ports on Windows due to typo


shirou:
Wow, good catch! Thank you, but could you fix v3/net/net_windows.go as well?

MashaSamoylova:
Thank you for your quick reply!

I've fixed `v3/net/net_windows.go` in the same commit

PR #985, opened 2020-11-04
Fix labeler CI "HttpError: Resource not accessible by integration" error

Lomanic:
See https://github.com/actions/labeler/issues/12#issuecomment-670967607


shirou:
oops. Thank you very much!

PR #988, opened 2020-11-05
[v3][cpu][windows] Remove exported windows-specific dead code

Lomanic:
Users wanting to use this Windows-specific feature can still use github.com/StackExchange/wmi by themselves.


shirou (approved):
LGTM, and thank you for fixing code and also document!

PR #989, opened 2020-11-05
[ci] Implement automatic tests on GH Actions on linux, windows and darwin

Lomanic:
Based off https://github.com/mvdan/github-actions-golang/blob/3710bfcc3e71fc692118d0d5a42df57481095db7/.github/workflows/test.yml

Linked to #734


Lomanic:
Maybe we should remove circleci tests, they really want me to create an account on their platform just to see logs for failing jobs, which is really annoying, and they are now just redundant with these Github Actions.

v3/process/process_darwin.go
```diff
@@ -212,17 +210,18 @@ func (p *Process) GidsWithContext(ctx context.Context) ([]int32, error) {
 }
 
 func (p *Process) GroupsWithContext(ctx context.Context) ([]int32, error) {
-	k, err := p.getKProc()
-	if err != nil {
-		return nil, err
-	}
+	return nil, common.ErrNotImplementedError
```
shirou, line 213: Sorry, I can not remember but is this `Groups` not working?

shirou (commented):
It looks awesome! Thank you so much! I commented to just one method. if ok, let's merge and remove CircleCI!

v3/process/process_darwin.go
Lomanic: I don't think it ever worked, see #913.

shirou: thanks!

shirou (approved):
LGTM!

PR #991, opened 2020-11-06
[cpu][linux] Get logical cores in Counts() with non-deprecated kernel API first

Lomanic:
Implementing https://github.com/giampaolo/psutil/pull/1727#issuecomment-707624964 and https://github.com/giampaolo/psutil/commit/bfae1fc4a371c9e08f2c3f5053a80542e43d18f7


shirou (approved):
LGTM and I confirmed it works on my Linux. Thank you!

PR #994, opened 2020-11-09
Detect wsl virtualization same way as detect-virt

afbjorklund:
It is the Windows Subsystem for Linux, version 2

Adopted from https://github.com/microsoft/WSL/issues/423 and systemd

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -178,6 +178,18 @@ func VirtualizationWithContext(ctx context.Context) (string, string, error) {
 		}
 	}
 
+	filename = HostProc("sys/kernel/osrelease")
+	if PathExists(filename) {
+		contents, err := ReadLines(filename)
+		if err == nil {
+			if StringsContains(contents, "Microsoft") ||
+				StringsContains(contents, "WSL") {
+				system = "wsl"
+				role = "guest"
+			}
+		}
+	}
+
 	filename = HostProc("bus/pci/devices")
 	if PathExists(filename) {
 		contents, err := ReadLines(filename)


shirou:
on my and my friend environment, the `/proc/sys/kernel/osrelease` is `4.19.104-microsoft-standard` at this timing. so, `Microsoft` and `WSL` are included.  Could you show us osrelease in your environment?

afbjorklund:
I didn't have any environment, and noticed that the proposed changed had been reverted. Situation now is a bit problematic...

https://github.com/microsoft/WSL/issues/423#issuecomment-679190758

i.e. "microsoft" and "standard" are not enough, and "WSL" was not accepted (since it broke the hardcoded "microsoft-standard")

So it needs more code...

afbjorklund:
Just that we got some bug reports (for minikube) labelled as "Ubuntu", but they really was running on Windows under WSL2

shirou:
Cloud you change to detect method to [following ](https://github.com/microsoft/WSL/issues/423#issuecomment-679190758) ? 

1. check `Microsoft` or `WSL` in `/proc/sys/kernel/osrelease` -> if yes, it's WSL
2. check `microsoft` in the osrelease file -> if yes, go to 3, if no, not WSL
3. check `/run/WSL` directory exists -> if yes, it's WSL. if no, go to 4.
4. check `wslpath` is exists in PATH by using `exec.LookPath` -> if yes, its' WSL, if no, not WSL.

afbjorklund:
I'll give it another go when I have access to Windows again.

The original idea was just to copy `systemd-detect-virt`

afbjorklund:
There is also https://github.com/giampaolo/psutil/issues/1251

PR #996, opened 2020-11-13
WIP: feat: Split cGroup stats into its own pkg

djgilcrease:
The current docker package requires the docker binary, and the container ID which makes sense if you are running gopsutil outside the container. If however you are running it within the container the `mem.VirtualMemory()` call still returns the host info not what the container is limited to. Same with the CPU info.

This feature / fix makes it so you can do something like
```go
type cpuTimesStat struct {
	cpu.TimesStat
}
func newCpuTimesStat() *cpuTimesStat {
	virtSystem, _, _:= host.VirtualizationWithContext(context.Background())
	if err != nil {
		log.WithError(err).Error("no virtualization")
	}
	if virtSystem == "docker" {
		ct, err := cgroup.CgroupCPU()
		if err != nil {
			log.WithError(err).Errorf("%+#v", ct)
		}
		return &cpuTimesStat{TimesStat: ct.TimesStat}
	} else {
		ct, err := cpu.Times(false)
		if err != nil {
			log.WithError(err).Errorf("%+#v", ct)
		}
		return &cpuTimesStat{TimesStat: ct[0]}
	}
	return nil
}
```

This is a Work In Progress mostly because I am not sure this is the best way to do this.


Lomanic:
I don't know, I remember (and I agree with that) that @shirou wanted to remove the `docker` package in v3, so it's not to split its API to a new package. It's yet another linux-only feature when gopsutil aims to provide the same APIs and features over its supported OSes.

djgilcrease:
Ya would be great to do that if it was correct when being called within a cgroup, but it is not and I do not want to change the current memory, cpu, and disk calls to be able to handle pulling stats from a cgroup just the minimum changes I need for my use case. So splitting cgroup out from docker is the fastest way to do so that I can see.

shirou:
First of all, thank you for your contribution. This seems good to some users. However, this `docker` module is Linux specific. When I created `docker` module, there are nothing about it so I have added. But recently there are already some good libraries to use Docker and cgroups. So I don't want to add Docker/cgroups features anymore. I have tried to remove docker at v3 migration, but there are already many breaking changes, so I didn't change it this time.

PR #997, opened 2020-11-19
remove circleci testing

shirou:
By #989, we have moved to use GitHub actions.

- remove circleci
- change to use go 1.13 for testing. (internal/common/common_windows.go uses '%w' )
- add badge


shirou:
the fail is CircleCI. I already unfollow, so all tests are passed.

Lomanic:
> change to use go 1.13 for testing. (internal/common/common_windows.go uses '%w' )

I don't see any `%w` in this file and the tests run fine on windows on this version, I don't understand this change

shirou:
Thank you for the comment!

You can see [here](https://github.com/shirou/gopsutil/blame/master/internal/common/common_windows.go#L178) which is introduced from #977, and failed at [github actions](https://github.com/shirou/gopsutil/runs/1424460760?check_suite_focus=true). we can change the code, but '%w' is introduced from [go 1.13](https://golang.org/doc/go1.13#error_wrapping) and go 1.12 is already not supported. so I have changed the version.

Lomanic:
[go 1.12 is not that old (2019-02-25)](https://blog.golang.org/go1.12) and the last version to support winxp/winvista (see [go 1.13 release notes](https://golang.org/doc/go1.13#windows)), I find it quite sad to drop this just for this syntactic sugar.

shirou:
I am sorry to drop that version support. But [golang itself supports only two newer major versions](https://golang.org/doc/devel/release.html), 1.12 is already deprecated, actually 1.13, too. Then, I think it is a good timing to reduce supported versions. But if you want to revert this, I can do it.

PR #999, opened 2020-11-21
cpufreq for apple silicon macs

shoenig:
Previously on darwin, it was safe to assume that `hw.cpufrequency`
would be defined in sysctl. On the new M1 macs instead use the base
frequency `hw.tbfrequency` and `kern.clockrate[hz]`. The computed
value appears to be 2.4 GHz, which is I believe is the correct base
clock.

Before:
```
 m1  cpu  master % go test
--- FAIL: TestCpuInfo (0.00s)
    cpu_test.go:117: error no such file or directory
    cpu_test.go:120: could not get CPU Info
FAIL
exit status 1
FAIL	github.com/shirou/gopsutil/v3/cpu	0.371s
```

After:
```
 m1  cpu  m1freq % go test
PASS
ok  	github.com/shirou/gopsutil/v3/cpu	0.248s
```

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -93,13 +93,33 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 
 	// Use the rated frequency of the CPU. This is a static value and does not
 	// account for low power or Turbo Boost modes.
-	cpuFrequency, err := unix.SysctlUint64("hw.cpufrequency")
+	cpuFrequency, err := getFrequency()
 	if err != nil {
 		return ret, err
 	}
 	c.Mhz = float64(cpuFrequency) / 1000000.0
 
 	return append(ret, c), nil
+}
+
+func getFrequency() (uint64, error) {
+	// This works for Intel macs.
+	if cpuFrequency, err := unix.SysctlUint64("hw.cpufrequency"); err == nil {
+		return cpuFrequency, nil
+	}
+
+	// On Apple Silicon fallback to hw.tbfrequency and kern.clockrate[hz]
+	tbFreq, err := unix.SysctlUint64("hw.tbfrequency")
+	if err != nil {
+		return 0, err
+	}
+
+	kernClockrate, err := unix.SysctlClockinfo("kern.clockrate")
+	if err != nil {
+		return 0, err
+	}
+
+	return uint64(kernClockrate.Hz) * tbFreq, nil
 }
 
 func CountsWithContext(ctx context.Context, logical bool) (int, error) {
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -93,33 +93,26 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 
 	// Use the rated frequency of the CPU. This is a static value and does not
 	// account for low power or Turbo Boost modes.
-	cpuFrequency, err := getFrequency()
-	if err != nil {
-		return ret, err
-	}
-	c.Mhz = float64(cpuFrequency) / 1000000.0
 
-	return append(ret, c), nil
-}
-
-func getFrequency() (uint64, error) {
 	// This works for Intel macs.
 	if cpuFrequency, err := unix.SysctlUint64("hw.cpufrequency"); err == nil {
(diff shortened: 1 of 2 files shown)


cpu/cpu_darwin.go
```diff
@@ -93,7 +93,7 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 
 	// Use the rated frequency of the CPU. This is a static value and does not
 	// account for low power or Turbo Boost modes.
-	cpuFrequency, err := unix.SysctlUint64("hw.cpufrequency")
+	cpuFrequency, err := getFrequency()
```
Lomanic, line 96: Would you please inline this function? It's only called in a single place (same thing in v3 also).

Lomanic (changes requested):
Thanks for this PR, I didn't test it yet (hoping `hw.tbfrequency` and `kern.clockrate` are also available on amd64 as I don't have access to any Apple ARM machine).

tmm1:
> hoping `hw.tbfrequency` and `kern.clockrate` are also available on amd64

Yes they appear to be.

Lomanic:
Having tested with the following small script on an amd64 hackintosh VM, I'm just confused by the result

```go
package main

import (
        "fmt"

        "golang.org/x/sys/unix"
)

func getFrequencyIntel() (uint64, error) {
        // This works for Intel macs.
        return unix.SysctlUint64("hw.cpufrequency")
}

func getFrequencyArm() (uint64, error) {
        // On Apple Silicon fallback to hw.tbfrequency and kern.clockrate[hz]
        tbFreq, err := unix.SysctlUint64("hw.tbfrequency")
        if err != nil {
                return 0, err
        }

        kernClockrate, err := unix.SysctlClockinfo("kern.clockrate")
        if err != nil {
                return 0, err
        }

        return uint64(kernClockrate.Hz) * tbFreq, nil
}

func main() {
        fmt.Println(getFrequencyIntel())
        fmt.Println(getFrequencyArm())
}
```
As it returns 
```
2500000000 <nil>                                                                                                                                                                                                                                                               
100000000000 <nil>
```
Also, is python psutil also failing with `psutil.cpu_freq()` on Apple M1?

shoenig:
Yeah @Lomanic, I'm still trying to dig through and make sense of what Apple is doing here. There is probably a more correct way to get this value that is consistent across Intel/ARM macs but maybe only with cgo. If anyone else wants to play along, https://github.com/knightsc/darwin-xnu/blob/master/pexpert/arm/pe_identify_machine.c#L45

cpu/cpu_darwin.go
shoenig: I've made this change; IMHO it was much more readable as a function.

shoenig:
> Also, is python psutil also failing with psutil.cpu_freq() on Apple M1?

I don't have Python setup, but I would suspect it fails since they use the absent sysctl values.
https://github.com/giampaolo/psutil/blob/master/psutil/_psutil_osx.c#L693

Lomanic:
Darwin kernel is under the [Apple Public Source License](https://en.wikipedia.org/wiki/Apple_Public_Source_License) which I understand is not compatible with ours so linking against is not an option. Porting its code (in Go?) looks heavily involved.

Reported the issue for psutil in https://github.com/giampaolo/psutil/issues/1892.

giampaolo:
Also see https://github.com/giampaolo/psutil/issues/1892#issuecomment-748485501.

shirou:
FYI: [psutil fixed by using sysctl](https://github.com/giampaolo/psutil/pull/1895). I don't have an Apple M1. So I hope some other kind person make a PR.

Lomanic:
> FYI: psutil fixed by using sysctl.

For the record, it is not fixed in psutil, see https://github.com/giampaolo/psutil/issues/1892#issuecomment-748642856. And @giampaolo concluded the same thing as https://github.com/shirou/gopsutil/pull/999#issuecomment-734525301 in https://github.com/giampaolo/psutil/issues/1892#issuecomment-748643798: these values are not linked 	and don't seem to represent the same thing.

For the moment, it seems impossible to get the nominal CPU frequency on Apple M1 devices.

shoenig:
Indeed, I'll go ahead and close this PR

PR #1001, opened 2020-11-25
Update README.rst

ZhangLei-cn:
Use pkg.dev.go instead of godoc. Because pkg.dev.go is clearer. Now godoc has popped up a tip: "Pkg.go.dev is a new destination for Go discovery & docs. Check it out at pkg.go.dev/github.com/shirou/gopsutil and share your feedback."


Lomanic:
You can't access gopsutil v2 documentation from pkg.go.dev, you can still opt into being redirected automatically from godoc.org to pkg.go.dev (a cookie setting probably) or manually by clicking in the link in the yellow header, and I really prefer godoc.org UX personally.

PR #1010, opened 2020-12-13
Fix wrong OpenBSD user count

gdbinit:
There are empty entries in OpenBSD's utmp that have a line and time entry but no user logged in (the entry is "cleared" after user logs out but not totally as expected in the code here). Current checks are insufficient so check if Name field is empty and skip in that case.


Lomanic:
Would you please fix this also in the same commit in `v3/host/host_openbsd.go`?

shirou:
I will create an another PR to adapt this PR to v3. Thank you!

PR #1017, opened 2021-01-03
[process][linux] Fixing smaps parser

rishabh-arya95:
There seems to be a bug in the `SMAPS` parsing, what I am observing is that the values in the last block of `/proc/pid/smaps` are not being included in the final output. If you see the [sample input](https://gist.github.com/rishabh-arya95/1d98ce3a4053d2989ca0320e06b42643), the value of Size should be `20` but it comes out `16`.

Also, the blocks are not getting the correct pathname, they are getting the name of the next block that is because in the first iteration the first `if`  condition becomes true, and the first line always gets skipped.

One minor issue is that we are creating a slice `blocks` of length 16 and then we are appending to it, so this will lead to increasing the length of the slice which is inefficient.

This PR fixes the above issues

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -399,9 +399,9 @@ func (p *Process) MemoryMapsWithContext(ctx context.Context, grouped bool) (*[]M
 	lines := strings.Split(string(contents), "\n")
 
 	// function of parsing a block
-	getBlock := func(first_line []string, block []string) (MemoryMapsStat, error) {
+	getBlock := func(firstLine []string, block []string) (MemoryMapsStat, error) {
 		m := MemoryMapsStat{}
-		m.Path = first_line[len(first_line)-1]
+		m.Path = firstLine[len(firstLine)-1]
 
 		for _, line := range block {
 			if strings.Contains(line, "VmFlags") {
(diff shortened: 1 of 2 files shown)


shirou:
I confirm current parser skips a first block.

Current

```
{"path":"[vvar]","rss":0,"size":0,"pss":0,"sharedClean":0,"sharedDirty":0,"privateClean":0,"privateDirty":0,"referenced":0,"anonymous":0,"swap":0}
{"path":"[vdso]","rss":0,"size":12,"pss":0,"sharedClean":0,"sharedDirty":0,"privateClean":0,"privateDirty":0,"referenced":0,"anonymous":0,"swap":0}
{"path":"[vsyscall]","rss":0,"size":4,"pss":0,"sharedClean":0,"sharedDirty":0,"privateClean":0,"privateDirty":0,"referenced":0,"anonymous":0,"swap":0}
```

After this PR.

```
{"path":"[vvar]","rss":0,"size":12,"pss":0,"sharedClean":0,"sharedDirty":0,"privateClean":0,"privateDirty":0,"referenced":0,"anonymous":0,"swap":0}
{"path":"[vdso]","rss":0,"size":4,"pss":0,"sharedClean":0,"sharedDirty":0,"privateClean":0,"privateDirty":0,"referenced":0,"anonymous":0,"swap":0}
{"path":"[vsyscall]","rss":0,"size":4,"pss":0,"sharedClean":0,"sharedDirty":0,"privateClean":0,"privateDirty":0,"referenced":0,"anonymous":0,"swap":0}
```

PR #1019, opened 2021-01-11
Fix lsof args for udp4 in unix

MashaSamoylova:
It's currently impossible to list UDP4 ports in Unix due to a typo


Lomanic:
It looks like the [last line effect](https://www.viva64.com/en/b/0260/) in action, thanks a lot for catching that @MashaSamoylova.

PR #1030, opened 2021-02-04
Clarify the godoc of the Children function.

pawelz:
The previous godoc string was slightly confusing and only described information that can be deduced from the function signature.


shirou:
Indeed. Thank you so much!

PR #1035, opened 2021-02-16
Fix spaces on long process names for MacOS

### LINKED ISSUE #1034: Processes with spaces in the name on MacOS get truncated
**Describe the bug**
Many end-user applications on MacOS have long names and spaces in the name (Google Chrome, Microsoft Word, etc). With the way the name field is retrieved for long names with ps and splitting on spaces, the name field will only include the first part of the process name.

**To Reproduce**
```go
package main

import (
	"fmt"
	"io/ioutil"
	"os"
	"os/exec"
	"path/filepath"
	"time"

)

func main() {

	tmpdir, _ := ioutil.TempDir("", "")

	defer os.RemoveAll(tmpdir) // clean up
	tmpfilepath := filepath.Join(tmpdir, "name with spaces.go")
	tmpfile, _ := os.Create(tmpfilepath)

	tmpfilecontent := []byte("package main\nimport(\n\"time\"\n)\nfunc main(){\nfor range time.Tick(time.Second) {}\n}")
	if _, err := tmpfile.Write(tmpfilecontent); err != nil {
		tmpfile.Close()
		fmt.Println("unable to write temp file %v", err)
	}
	if err := tmpfile.Close(); err != nil {
		fmt.Println("unable to close temp file %v", err)
	}

	_ = exec.Command("go", "build", "-o", tmpfile.Name()+".exe", tmpfile.Name()).Run()

	cmd := exec.Command(tmpfile.Name() + ".exe")

	cmd.Start()
	time.Sleep(100 * time.Millisecond)
	p, _ := process.NewProcess(int32(cmd.Process.Pid))

	n, _ := p.Name()

	basename := filepath.Base(tmpfile.Name() + ".exe")
	if basename != n {
		fmt.Printf("%q != %q\n", basename, n)
	} else {
		fmt.Printf("%q == %q\n", basename, n)
	}
	cmd.Process.Kill()
}

```

**Expected behavior**
The full name of the process including spaces should be retrieved.

 - [x] Mac OS: [paste the result of `sw_vers` and `uname -a`
ProductName:    Mac OS X
ProductVersion: 10.15.7
BuildVersion:   19H2
Darwin xyz 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64

**Additional context**

jblesener:
Fixes #1034.

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -76,7 +76,7 @@ func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 	name := common.IntToString(k.Proc.P_comm[:])
 
 	if len(name) >= 15 {
-		cmdName, err := p.CmdNameWithContext(ctx)
+		cmdName, err := p.cmdNameWithContext(ctx)
 		if err != nil {
 			return "", err
 		}
@@ -101,8 +101,8 @@ func (p *Process) CmdlineWithContext(ctx context.Context) (string, error) {
 	return strings.Join(r[0], " "), err
 }
 
-// CmdNameWithContext returns the command name (including spaces) without any arguments
-func (p *Process) CmdNameWithContext(ctx context.Context) ([]string, error) {
+// cmdNameWithContext returns the command name (including spaces) without any arguments
+func (p *Process) cmdNameWithContext(ctx context.Context) ([]string, error) {
 	r, err := callPsWithContext(ctx, "command", p.Pid, false, true)
 	if err != nil {
 		return nil, err
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -38,7 +38,7 @@ type _Ctype_struct___0 struct {
 func pidsWithContext(ctx context.Context) ([]int32, error) {
 	var ret []int32
 
-	pids, err := callPsWithContext(ctx, "pid", 0, false)
+	pids, err := callPsWithContext(ctx, "pid", 0, false, false)
 	if err != nil {
 		return ret, err
 	}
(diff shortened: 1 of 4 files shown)


process/process_darwin.go
```diff
@@ -94,28 +94,37 @@ func (p *Process) NameWithContext(ctx context.Context) (string, error) {
...
-	r, err := callPsWithContext(ctx, "command", p.Pid, false)
+	r, err := callPsWithContext(ctx, "command", p.Pid, false, false)
 	if err != nil {
 		return "", err
 	}
 	return strings.Join(r[0], " "), err
 }
 
+// CmdNameWithContext returns the command name (including spaces) without any arguments
+func (p *Process) CmdNameWithContext(ctx context.Context) ([]string, error) {
```
shirou, line 105: Could you change this function name to lower camel in order to not expose ?

shirou:
Thank you for your great PR. I add a comment but others are good!

I can confirm it works on Mac BigSur. And from `man ps` on Mac BigSur, 
```
     -c      Change the ``command'' column output to just contain the executable name, rather than the full command line.
```

So it is good way to add this arg to get only command name.

(by the way, ps on Mac comes from BSD ps, so I think this patch is very useful for other bsd families)

process/process_darwin.go
jblesener: Done and thanks for the review. Sorry to cause trouble, I'm relatively new to golang.

jblesener:
Hmm, I don't have any BSD instances to test against. Let me see if I can spin up a VM or two in order to port these changes over in a separate PR.

shirou:

PR #1036, opened 2021-02-16
cpu, v3/cpu: use sysconf package instead of exec'ing getconf

tklauser:
Currently, `ClocksPerSec` is determined by `exec`'ing `getconf` in `func init`,
i.e. on startup of every program importing the package. `getconf` might
not be present on some systems or is not executable by the current user.
To avoid this hard to control dependency, use the
`github.com/tklauser/go-sysconf` package which implements `sysconf(3)`
entirely in Go without cgo. The package is supported on all platforms
currently supported by the `cpu` and `v3/cpu` package of gopsutil.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -48,6 +48,22 @@
   version = "v1.6.0"
 
 [[projects]]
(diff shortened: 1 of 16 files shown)


shirou:
I read your code and this library is great, and the license is BSD. Thank you! I add your awesome library with big appreciate.

PR #1038, opened 2021-02-19
add compile assert

wziww:


wziww:
```go
math.Abs(float64(v[0].PacketsRecv-pr)) > 5 // went error in the situation like math.Abs(1-2)
```

shirou:
Excuse me, could you tell us why you are adding interface and compiler check while we are not provide interface?

wziww:
Isn't it more standard to provide a "header file" throughing `interface`?(Which can be more clearly to tell others what they need to do to maintain other OSs or Archs while some one adding or changing methods ?

shirou:
No. We don't provide an `interface` and even if we provide an interface, checking to ensure a type satisfies an interface is a library user's responsibility.

PR #1042, opened 2021-03-01
Add missing TargetConditionals.h include, closes #976

### LINKED ISSUE #976: error: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Werror,-Wundef-prefix=TARGET_OS_]
If you attempt to build this package on macOS using LLVM/Clang 11.0 (the upstream version number, not Apple's version numbers, which are different), you get the following error:

```
cpu/cpu_darwin_cgo.go:13:5: error: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Werror,-Wundef-prefix=TARGET_OS_]
#if TARGET_OS_MAC
    ^
1 error generated.
```

This is because LLVM/Clang 11 automatically adds `-Werror,-Wundef-prefix=TARGET_OS_` as default compiler arguments and apparently `TARGET_OS_MAC` isn't defined in this compilation environment.

The reason it is not defined, I'm not sure. `TargetConditionals.h` defines this symbol and unless I'm looking at the wrong SDK, I think `TARGET_OS_MAC` should always be defined when building on/for macOS.

I can work around this by adding `#cgo CFLAGS: -Wno-undef-prefix` or `#include <TargetConditionals.h>` to the failing `import "C"` comment. The latter is likely more robust and seemingly indicates that `TargetConditions.h` isn't being included by another `#include` in this block.

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15.7
BuildVersion:   19H2
```

I'm also building with the `macosx10.15` SDK.

bobrik:


bobrik:
I'm fairly certain that the test failure on macos 10.12 (EOL 2019-11-30) is not due to this change:

```
--- FAIL: TestNetIOCountersAll (0.01s)
    net_test.go:91: invalid sum value: 127622, 127623
FAIL
FAIL	github.com/shirou/gopsutil/net	0.130s
ok  	github.com/shirou/gopsutil/process	5.232s
FAIL
Error: Process completed with exit code 1.
```

shirou:
I don't have a Mac but it seems [TargetConditionals.h](https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-18.1/TargetConditionals.h.auto.html) has `TARGET_OS_MAC` , so it might fix the issue. I think `TargetConditionals.h` is provided Apple and all Mac version includes it.

PR #1045, opened 2021-03-06
[process][windows] Fix #714 call OpenProcess with PROCESS_QUERY_INFORMATION on WinXP

### LINKED ISSUE #714: OpenProcess fails with PROCESS_QUERY_LIMITED_INFORMATION on Windows XP
Creating this issue to reference this behavior on Windows XP, quoting myself from https://github.com/shirou/gopsutil/pull/713#issue-295118285

> Regarding WinXP, this is failing because of PROCESS_QUERY_LIMITED_INFORMATION being introduced in Vista. Calling `windows.OpenProcess` with `windows.PROCESS_QUERY_INFORMATION` works properly on WinXP (hinting that every calls to OpenProcess with PROCESS_QUERY_LIMITED_INFORMATION in gopsutil fail on WinXP). Handling this (opening process with right parameter according to Windows version) would complicate a little bit the code in process_window.go (while being prudent not too slow down gopsutil functions), I'm not sure if we should bother so much about WinXP now that it's deprecated for quite a few years.

I think it would be nice to keep compatibility with WinXP, but with a low priority (we didn't even have report from users that these calls to OpenProcess are broken on XP), I will probably implement something in the future if nobody else does.

Lomanic:
[process][windows] Fix #714 call OpenProcess with PROCESS_QUERY_INFORMATION on WinXP

That's just for the beauty and the fun of it, WinXP is unsupported for years but this change doesn't harm the code much.

If I remember correctly, psutil still supports WinXP also, so it makes us a little bit more aligned with them.

This was tested successfully on WinXP.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -38,6 +38,8 @@ var (
 	processorArchitecture uint
 )
 
+const processQueryInformation = windows.PROCESS_QUERY_LIMITED_INFORMATION | windows.PROCESS_QUERY_INFORMATION // WinXP doesn't know PROCESS_QUERY_LIMITED_INFORMATION
+
 type SystemProcessInformation struct {
 	NextEntryOffset   uint64
 	NumberOfThreads   uint64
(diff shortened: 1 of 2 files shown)


shirou:
LGTM! Thank you so much!

PR #1051, opened 2021-03-18
[process][posix] Fix #1049 check if procfs is mounted before checking if pid exists there

### LINKED ISSUE #1049: Freebsd(jail): process.NewProcess doesnt work if no procfs mounted
**Describe the bug**
Important, this is within jail. I don't have freshly installed FreeBSD to check what is situation on fresh install.

Within jails procfs is not mounted by default, but /proc exists. 

Due to this, process.NewProcess(pid) doesnt work.

The issue is handling within process_posix.go

`func PidExistsWithContext(ctx context.Context, pid int32) (bool, error)`

`if _, err := os.Stat(common.HostProc()); err == nil { //Means that proc filesystem exist`

The issue is that HostProc returns /proc by default, and the directory is really there but if procfs is not in use, it is just empty directory. When stat checks for /proc it is found and presumes procfs is there. 

Further down the code it checks if process is within the /proc direcectory and always fails.
 
`_, err := os.Stat(common.HostProc(strconv.Itoa(int(pid))))`

Even if you are having handling for systems without procfs mounted, it is never called.

**To Reproduce**
```go
    _, err := process.NewProcess(int32(os.Getpid()))
```

**Expected behavior**
The check for procfs is inadequate, fast fix would be to instead rather check if /proc is empty while the ultimate solution would be to actually check if procfs is mounted. Workaround would be to mount procfs but it would be nice if this wouldnt be a requirement.

 - [x] FreeBSD: [
FreeBSD build 12.2-RELEASE FreeBSD 12.2-RELEASE r366954 GENERIC  amd64
12.2-RELEASE-p3

**Additional context**
No cross compiling.

Lomanic:
[process][posix] Fix #1049 check if procfs is mounted before checking if pid exists there

Benchmark before this change (process.NewProcess() calls process.PidExistsWithContext()
internally)

    go test -bench=BenchmarkNewProcess github.com/shirou/gopsutil/process
    goos: linux
    goarch: amd64
    BenchmarkNewProcess-4              14722             78751 ns/op
    PASS
    ok      github.com/shirou/gopsutil/process      3.685s

Benchmark with this change applied

    go test -bench=BenchmarkNewProcess github.com/shirou/gopsutil/process
    goos: linux
    goarch: amd64
    BenchmarkNewProcess-4              14835             80180 ns/op
    PASS
    ok      github.com/shirou/gopsutil/process      3.761s

### DIFF
--- a/process/process_posix.go
+++ b/process/process_posix.go
@@ -71,6 +71,25 @@ func getTerminalMap() (map[uint64]string, error) {
 	return ret, nil
 }
 
+// isMount is a port of python's os.path.ismount()
+// https://github.com/python/cpython/blob/08ff4369afca84587b1c82034af4e9f64caddbf2/Lib/posixpath.py#L186-L216
+// https://docs.python.org/3/library/os.path.html#os.path.ismount
+func isMount(path string) bool {
+	var stat1 unix.Stat_t
+	if err := unix.Lstat(path, &stat1); err != nil {
+		return false
+	}
+	if stat1.Mode == unix.DT_LNK {
+		return false
+	}
+	parent := filepath.Join(path, "..")
+	var stat2 unix.Stat_t
+	if err := unix.Lstat(parent, &stat2); err != nil {
+		return false
+	}
+	return stat1.Dev != stat2.Dev || stat1.Ino == stat2.Ino
+}
+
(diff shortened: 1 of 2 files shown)


shirou (approved):
I have confirmed this works on my FreeBSD which I unmount the procfs. Thank you so much!

PR #993, opened 2020-11-09
Make types for OpenBSD/arm64

vext01:
As requested [here](https://github.com/shirou/gopsutil/issues/992#issuecomment-723638840).

These were generated on OpenBSD-current on a Raspberry Pi 4.

Can you tell me how to test the build and run the test suite? I don't know anything about golang!

Fixes #992

PR #1052, opened 2021-03-23
add support for OpenBSD arm64

gballet:


Lomanic:
I finally ordered a Raspberry Pi 4 to be able to test gopsutil on arm64 on various OSes, I received it last week but I failed to install openbsd 6.8 on it last weekend (it is not as simple as running `dd` over an sd card and booting it). I will follow up on this not before the next weekend. Previous work was started in #993 but hit some roadblocks as I was only able to test things without hardware with @vext01 helpful assistance.

Can any of @vext01 or @tobias-hildebrandt run tests on this PR on your hardware (`go test github.com/shirou/gopsutil/...` or `go test ./...`)?

shirou:
(sorry to off-topic)

How about to start a GitHub sponsor, @Lomanic ? I want to pay your Raspberry Pi!

tobias-hildebrandt:
Hello, I've run the tests on my Raspberry Pi 4:
[gopsutil-log.txt](https://github.com/shirou/gopsutil/files/6313983/gopsutil-log.txt)

gballet:
@tobias-hildebrandt are you running OpenBSD on bare metal or in a VM? `error operation not supported` comes from a `sysctl` that checks for CPU frequencies.

tobias-hildebrandt:
I'm running OpenBSD on bare metal. 

Your diagnosis seems correct, it looks like I'm unable to check the CPU frequency via `sysctl hw.cpuspeed`, the output is `sysctl: hw.cpuspeed: value is not available`

I noticed said command when looking at the man pages for [cpu(4)](https://man.openbsd.org/cpu.4), though it seems that the man page doesn't exist for arm64, only amd64, hppa, and i386. In my brief search online, I wasn't able to find another way to check the processor frequency.

Perhaps the output of my `sysctl hw` could help? [sysctl_hw.txt](https://github.com/shirou/gopsutil/files/6325368/sysctl_hw.txt) (I've redacted some information that might be sensitive)

Let me know if there is any other information I can provide that would be of help.

gballet:
I see, the problem isn't the VM part, the `sysctl` isn't implemented on `arm64` at all :angry: Ok, I'll work around it, thank you.

Lomanic:
As-is it's OK to merge, thanks a lot @gballet.

About the missing `sysctl hw.cpuspeed`, it can be fixed afterwards.

@shirou you can setup GitHub sponsors if you want, but I didn't mind buying that Pi, I could also find it some use outside of gopsutil.

PR #1056, opened 2021-04-05
add a reliable way to get fields from /prod/PID/stat

sakateka:
The original way of splitting /proc/[PID]/stat will fail if the comm name contains a right parenthesis, this character is valid for the comm name.
So a reliable way to parse "stat" is to look for the right parenthesis at the end of the content.


shirou:
Sorry to late response. It looks great! but could you add some tests which read a file from under `testdata` and parse it by using `splitProcStat` ?

sakateka:
@shirou PTAL

shirou:
sorry to late. Thank you for adding tests!

PR #1059, opened 2021-04-13
use ioreg to read IOPlatformUUID as HostID

d1ss0nanz:
Since "sysctl kern.uuid" is not unique across Darwin hosts, "IOPlatformUUID" should be used instead.


d1ss0nanz:
I've no idea why the checks failed. Let me know what I can do to fix it.

shirou:
Sorry to late response. Now I have a darwin (for a while) and can confirm this PR is works. Thank you so much!

PR #1063, opened 2021-04-24
Make limitToUint parse to uint instead of int

damilola-bello:
Attempting to parse a soft/hard limit greater than int32 fails and returns 0 instead.
I believe soft/hard limits are always unsigned, hence no need to attempt to parse as int.

### DIFF
--- a/v3/process/process_linux.go
+++ b/v3/process/process_linux.go
@@ -485,11 +485,11 @@ func limitToUint(val string) (uint64, error) {
 	if val == "unlimited" {
 		return math.MaxUint64, nil
 	} else {
-		res, err := strconv.ParseInt(val, 10, 32)
+		res, err := strconv.ParseUint(val, 10, 64)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


shirou:
Good catch! LGM, but could you add same thing to v2 also? thanks!

damilola-bello:
@shirou 
Fixing v2 will break backward compatibility as it uses `int32` for soft/hard limits: https://github.com/shirou/gopsutil/blob/c7a38de76ee583adf3f4aa276d834e2e753be321/process/process.go#L69

So nothing much can be done to change `limitToInt` here: https://github.com/shirou/gopsutil/blob/c7a38de76ee583adf3f4aa276d834e2e753be321/process/process_linux.go#L489

Let me know if I'm missing anything. Thanks.

shirou:
Sorry, you are correct. That struct has been changed at v3. so nothing can be done. Thank you for your contribution!

PR #1064, opened 2021-05-03
[cpu][solaris] psrinfo parsing fixes

scop:


shirou:
Could you add same diff to v2 (this means [here](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_solaris.go))?

scop:
Done.

shirou (approved):

PR #1065, opened 2021-05-03
Basic Solaris process support

scop:


scop:
Done, but pretty much blindfolded, my setup/IDE is not really up to building v2 properly. Result seems to work though. Did not run the tests.

Noteworthy is that on OpenIndiana oi_151a8, `/proc/<pid>/execname` and `/proc/<pid>/cmdline` don't exist so `Exe()` and `Cmdline*()` won't work, they fail with an error trying to open those files. They are present and the calls thus do work on Solaris 11.3.

shirou (approved):
LGTM! Thank you so much!

PR #1074, opened 2021-05-19
[load][solaris] support MiscStat.ProcsRunning

scop:
Seems other states reported are not useful for us, https://docs.oracle.com/cd/E26502_01/html/E29030/ps-1.html


shirou:
Thank you so much! I thought for a minute that `W` was the same as blocking, but a closer reading shows it's not because it can be interrupted.

PR #1075, opened 2021-05-19
[process][solaris] improve Exe portability

scop:
Resolving from path/a.out seems more portable than from execname; works both in OpenIndiana oi_151a7 and Solaris 11.3. I left execname lookup in as a fallback although I don't know on which systems that code path would be excercised any more.

### DIFF
--- a/process/process_solaris.go
+++ b/process/process_solaris.go
@@ -66,7 +66,11 @@ func (p *Process) TgidWithContext(ctx context.Context) (int32, error) {
 }
 
 func (p *Process) ExeWithContext(ctx context.Context) (string, error) {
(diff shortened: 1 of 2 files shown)


shirou:
I don't have a much experience about solaris, but a solaris have a `/proc/<pid>/path/a.out` path? Could you share some documentation of proc structure?

scop:
https://docs.oracle.com/cd/E23824_01/html/821-1473/proc-4.html -- see the `object` and `path` entries. The `a.out` in `object` is not a symlink but a regular file (the executable itself).

shirou:
Thank you for the information. very interesting.
```
/proc/pid/object/a.out
    alias for process's executable file
```

PR #1076, opened 2021-05-20
[cpu][linux] more info for ARM processors

scop:
https://developer.arm.com/documentation/ddi0337/h/system-control/register-descriptions/cpuid-base-register--cpuid

See the existing `testdata/linux/1037/proc/cpuinfo` sample.

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -142,8 +142,39 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 		case "vendorId", "vendor_id":
 			c.VendorID = value
 		case "CPU implementer":
-			if value == "0x41" {
-				c.VendorID = "ARM"
+			if v, err := strconv.ParseUint(value, 0, 8); err == nil {
+				switch v {
+				case 0x41:
+					c.VendorID = "ARM"
+				case 0x42:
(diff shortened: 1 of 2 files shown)


cpu/cpu_linux.go
```diff
@@ -141,9 +141,13 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 			c.CPU = int32(t)
 		case "vendorId", "vendor_id":
 			c.VendorID = value
+		case "CPU implementer":
+			if value == "0x41" {
```
shirou, line 145: Could you add more implementers according to [ARM document](https://developer.arm.com/documentation/ihi0014/q/Programmers--Model/Detailed-register-descriptions/ID-Register--ETMIDR--ETMv2-0-and-later) and [linux code](https://github.com/torvalds/linux/blob/master/arch/arm64/include/asm/cputype.h#L54)?
```
// from https://github.com/torvalds/linux/blob/master/arch/arm64/include/asm/cputype.h#L54
switch value {
  case "0x41":
     c.VendorID = "ARM"
  case "0x50":
     c.VendorID = "APM"
.....
(snip)
  case "0x61":
     `c.VndorID` = "APPLE"  
}
```

shirou:
Thank you for your great work!

PR #1081, opened 2021-06-02
process, v3/process: use SC_CLK_TCK sysconf value instead of hard-coding clock ticks

tklauser:
The github.com/tklauser/go-sysconf package is already a dependency used
in the cpu and v3/cpu packages to determine clock ticks using
`sysconf.Sysconf(sysconf.SC_CLK_TCK)`, see #1036. Use the same in
packages process and v3/process as well instead of hard-coding clock
ticks to 100.


shirou:
Great, thank you!

PR #1083, opened 2021-06-04
process, v3/process (Linux): fix Processes in SmartOS lx containers

elmeyer:
In Linux containers running in LX Branded Zones on SmartOS (potentially
other Solaris-based OSes with LX Branded Zones),
`fillFromTIDStatWithContext` panics as the `delayacct_blkio_ticks` field is
not present and thus results in an out-of-bounds slice access.
Check the slice length before and only attempt to parse the
`delayacct_blkio_ticks` field if there is an appropriate number of fields.

I am not sure if the way I implemented the test is acceptable, especially
w.r.t. naming. Please suggest a correction if necessary.


shirou:
Thank you for your contribution. This PR looks good and test naming is ok. but could you copy the test function to v3?

elmeyer:
Done, thanks for the hint!

shirou:
LGTM! Thank you for your contribution!

PR #1087, opened 2021-06-11
process, v3/process (openbsd):  remove remaining CGO bits from CmdlineSliceWithContext

ivandeex:
This PR is a rework of stalled PR #965 which rebases it to current master and extends the patch to v3.

We are using gopsutil v3 in the [rclone project](https://github.com/rclone/rclone) (https://github.com/rclone/rclone)
so users can report their system environment to the project support team.
In fact we use only a [limited subset](https://github.com/ivandeex/rclone/blob/ci-version-openbsd/lib/buildinfo/osversion.go#L12) of gopsutil features.

rclone is widely deployed on Linux, Windows, MacOS and a range of other OSes as well as in docker.
Due to the wide range of supported systems we distribute rclone as a statically built binary.
Unfortunately the "C" bits in process_openbsd.go prevent us from building rclone with gopsutil on OpenBSD.

**We are very interested in having a MVP gopsutil for OpenBSD to fill in the last missing bit.**

I noticed that #965 is stalled since October 2020 and improved the patch so it applies successfully
to the current gopsutil master, also on v3. I then built rclone with patched gopsutil and successfully
ran rclone environment reporter on OpenBSD 6.7 x86_64 ([Virtualbox](https://app.vagrantup.com/fnichol/boxes/openbsd-6.7)).

```
rclone v1.56.0-DEV
- os/version: openbsd 6.7 (64 bit)
- os/kernel: 6.7 (amd64)
- os/type: openbsd
- os/arch: amd64
- go/version: go1.16.4
- go/linking: static
- go/tags: none
```

@shirou 
Please review and merge this PR (or help #965 move a little faster) :-)
:pray: :pray: :pray:

_Original commit message follows_:

> Prior to this commit CGO was used in OpenBSD implementation of
> Process.CmdlineSliceWithContext() for parsing the "kern.proc.args"
> sysctl output. It requires some pointer arithmetics and raw pointer
> dereferencing.
> 
> Having CGO in the "process" module prevents it from being go vet'ted
> on any platform other than OpenBSD. In order to overcome this
> limitation, the sysctl output parsing was reimplemented without raw
> pointer deferencing. The resulting code might be slightly slower
> than the original one, but it is cleaner and safer.
> 
> Since this fix allows go vet with GOOS=openbsd to run without any
> issues on all platforms, openbsd entries were also added to the "vet"
> Makefile target.


ivandeex:
https://github.com/rclone/rclone/pull/5402 is waiting for this PR

shirou:
Thank you for waiting. I adapt this PR. Thank you for your contribution!

PR #1088, opened 2021-06-15
processor, v3/processor (Win): fix slow cpuinfo on multisocket config

fredwangwang:
updated win32_Processor struct to exclude (unused) LoadPercentage field.
The loadpercentage takes linearly more time as the # of sockets
increases. By default vSphere maps 1 vCPU to 1 socket, resulting in very
poor performance when getting CPU info against, saying, 40 vCPU VM
(basically 40 sockets as seen by the VM).

Here is the before and after time comparison:
printcpu-diff

changes:

for cpu:
since `Win32_Processor` is a public struct, created a substruct named: `Win32_ProcessorWithoutLoadPct` for use in all queries without breaking any potential usage of the public api.

for v3/cpu:
since the struct now is private, simply removed `LoadPercentage`

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -5,6 +5,7 @@ package cpu
 import (
 	"context"
 	"fmt"
+	"strings"
 	"unsafe"
 
 	"github.com/StackExchange/wmi"
@@ -18,7 +19,14 @@ var (
 )
 
 type Win32_Processor struct {
-	LoadPercentage            *uint16
+	Win32_ProcessorWithoutLoadPct
+	LoadPercentage *uint16
+}
+
+// LoadPercentage takes a linearly more time as the number of sockets increases.
+// For vSphere by default corespersocket = 1, meaning for a 40 vCPU VM Get Processor Info
+// could take more than half a minute.
+type Win32_ProcessorWithoutLoadPct struct {
 	Family                    uint16
 	Manufacturer              string
 	Name                      string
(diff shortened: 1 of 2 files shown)


idrennanvmware:
@shirou  - any chance we could prioritize this?  This particular issue is trickling down through our usage of Nomad and in turn the long fingerprint times on our ESXI hosts are causing issues.   Very much appreciate the work you've done, and our ability to contribute to this.  If you need us to address anything on the PR, please let us know and we will be happy to do so.

Thanks!

shirou:
I can only have a time on weekends.

idrennanvmware:
Totally understand.  And ty again for all your hard work!


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On June 17, 2021 at 7:21:50 PM EDT, shirou ***@***.***> wrote:

I can only have a time on weekends.

shirou:
Thank you for your contribution and your patiant.

- As you mentioned, `LoadPercentage` is not used currently. so please just remove `LoadPercentage` without creating another struct?
- Could you also implement to v3 in same way?

fredwangwang:
Hi @shirou thanks for reviewing!

See my comments above:
> changes:
>
> for cpu:
> since Win32_Processor is a public struct, created a substruct named: Win32_ProcessorWithoutLoadPct for use in all queries without breaking any potential usage of the public api.
> 
> for v3/cpu:
> since the struct now is private, simply removed LoadPercentage

shirou:
Oh, OK, I understand. Then, it is a good implementation! Thank you so much!

fredwangwang:
Hi @shirou looks like the shellcheck is the only part fails, but that feels unrelated to the pr's change. Could you help to take a look at that? Thank you!

shirou:
some of the tests are stoled but mostly ok. So I merge this PR. Thank you!

idrennanvmware:
Thanks so much @fredwangwang and @shirou !

fredwangwang:
Hi @shirou is it possible to cut a new release? The downstream package consumer (hashicorp nomad) prefers a tagged build instead of using a specific sha.

Thanks in advance!

shirou:
gopsutil will be released on [end of month](https://github.com/shirou/gopsutil/#tag-semantics).

fredwangwang:
Ty!

PR #1090, opened 2021-06-24
simplify mutex usage

MyaLongmire:
Mutexs don't have to be allocated so this was an unnecessary pointer.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -27,7 +27,7 @@ type Process struct {
 	name           string
 	status         string
 	parent         int32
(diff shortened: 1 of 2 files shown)


MyaLongmire:
The test that is failing is not failing because of code but because it doesn't like the command being run. 
```
Run git ls-files --exclude='*.sh' --ignored | xargs shellcheck
fatal: ls-files -i must be used with either -o or -c
No files specified.
```

shirou (approved):
Indeed. Thank you so much!

PR #1096, opened 2021-07-09
Update wmi to go.mod supported version

lespea:


shirou (approved):
Great! Thank you so much for your contribution to not only gopsutil but also wmi!

PR #1097, opened 2021-07-13
feat: Add support for environment variable read

secDre4mer:
Add support (for Linux and Windows for now) for reading the
environment variables of a process, with a similar syntax to
os.Environ().
For Windows, this includes some refactoring for clean access to the
RTL_USER_PROCESS_PARAMETERS structure which points to the command
line and the environment variables.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -65,7 +65,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:0afe79d034c63eea8c5977401f5fc135394751c1213ee6c991b0c6bec44f8f60"
+  digest = "1:1672d2eb5799465379f6016fa509eadd72c69e3eb2bf3cdfea9222e64e163b94"
   name = "golang.org/x/sys"
   packages = [
     "internal/unsafeheader",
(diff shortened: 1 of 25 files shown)


shirou:
Thank you for great PR! [psutil has `environ()`](https://psutil.readthedocs.io/en/latest/#psutil.Process.environ), so we are very welcome about this feature.

However, when I test on my local Windows 10(64bit), I got these error.

```
--- FAIL: Test_Process_CmdLine (0.05s)
    process_test.go:153: getting cmdline error could not get CommandLine: cannot read cmdline
    process_test.go:156: invalid cmd line
--- FAIL: Test_Process_CmdLineSlice (0.00s)
    process_test.go:166: getting cmdline slice error could not get CommandLine: cannot read cmdline
```

I don't have a time to dig deeper, but these values are different on [here](https://github.com/shirou/gopsutil/pull/1097/files#diff-8c35ce0e2744b651f0d5cc72ca3bf15fc4be83a52ae6be5737dbd58d05c702d6R995).

```
len(cmdLine) = 114688
userProcParams.CommandLineLength = 16515322
```

Tests on github action are also failed. Could you check it out? Thank you!

secDre4mer:
Hi shirou, thanks for the reception! I've pushed a fix to the issue you highlighted and the tests should (hopefully) succeed now.

The v2 tests seem to fail because the golang.org/x/sys/windows dependency needs to be upgraded: https://github.com/shirou/gopsutil/runs/3076188820?check_suite_focus=true
I'm don't know how to do this, to be honest, since v2 uses an older dependency management system I'm not familiar with.

shirou (approved):
Now this PR can work on my windows10 environment. Thank you for your cool contritbution!

PR #1099, opened 2021-07-13
fix wmi version

annybs:
It seems like the StackExchange/wmi tag was changed after being installed to this repo, and it's caused us major problems trying to use the latest gopsutil:

```
verifying github.com/StackExchange/wmi@v1.2.0/go.mod: checksum mismatch
        downloaded: h1:rcmrprowKIVzvc+NUiLncP2uuArMWLCbu9SBzvHz7e8=
        go.sum:     h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
For more information, see 'go help module-auth'.
```

This PR redownloads dependencies so the tag actually matches the commit again.

### DIFF
--- a/v3/go.mod
+++ b/v3/go.mod
@@ -4,7 +4,6 @@ go 1.15
 
 require (
 	github.com/StackExchange/wmi v1.2.0
-	github.com/go-ole/go-ole v1.2.5 // indirect
 	github.com/stretchr/testify v1.6.1
 	github.com/tklauser/go-sysconf v0.3.6
 	golang.org/x/sys v0.0.0-20210316164454-77fc1eacc6aa
--- a/v3/go.sum
+++ b/v3/go.sum
@@ -1,5 +1,5 @@
-github.com/StackExchange/wmi v1.2.0 h1:noJEYkMQVlFCEAc+2ma5YyRhlfjcWfZqk5sBRYozdyM=
-github.com/StackExchange/wmi v1.2.0/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
+github.com/StackExchange/wmi v1.2.0 h1:BfLCNdXnvwgy5RrRI3IyQ64ZItZngXHN+7PxU5RvKxA=
+github.com/StackExchange/wmi v1.2.0/go.mod h1:rcmrprowKIVzvc+NUiLncP2uuArMWLCbu9SBzvHz7e8=
 github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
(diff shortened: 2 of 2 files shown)


shirou:
On my environment, there are no errors on current master branch. After patching this PR, I got this error.

```
verifying github.com/StackExchange/wmi@v1.2.0/go.mod: checksum mismatch
        downloaded: h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
        go.sum:     h1:rcmrprowKIVzvc+NUiLncP2uuArMWLCbu9SBzvHz7e8=

SECURITY ERROR
```

After `go get -u && go mod tidy`, there are no diff related to wmi (there are other diff's instead)

mmorel-35:
Dependabot seems to agree with the issue as it saying : 

image

lespea:
So I cleared out all of my go caches and re-verified everything and the current hashes do appear to be correct.  When WMI 1.2.0 was released the wrong commit was [very briefly pointing at the wrong commit](https://github.com/StackExchange/wmi/pull/58)... maybe it's possible that's the cause?  I'll open an issue to see if he can just do a new v1.2.1 tag/release and hopefully that will solve this.

annybs:
I believe @lespea is correct and that this is what I was referring to about the tag being changed: https://github.com/StackExchange/wmi/pull/58#issuecomment-877260951

In general, it's poor form to relocate a tag in a public repository, and this issue originates from precisely that being done in the WMI repository. It leads to scenarios like this where developer A with the original tag in their cache (shirou) has no problems, but developer B without it (me, Dependabot) cannot download the dependency, since the cached tag refers to a different commit than the one that is now tagged.

@shirou You will probably be able to reproduce the issue by installing to a 'clean' machine or, if you don't mind wiping your cache, executing `go clean -cache` and then redownloading the dependencies.

Unfortunately, the only 'good' way to fix this is for WMI to receive a new and completely unambiguous tag v1.2.1 - so, @lespea is correct again by raising the issue there. gopsutil can then be updated to use that tag and resolve the conflict 👍

annybs:
One note - I am installing `github.com/shirou/gopsutil v3.21.6+incompatible` to an OS X host, which _may_ produce different results than on other systems?

shirou:
After `go clean -cache` on this PR, but still "SECURITY ERROR" occurred. But it is true that dependabot is failing. So I can try to merge this PR. Thank you so much!

(I am really not a fan of current golang mod system).

shirou:
The dependabot [works](https://github.com/shirou/gopsutil/pull/1108). Great work!

shirou:
However, some of tests are [failing](https://github.com/shirou/gopsutil/pull/1110). perhaps caches in github actions.

lespea:
Might have to wait for a new tag to be pushed for wmi :/

Sorry this turned into such a headache.

annybs:
Yeah. Sorry!

PR #1105, opened 2021-07-20
Switch usage from os.*File.Readdir to os.*File.ReadDir.

fmarmol:
I suggest theses changes because I could see some improvements on my tests regarding the following benchmark:

```
import (
        "testing"

        "github.com/shirou/gopsutil/v3/net"
)

func BenchmarkConnections(b *testing.B) {
        for i := 0; i < b.N; i++ {
                net.Connections("all")
        }
}
```

```
✦ ➜ benchstat old.txt new.txt
name           old time/op    new time/op    delta
Connections-8    20.8ms ± 0%    16.7ms ± 0%   ~     (p=1.000 n=1+1)

name           old alloc/op   new alloc/op   delta
Connections-8    4.38MB ± 0%    4.01MB ± 0%   ~     (p=1.000 n=1+1)

name           old allocs/op  new allocs/op  delta
Connections-8     45.2k ± 0%     42.4k ± 0%   ~     (p=1.000 n=1+1)
```

Can you confirm that these improvements are relevants?

### DIFF
--- a/v3/net/net_linux.go
+++ b/v3/net/net_linux.go
@@ -545,12 +545,12 @@ func getProcInodes(root string, pid int32, max int) (map[string][]inodeMap, erro
 		return ret, err
 	}
 	defer f.Close()
-	files, err := f.Readdir(max)
+	dirEntries, err := f.ReadDir(max)
 	if err != nil {
 		return ret, err
 	}
-	for _, fd := range files {
-		inodePath := fmt.Sprintf("%s/%d/fd/%s", root, pid, fd.Name())
+	for _, dirEntry := range dirEntries {
+		inodePath := fmt.Sprintf("%s/%d/fd/%s", root, pid, dirEntry.Name())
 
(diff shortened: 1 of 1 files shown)


Lomanic:
[This was introduced in go1.16](https://pkg.go.dev/os#ReadDir), while the [v3 package minimum version is 1.15](https://github.com/shirou/gopsutil/blob/4746baa8c71363bf5f0456d006136156e53c44be/v3/go.mod#L3).

fmarmol:
@Lomanic  Ok thx for the reply, I didn't pay attention about the version. I think we can close this PR.

shirou:
Go 1.15 will be out of support when 1.17 is released. And perhaps 1.17 will be release within a few month, I want to leave this PR open.

shirou (approved):
Sorry to late response. Time passed from 1.17 released, I think it's safe to merge now. Thank you for your contribution!

PR #1106, opened 2021-07-21
Ensure rhel platform family for Rocky Linux

odinje:
Change: Ensure that the platform family field is set to rhel when using Rocky Linux.


shirou (approved):

PR #1112, opened 2021-07-29
[process][linux] get name from /proc/(pid)/comm

xuyang2:


shirou:
Thank you for opening PR.  `Name` is already retrieved from `proc/(pid)/status` by using `fillFromStatusWithContext`. According to [man proc 5](https://man7.org/linux/man-pages/man5/proc.5.html), `comm` and `Name` in `status` seems same value. Could you tell us why you are changing to use `comm` in this PR?

(I found [psutil uses `(pid)/stat` to get name](https://github.com/giampaolo/psutil/blob/master/psutil/_pslinux.py#L1669))

xuyang2:
@shirou Reading `comm`  is a few lighter than parsing the hole `status` file.

shirou:
How light is this? Could you share us the benchmark?

xuyang2:
@shirou Added two benchmark tests.
The test results on my vm are as follows:

```
goos: linux
goarch: amd64
pkg: github.com/shirou/gopsutil/v3/process
cpu: Intel(R) Core(TM) i7-4710MQ CPU @ 2.50GHz
Benchmark_fillFromCommWithContext
Benchmark_fillFromCommWithContext-4   	   74007	     15435 ns/op
Benchmark_fillFromStatusWithContext
Benchmark_fillFromStatusWithContext-4   	   37192	     31859 ns/op
```

shirou (approved):
Thank you for your update! Now it is easy to understand and the benchmark shows significant improvement. Great work!

PR #1113, opened 2021-07-29
Newest version of wmi

lespea:
Okay so this will hopefully fix all of the wmi woes.  No real changes just a new tag (and an updated readme).


shirou:
Great! Thank you so much!!

PR #1119, opened 2021-08-17
Fixes for Test_Connections in process_linux.go

tbarker25:
Test_Connections currently fails intermittently on Linux (and maybe other OSs).

On inspection, Go closes TCP connections when they go out of scope and are garbage collected. I've re-written Test_Connections() to avoid closing until the test has finished.

Other changes:
- Close connections properly in TestConnections so running `go test -times=N` works
- t.Skip() can not be called inside a test go-routine
- Make sure that Test_AllProcesses_cmdLine doesn't ignore failures.
- Some minor style changes motivated by staticcheck warnings.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -351,7 +351,7 @@ func (p *Process) PageFaultsWithContext(ctx context.Context) (*PageFaultsStat, e
 func (p *Process) ChildrenWithContext(ctx context.Context) ([]*Process, error) {
 	pids, err := common.CallPgrepWithContext(ctx, invoke, p.Pid)
 	if err != nil {
-		if pids == nil || len(pids) == 0 {
+		if len(pids) == 0 {
 			return nil, ErrorNoChildren
 		}
 		return nil, err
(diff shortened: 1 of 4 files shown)


Lomanic:
These changes also need to be applied in the v3 modules.

Commits history needs some love, the first one would be "Fix Test_Connections in process_linux.go" (only including changes to the Test_Connections function) with body the first part of your message in the PR and the proper connections closing bit, second commit would be about the various other fixes, with its body the rest of this PR message (or a separate commit for each point).

LGTM otherwise, tests are all green

tbarker25:
> These changes also need to be applied in the v3 modules.
> 
> Commits history needs some love, the first one would be "Fix Test_Connections in process_linux.go" (only including changes to the Test_Connections function) with body the first part of your message in the PR and the proper connections closing bit, second commit would be about the various other fixes, with its body the rest of this PR message (or a separate commit for each point).
> 
> LGTM otherwise, tests are all green

Thanks. I've made those changes.

shirou:
Sorry to late response. Thank you so much!

PR #1122, opened 2021-08-19
chore: Drop PROCESS_QUERY_INFORMATION support

secDre4mer:
PROCESS_QUERY_INFORMATION was used for compatibility reasons with
Windows XP / Server 2003. Both are no longer supported in Golang
and haven't been for a while (since Go 1.11). It should be safe
to drop this flag. PROCESS_QUERY_LIMITED_INFORMATION supports
PPL process such as LSASS, which are not queryable without this
change.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -40,7 +40,7 @@ var (
 	processorArchitecture uint
 )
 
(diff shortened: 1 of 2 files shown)


shirou:
Indeed. Thank you so much!

PR #1120, opened 2021-08-18
Add swapdevices method for Linux.

tbarker25:

PR #1124, opened 2021-08-19
Add mem.SwapDevices() method

tbarker25:
Same as #1120 but applied to v3


shirou:
Closed via #1131

shirou:
no, #1131 does not include v3. I will create PR later.

PR #1128, opened 2021-09-02
convert README from reStructuredText to Markdown

shibukawa:
I love reStructuredText as same as you. But pkg.go.dev can interpret Markdown only. This PR improve readbility of the following page.

https://pkg.go.dev/github.com/shirou/gopsutil


shirou (approved):
I love reST. However, the current situation is that only markdown is used. It is very, very sad, but I have to approve this PR.

Thank you anyway.

PR #1130, opened 2021-09-05
v3/process (Win): fix Kill() DuplicateHandle error

sunshineplan:
On golang 1.17 and latest windows 10, go test failed on testing Kill method because of DuplicateHandle error.
`DuplicateHandle: The handle is invalid.`


shirou:
Great catch!
I confirmed it works on the test and also usual go code.
This change has been introduced [4 month ago](https://github.com/golang/go/commit/105c5b50e0098720b9e24aea5efa8e161c31db6d).

Thank you for your contributon!

PR #1133, opened 2021-09-18
v3/diss (Linux): fix disk retrieve partition issue while using in container

derekhjray:
v3/disk (Linux): fix disk retrieve partition issue while using in container

if invoke 'disk.PartitionsWithContext(...)' in container, which '/proc/self/mountinfo' refered to a pid in container that has different mount namespace, and this function will return invalid results otherthan we required (host disk mount info)

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -221,15 +221,15 @@ var fsTypeMap = map[int64]string{
 func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, error) {
 	useMounts := false
 
-	filename := common.HostProc("self/mountinfo")
+	filename := common.HostProc("1/mountinfo")
 	lines, err := common.ReadLines(filename)
 	if err != nil {
 		if err != err.(*os.PathError) {
 			return nil, err
 		}
-		// if kernel does not support self/mountinfo, fallback to self/mounts (<2.6.26)
+		// if kernel does not support 1/mountinfo, fallback to 1/mounts (<2.6.26)
 		useMounts = true
-		filename = common.HostProc("self/mounts")
+		filename = common.HostProc("1/mounts")
 		lines, err = common.ReadLines(filename)
 		if err != nil {
 			return nil, err
@@ -261,7 +261,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 				}
 			}
 		} else {
-			// a line of self/mountinfo has the following structure:
+			// a line of 1/mountinfo has the following structure:
 			// 36  35  98:0 /mnt1 /mnt2 rw,noatime master:1 - ext3 /dev/root rw,errors=continue
 			// (1) (2) (3)   (4)   (5)      (6)      (7)   (8) (9)   (10)         (11)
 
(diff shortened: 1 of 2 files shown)


shirou:
From my poor understanding, `/proc/self` is a symbolic link on the process's own directory (from [man proc](https://man7.org/linux/man-pages/man5/proc.5.html)). In the container environment, PID is 1, so `self` and `1` is a same, I think, and according to my Docker environment, it is same file.

Could you tell us the situation what you are thinking more detail? Thank you!

derekhjray:
> From my poor understanding, `/proc/self` is a symbolic link on the process's own directory (from [man proc](https://man7.org/linux/man-pages/man5/proc.5.html)). In the container environment, PID is 1, so `self` and `1` is a same, I think, and according to my Docker environment, it is same file.
> 
> Could you tell us the situation what you are thinking more detail? Thank you!

`/proc/self ` is a symblic to process's own directory in /proc in container's namespace if process is running in a container, which leading an incorrect result. If want parse partitions on host's mount namespace, we should refer to `${HOST_PROC}/1 ` which means the initial process on host, HOST_PROC is leading path mapped in container.

Start a container named test, and map /proc to /host/proc in container, `docker run --rm -it --name test -v /proc:/host/proc alpine:3.14 sh`.

And execute command `docker exec test cat /host/proc/self/mountinfo` will get an incorrect results as:
image

And execute command `docker exec test cat /host/proc/1/mountinfo` will get result as follows:
image

shirou:
Thank you for the info. By using your code, a container with `-v /proc:/host/proc`, I can understand what you want to say.

Below is a command history from inside a container.

```
##### /proc/1
/host/proc # sort /proc/1/mountinfo |tail -1
998 968 8:2 /var/lib/docker/containers/8ebef3985c8138813864c63344b4c09a6fc690ad6fe00dad525c18d27b22daea/hosts /etc/hosts rw,relatime - ext4 /dev/sda2 rw

##### /proc/self. same as /proc/1
/host/proc # sort /proc/self/mountinfo |tail -1
998 968 8:2 /var/lib/docker/containers/8ebef3985c8138813864c63344b4c09a6fc690ad6fe00dad525c18d27b22daea/hosts /etc/hosts rw,relatime - ext4 /dev/sda2 rw

##### /host/proc/1. get "host's" proc
/host/proc # sort /host/proc/1/mountinfo |tail -1
999 28 0:4 net:[4026532474] /run/docker/netns/2c82aa5474a8 rw shared:363 - nsfs nsfs rw

##### /host/proc/self. can not get host's proc. it is same as "proc" in the container
/host/proc # sort /host/proc/self/mountinfo |tail -1
998 968 8:2 /var/lib/docker/containers/8ebef3985c8138813864c63344b4c09a6fc690ad6fe00dad525c18d27b22daea/hosts /etc/hosts rw,relatime - ext4 /dev/sda2 rw
```

The problem is, `pid` in "process's own /proc/[pid]" is different between host and a container inside. `pid` in a container inside, it refer to container's pid ofcourse. so "self" is resolved to *container's* directory. But we want to get mountinfo from `mounted /proc`. So this PR is correct.

tmm1:
This change broke disk methods on Android, where there is no permission to read pid 1:

```
$ cat /proc/1/mountinfo
cat: /proc/1/mountinfo: Permission denied
```

Using /proc/self still works.

Issue #1078, opened 2021-05-29
Plan to stop v2, target is 2021-11-30


shirou:
**Is your feature request related to a problem? Please describe.**

This is a feature request.

**Describe the solution you'd like**

Currently gopsutil have two versions, `v2` and `v3`. But it is difficult to support both versions from our limited resource. I want to terminate v2 and move v3 version to root directory.

My plan is 6 months later, end of November of this year, this means `2021-11-30`. After that day, we will not have v2 any more. Please migrate your code from v2 to v3 until that day.

**Plan**

- [ ] Oct.: check difference between v2 and v3. If there is something which v3 doesn't have, open a PR.
- [ ] Beginning of Nov.: Open a PR which move v3 to root, and check feasibility.
- [ ] End of Nov.: merge the PR.

Lomanic:
Do you plan on removing v2 files or just freeze the sub-packages? I fear that that removing files would break a lot of things downstream.

Otherwise, I agree with the idea of freezing v2.

shirou:
I am planning remove v2 files and move v3 to top. I understand it breaks a lot, but it's something we are going to have to do at some point.

Lomanic:
Closing with #1174 merged

PR #1137, opened 2021-09-29
Rename `readdrivestat` function on v3

### LINKED ISSUE #1136: Version 2 and 3 of the library should have different symbol names
**Describe the bug**

gopsutil provides a version 2 and a version 3 of the library, both of which use cgo. If the same project has both version 2 and 3 on their dependency tree, this causes build issues on macOS (and possibly other platforms).

**To Reproduce**

A minimal example is the following program

```go
package main

import (
	_ "github.com/shirou/gopsutil/disk"
	_ "github.com/shirou/gopsutil/v3/disk"
)

func main() {}
```

with `go.mod`:

```
module example.com

go 1.17

require (
	github.com/shirou/gopsutil v3.21.8+incompatible
	github.com/shirou/gopsutil/v3 v3.21.8
)

require (
	github.com/StackExchange/wmi v1.2.1 // indirect
	github.com/go-ole/go-ole v1.2.5 // indirect
	golang.org/x/sys v0.0.0-20210816074244-15123e1e1f71 // indirect
)
```

Copy both files to a directory and run `go build`. This will cause a linker error:

```
❯ go build
# example.com
/usr/local/Cellar/go/1.17.1/libexec/pkg/tool/darwin_amd64/link: running clang failed: exit status 1
duplicate symbol '_readdrivestat' in:
    /var/folders/6d/hrpbskwn3nv8qnm51fdpmh5c0000gn/T/go-link-455232793/000002.o
    /var/folders/6d/hrpbskwn3nv8qnm51fdpmh5c0000gn/T/go-link-455232793/000005.o
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

While no one would directly import both library versions, it is possible for both library versions to be on the dependency tree at the same time (this is what happens on open-telemetry/opentelemetry-collector-contrib#5425).

**Expected behavior**

The program should build correctly.

 - [x] Mac OS: [paste the result of `sw_vers` and `uname -a`]

```
❯ sw_vers
ProductName:    macOS
ProductVersion: 11.6
BuildVersion:   20G165

❯ uname -a
Darwin Pablos-MacBook-Pro.local 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64 i386 MacBookPro15,2 Darwin
```

**Additional context**

This happens because both v2 and v3 define `readdrivestat` on `gopsutil/disk/iostat_darwin.h` and `gopsutil/v3/disk/iostat_darwin.h`. Renaming symbols on one of the versions solves the issue for me.

mx-psi:
- Rename `readdrivestat` function on v3 to `v3readdrivestat`
- Delete unused `extern` declaration `readcpustat`.

This fixes #1136.

Signed-off-by: Pablo Baeyens <pablo.baeyens@datadoghq.com>

### DIFF
--- a/v3/disk/iostat_darwin.h
+++ b/v3/disk/iostat_darwin.h
@@ -29,5 +29,4 @@ struct CPUStats {
 	natural_t idle;
 };
 
-extern int readdrivestat(DriveStats a[], int n);
-extern int readcpustat(CPUStats *cpu);
+extern int v3readdrivestat(DriveStats a[], int n);
--- a/v3/disk/disk_darwin_cgo.go
+++ b/v3/disk/disk_darwin_cgo.go
@@ -19,7 +19,7 @@ import (
 
 func IOCountersWithContext(ctx context.Context, names ...string) (map[string]IOCountersStat, error) {
 	var buf [C.NDRIVE]C.DriveStats
-	n, err := C.readdrivestat(&buf[0], C.int(len(buf)))
+	n, err := C.v3readdrivestat(&buf[0], C.int(len(buf)))
 	if err != nil {
 		return nil, err
 	}
--- a/v3/disk/iostat_darwin.c
+++ b/v3/disk/iostat_darwin.c
@@ -16,7 +16,7 @@ static int getdrivestat(io_registry_entry_t d, DriveStats *stat);
 static int fillstat(io_registry_entry_t d, DriveStats *stat);
 
 int
-readdrivestat(DriveStats a[], int n)
+v3readdrivestat(DriveStats a[], int n)
 {
 	mach_port_t port;
 	CFMutableDictionaryRef match;


v3/disk/iostat_darwin.h
```diff
@@ -29,5 +29,4 @@ struct CPUStats {
 	natural_t idle;
 };
 
-extern int readdrivestat(DriveStats a[], int n);
-extern int readcpustat(CPUStats *cpu);
+extern int v3readdrivestat(DriveStats a[], int n);
```
shirou, line 32: memo: `readcpustat` is removed, but it is not used.

shirou (approved):
Good catch! Thank you so much. By the way, we have a plan to terminate v2 (#1078). (yes, this PR is important because after terminated, some user want to use v2)

Thank you so much!

v3/disk/iostat_darwin.h
mx-psi: That's right, I should have mentioned it :)

PR #1145, opened 2021-10-07
[v3] update x/sys version and fix v2 process kill

shirou:
To fix flaky tests, this PR updates `x/sys` according to [this issue](https://github.com/golang/go/issues/46763)

- update x/sys
- fix v2 process/process_windows.go `Kill()` used `os.Process` directly, not using `os.FindProcess` .


shirou:
hmm, no luck.

PR #1154, opened 2021-10-11
implement the 'OpenFilesWithContext' function of the windows system

kestrelcjx:


kestrelcjx:
implement the 'OpenFilesWithContext' function of the windows system

shirou:
Thank you for awesome PR!

However, on my environment, Windows 10 + golang 1.17.1, the Test_OpenFiles test failed on both v2 and v3.

```
--- FAIL: Test_OpenFiles (0.07s)
    process_test.go:670:
                Error Trace:    process_test.go:670
                Error:          Should NOT be empty, but was []
                Test:           Test_OpenFiles
```

I have not dig deeper, but there is no error but only one handle related to "process" directory file name comes [here](https://github.com/shirou/gopsutil/pull/1154/files#diff-ae26c3608d9221e521907319158f351ef1a80515c3dc5b2e780c5eda6758f5a8R666), and because this is directory, no handler fileName is added, then, returned empty. What environment have you tested?

kestrelcjx:
I forgot to run "Test_OpenFiles". When I run it, I get the same result as you, but if I add "f, err := os.Create("hello.log")", the test will pass. I don’t know why "test always open files."

```golang
func Test_OpenFiles(t *testing.T) {
	f, _ := os.Create("hello.log")
	pid := os.Getpid()
	p, err := NewProcess(int32(pid))
	skipIfNotImplementedErr(t, err)
	assert.Nil(t, err)

	v, err := p.OpenFiles()
	skipIfNotImplementedErr(t, err)
	assert.Nil(t, err)
	assert.NotEmpty(t, v) // test always open files.

	for _, vv := range v {
		assert.NotEqual(t, "", vv.Path)
	}
	f.Close()
}
```

PR #1158, opened 2021-10-20
fix: windows arm64 support

JanDeDobbeleer:
The title says it all, proven via [oh-my-posh](https://github.com/JanDeDobbeleer/oh-my-posh)

### DIFF
--- a/process/process_windows_arm64.go
+++ b/process/process_windows_arm64.go
@@ -1,0 +1,104 @@
+//go:build windows && arm64
+// +build windows,arm64
+
+package process
+
+import (
+	"errors"
+	"syscall"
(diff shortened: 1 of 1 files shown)


shirou:
@JanDeDobbeleer Thank you so much! I don't have a ARM windows so I can not confirm this, but it seems awesome.

Perhaps, #1157 (Thank you @acabarbaye !) also fixes the issue. But it seems merging #1157 is harmless. Then, I will merge both. Great contributions!

JanDeDobbeleer:
@shirou you're welcome, I should've contributed sooner as I fixed this about a month ago for oh-my-posh support on Windows arm64 😓

PR #1172, opened 2021-11-04
[disk] Modify stat free, from bavail to bfree

secake:
```
From:
Free:        (uint64(stat.Bavail) * uint64(bsize)), 
To:
Free:        (uint64(stat.Bfree) * uint64(bsize)),
```

https://pkg.go.dev/golang.org/x/sys/unix#Statfs_t
```
type Statfs_t
type Statfs_t struct {
	Type    int64
	Bsize   int64
	Blocks  uint64
	Bfree   uint64   /* free blocks in fs */
	Bavail  uint64   /* free blocks avail to non-superuser */
	Files   uint64
	Ffree   uint64
	Fsid    Fsid
	Namelen int64
	Frsize  int64
	Flags   int64
	Spare   [4]int64
}
```
In some cases, the old implementation may have Total-Free!=Used。

### DIFF
--- a/disk/disk_openbsd.go
+++ b/disk/disk_openbsd.go
@@ -132,7 +132,7 @@ func UsageWithContext(ctx context.Context, path string) (*UsageStat, error) {
 		Path:        path,
 		Fstype:      getFsType(stat),
 		Total:       (uint64(stat.F_blocks) * uint64(bsize)),
-		Free:        (uint64(stat.F_bavail) * uint64(bsize)),
+		Free:        (uint64(stat.F_bfree) * uint64(bsize)),
 		InodesTotal: (uint64(stat.F_files)),
 		InodesFree:  (uint64(stat.F_ffree)),
 	}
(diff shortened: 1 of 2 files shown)


shirou:
According to the [psutil implementation](https://github.com/giampaolo/psutil/blob/master/psutil/_psposix.py#L192), `Free` is a "avail_to_user", not a "avail_to_root". Could you tell us what Linux kernel version think `Total-Free! = Used`?

secake:
> According to the [psutil implementation](https://github.com/giampaolo/psutil/blob/master/psutil/_psposix.py#L192), `Free` is a "avail_to_user", not a "avail_to_root". Could you tell us what Linux kernel version think `Total-Free! = Used`?

Thanks for your replay！ 
I encountered this problem in ‘CentOS Linux release 7.7.1908 (Core)’.
I think `Total-Free! = Used` may be caused by the following reasons ：total is only available to root, free is avail_to_user, not avail_to_root, used is (total-avail_to_root), so `total - free != uesd`。  
But I don't know if it is intentional to do so. 
Because the [github.com/giampaolo/psutil](https://github.com/giampaolo/psutil/blob/master/README.rst#disks) readme example is also `Total-Free! = Used`.

secake:
According to the [psutil implementation line 190](https://github.com/giampaolo/psutil/blob/master/psutil/_psposix.py#L190), [psutil implementation line 194](https://github.com/giampaolo/psutil/blob/master/psutil/_psposix.py#L194), Free is a "avail_to_root", not a "avail_to_user".  the purpose of "avail_to_user"  is to calculate "usage_percent_user".

PR #1177, opened 2021-11-12
add initial aix support

chbuescher:
aix support using the perfstat library

### DIFF
--- a/v3/load/load_aix.go
+++ b/v3/load/load_aix.go
@@ -1,0 +1,71 @@
+// +build aix
+
+package load
+
+/*
+#cgo LDFLAGS: -L/usr/lib -lperfstat
+
+#include <libperfstat.h>
+#include <procinfo.h>
+*/
+import "C"
+
+import (
+        "context"
+        "unsafe"
+
+        "github.com/power-devops/perfstat"
+)
+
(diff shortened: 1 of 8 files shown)


shirou (approved):

shirou:
Since I don't have AIX server, I can not confirm this PR works or not. But it looks great. Thank you so much!

v3/load/load_aix.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Misc returns miscellaneous host-wide statistics.
+// darwin use ps command to get process running/blocked count.
+// Almost same as Darwin implementation, but state is different.
+func Misc() (*MiscStat, error) {
+        return MiscWithContext(context.Background())
+}
+
+func MiscWithContext(ctx context.Context) (*MiscStat, error) {
+		info := C.struct_procentry64{}
```
aklyachkin, line 46: Just as a side notice. In my version for AIX I do it a little bit another way:

```
        psinfo, err := perfstat.ProcessStat()
        if err != nil {
                return nil, err
        }
        ret := &MiscStat{
                ProcsTotal: len(psinfo),
        }
        return ret, nil
```

PR #1178, opened 2021-11-12
Check if path exists and is nonempty before reading host files

ISauve:
When we mount the required `/etc` files for host detection to a container, whenever a file isn't present on the host we end up creating an empty file in that location in the container. This can break host OS detection, so we need to make sure that when we check if a file exists, we also check that it contains data.


shirou:
Great suggestion. Can I create a PR based on this PR, or reopen this?

PR #1179, opened 2021-11-12
Trim quotes from /etc/os-release

lebauce:


shirou:
Thank you. Can I create a PR based on this commit 1e2204a98d89f1176f0fda6a54cc937c97dc550f ? ( I think using [strings.Trim](https://pkg.go.dev/strings#Trim) is better, though)

shirou:
oh, [here](https://github.com/DataDog/gopsutil/pull/37) it is.

PR #1180, opened 2021-11-15
mem: use unix.SysctlUint64 for hw.memsize on darwin

tklauser:
Use unix.SysctlUint64 which directly returns an uint64 rather than
converting it from a string.


shirou (approved):
Indeed. Thank you so much!!

PR #1181, opened 2021-11-16
deps: switch to recommended fork of StackExchange/wmi

mikemorris:
The README at https://github.com/StackExchange/wmi notes that this library is no longer actively maintained, and recommends switching to the fork at https://github.com/yusufpapurcu/wmi.

The primary change so far in the fork has been bumping https://github.com/go-ole/go-ole to v1.2.6 to fix Windows ARM and ARM64 compatibility, so this may not strictly be necessary since https://github.com/shirou/gopsutil/pull/1145 did bump that transitive dependency (and updating to a version of gopsutil with that PR included did resolve the build errors I was seeing in my project).

I wasn't able to update `Gopkg.lock` as I'm having some issues getting the `dep` build system installed with Go 1.17, would appreciate a maintainer adding a commit to fixup the lockfile.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,14 +2,6 @@
 
 
 [[projects]]
-  digest = "1:3a991e1f8ab48d1445b5106b628609285cecb1bfe5abdd078b4320d32ec0f3ee"
-  name = "github.com/StackExchange/wmi"
-  packages = ["."]
-  pruneopts = "UT"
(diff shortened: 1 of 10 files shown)


shirou (approved):
Thank you for your PR! I have updated to remove all StackExchange from import, update Gopkg.lock and modules on v3.

PR #1182, opened 2021-11-20
fix(process): fix the bug that the program is hung when getting the f…

kestrelcjx:
psutil(https://psutil.readthedocs.io/en/latest/index.html?highlight=oepn_files#psutil.Process.open_files): Warning on Windows this method is not reliable due to some limitations of the underlying Windows API which may hang when retrieving certain file handles. In order to work around that psutil spawns a thread to determine the file handle name and kills it if it’s not responding after 100ms. That implies that this method on Windows is not guaranteed to enumerate all regular file handles (see issue 597). Tools like ProcessHacker has the same limitation.

I tried it, and when I called the OpenFile method to the graphics driver, the program was suspended. So, I use the same method as psutil to deal with it.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -12,6 +12,7 @@ import (
 	"reflect"
 	"strings"
 	"syscall"
+	"time"
 	"unicode/utf16"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


shirou (commented):
Thank you for your cool PR! These changes looks good to me. 

How about to add "case <- ctx.Done()" in `select` ? It may takes a time if many processes exists. Canceable by context is better in this situation, I think.

shirou (approved):
Thank you for the quick update! it looks great and I have confirmed it works on my environment.

Issue #1171, opened 2021-11-01
[v3]: Process State "D" Uninterruptible sleep


bogdandrutu:
**Describe the bug**
When a process in "D" (Uninterruptible sleep) state is found on linux the return value is a string array with an empty string in it.

bogdandrutu:
On darwin seems to be "U" the state.

```
$ ps -axo state | grep U 
U   
```

bogdandrutu:
I think this is resolved now. Thanks

PR #1185, opened 2021-11-22
Add additional process states

tbarker25:
In the move from v1 to v3, process states has changed from returning a single character for status to an enum. https://github.com/open-telemetry/opentelemetry-collector was depending on some of these states.

I've looked through the documentation for most platforms and ensured that we are capturing all the cases (FreeBSD, OpenBSD, Linux, Solaris, Darwin).

See tracking issue: #1171

### DIFF
--- a/v3/process/process.go
+++ b/v3/process/process.go
@@ -45,13 +45,30 @@ type Process struct {
 
 // Process status
 const (
+	// Running marks a task a running or runnable (on the run queue)
 	Running = "running"
-	Sleep   = "sleep"
-	Stop    = "stop"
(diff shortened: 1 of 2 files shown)


shirou (approved):
Great! Unfortunately, this function is used by only Linux and Darwin. Others like [FreeBSD converts from kproc directly](https://github.com/shirou/gopsutil/blob/master/v3/process/process_freebsd.go#L126). But this PR provides room for extensions to FreeBSD and OpenBSD, and does not break compatibility. So LGTM! Thank you so much!!

PR #1186, opened 2021-11-23
v3/process: remove unused .cgo file

tklauser:
The process_darwin_amd64.go.cgo file was probably added by mistake. The
go tool does not recognize the .cgo extension, so it's already excluded
from the build. The real type and const definitions are in
process_darwin_amd64.go already.


shirou (approved):
Thank you so much! perhaps I made a mistake while migrating.

PR #1194, opened 2021-12-04
Enable errorlint and gci linters

mmorel-35:


shirou (approved):
`errors.Is` is introduced Go 1.13 but did not adapted yet. But now, we can fix those. Thank you so much!

PR #809, opened 2019-12-26
WIP: [net][linux]: add netlink implementation for Linux connection stat

shirou:
This PR introduces netlink support for Linux Connections.
To use netlink, users should add build tag `-tags=netlink` 

### Limitation for using netlink

- fd is always 0
  - to get fd becomes slow

### TODO

- UNIX Domain
- benchmark

PR #1199, opened 2021-12-11
add ConnectionsOptions api for walker callback instead of large slice return (only support in linux)

bingoohuang:
When there's amount of TIME_WAIT in the system ( when keep-alive does not work), for example, more than 10K, the memory used for the existing `Connections` will return a slice which has more than 10K elements and then use and free it.
A new ConnectionsOptions api which can support a walker callback func as an option to avoid the large slice return.

a memory usage difference can be checked in the following charts:

image

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -42,6 +42,59 @@ type ConnectionStat struct {
 	Pid    int32   `json:"pid"`
 }
 
+type ConnectionStatOptions struct {
+	Context  context.Context
+	Kind     string
+	Pid      int32
+	Max      int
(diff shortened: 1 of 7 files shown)


shirou:
Thank you for your contribution. Perhaps this is related to #809.

I think this "Functional Options Pattern" is a solution of this bottle neck on Linux. But this kind of change may break compatibility. I want to check this PR carefully, but I have no so many time to year-end. plz wait for a while. (or I hope any other contributor may check and add +1)

PR #1200, opened 2021-12-15
Add drivetype for windows

jasonwbarnett:
This adds a drive type to windows so you can easily discern what type of drive it is.


shirou:
From my understanding, the "Drive Type" in Windows indicates what the drive is fixed, cdrom, or else. And other platform does not have a same information. And also at least [psutil does not have this information](https://psutil.readthedocs.io/en/latest/#disks). So I don't want to add platform dependent information.

jasonwbarnett:
@shirou what about adding an `platformExtras` field that is platform specific? Would that be considered?

shirou:
We don't want to add more platform specific API on this library. How about to create a new library which can get drive information on Windows specific?

PR #1204, opened 2021-12-21
fix unstable boot time because of float conversion

YangKeao:
Signed-off-by: YangKeao <yangkeao@chunibyo.icu>

The original calculation of create time inside a container may cause unstable result:

```go
b, err := strconv.ParseFloat(f[0], 64)
if err != nil {
    return 0, err
}
t := uint64(time.Now().Unix()) - uint64(b)
```

The `b` is a float number, converting it to a integer will discard the fraction part. Considering following situation:

```
uptime: 5.5
now: 6.9
```

Then, the boot time is `1`. After 0.1 seconds, the uptime will be `5.6` and the `now` is `7.0`, which will give us `2` as the boot time. Converting after subtraction will solve this problem.

This issue will also affect the `createTime` of a process.

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package common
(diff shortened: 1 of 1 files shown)


shirou:
So you mean there are 2 issues.
1. `time.Now().Unix()` has only "second" precision
2. Casting to `uint64` from float value discards fraction part.

I agree to those. Thank you for your great PR!

PR #1210, opened 2021-12-29
enable contextcheck linter

mmorel-35:
I didn't know this linter exists. It has restored the context as a parameter for some functions and add it to some.


shirou (approved):
Sorry for late response. This seems great. Thank you for your contribution!

PR #1217, opened 2022-01-04
[disk][freebsd] implement SerialNumberWithContext

scop:


shirou:
Thank you!

From my FreeBSD 12.3 on VPS, `geom` returns `(null)` on `ident`.
```
Geom name: vtbd0
Providers:
1. Name: vtbd0
   Mediasize: 214748364800 (200G)
   Sectorsize: 512
   Mode: r2w2e3
   descr: (null)
   ident: (null)
   rotationrate: unknown
   fwsectors: 63
   fwheads: 16
```

Sorry, I don't know what kind of string geom returns other than "null", but could you add a check and return empty if at least it is `(null)` ?

scop:
Ah, good catch. Searching the web for "geom disk list ident" gives a bunch of hits with example outputs containing serial numbers (and some `(null)`s as well).

scop:
Not sure what to say about the codecov result, this touches FreeBSD specific code only, resulting somehow Linux coverage increasing and Windows and macOS decreasing... :thinking:

shirou (approved):
Thank you so much! It works perfect on my FreeBSD 12.3 machine.

PR #1225, opened 2022-01-12
Make createTimeWithContext faster on MacOS

PierreF:
`Processes()` call is really slow on MacOS (more than 1 seconds on my system) because it calls `createTimeWithContext` for each PID which is slow (due to running "ps" command).

This PR make use of sysctl interface to speed-up this function. This PR introduce a slight change: before the createTime was a rounded number of seconds, now it will include milliseconds.

We could easily revert to returning createTime rounded to seconds and in this case both code return the same value.

The choice to use sysctl interface come from psutil: https://github.com/giampaolo/psutil/blob/18348c9b6b66ff12abbbfeaebbe2218cdc6b5556/psutil/_psutil_osx.c#L257 and https://github.com/giampaolo/psutil/blob/18348c9b6b66ff12abbbfeaebbe2218cdc6b5556/psutil/arch/osx/process_info.c#L347

The KinfoProc may contains more (static) information like PPID which could be cached to avoid futher calls. I can create another PR (or update this one) to cache all static information available.

Step to repoduce the "slow" issue:

```
$ cat test_perf.go
package main

import (
	"log"
	"time"

)

func main() {
	start := time.Now()

	if err != nil {
		log.Fatal(err)
	}

	log.Printf("had %d processes. Listed in %v", len(procs), time.Since(start))
}

$ uname -a
Darwin MBP-de-pierre.bleemeo.work 21.2.0 Darwin Kernel Version 21.2.0: Sun Nov 28 20:28:41 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T6000 arm64

$ git checkout master
$ go run test_test.go
2022/01/12 16:29:39 had 696 processes. Listed in 1.327848292s

$ git checkout fast-process-list-on-macos
$ go run test_test.go
2022/01/12 16:29:40 had 697 processes. Listed in 29.70225ms
```

So with master, it took 1.3 seconds to call `Processes()` and just 30 milliseconds with this PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,5 +9,5 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/tklauser/go-sysconf v0.3.9
 	github.com/yusufpapurcu/wmi v1.2.2
-	golang.org/x/sys v0.0.0-20211013075003-97ac67df715c
+	golang.org/x/sys v0.0.0-20220111092808-5a964db01320
 )
--- a/go.sum
+++ b/go.sum
@@ -26,8 +26,8 @@ github.com/yusufpapurcu/wmi v1.2.2/go.mod h1:SBZ9tNy3G9/m5Oi98Zks0QjeHVDvuK0qfxQ
 golang.org/x/sys v0.0.0-20190916202348-b4ddaad3f8a3/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20201204225414-ed752295db88/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20210816074244-15123e1e1f71/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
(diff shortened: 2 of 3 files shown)


PierreF:
I've just seen we already get the KprocInfo with `getKProc()` method on Process. I'll use `getKProc()`

Do you want me to pass though existing `getKProc()` (unmodified) or change `getKProc()` to use `SysctlKinfoProc` from 
golang.org/x/sys ?

Lomanic:
Thanks a lot for this change, `unix.SysctlKinfoProc()`  was added to x/sys/unix in [`95da234e122c5f925d4f955e47843f027369c273`](https://go.googlesource.com/sys/+/95da234e122c5f925d4f955e47843f027369c273) two months ago.

> The KinfoProc may contains more (static) information like PPID which could be cached to avoid futher calls.

Ppid are not static on unix, orphan/zombie processes can get attached to pid 1 (init) if their parent is killed. [Process ppid is only cached and cacheable on windows](https://github.com/shirou/gopsutil/blob/1ccce7b91c1903b0e280547bb1b1a32aa815fc28/process/process_windows.go#L306).

> Do you want me to pass though existing getKProc() (unmodified) or change getKProc() to use SysctlKinfoProc from
golang.org/x/sys ?

Let's modify getKProc to use the `unix.SysctlKinfoProc()` function indeed.

I don't have any darwin host or VM these days so won't be able to test any of this.

PierreF:
Indeed PPID can't be cached. Missed this one.

Superseeded this PR by #1226 which use KProc even for more thing that just createTime

PR #1226, opened 2022-01-13
Avoid ps command and use KProc on MacOS

PierreF:
Follow up on #1225

I tried to avoid all call to "ps" on MacOS by using fields from KProc structure.

I've seen the following changes in value returned:
* The create time now use milliseconds precision.
* Parents() works much better. It used to fail quite often. Now it always return the Process associated with PPID.
* GIDs works better
   * it used to return mostly `0, 0, 0` and few times `0, $GID, 0`
   * now it return `$GID, $GID, $GID` (actually "real GID", "real GID", "saved effective GID"), but all my processes used the same value (according to "ps -o gid,rgid,svgid").

I'll try to push another PR based on this one, that try to remove even more call to ps command, by using libproc.h. But since this will introduce CGO dependency, I will kept old implementation in nocgo file.

The speed-up gain (tested using https://gist.github.com/PierreF/19a7ec569d1486d2f87ce4b0d5245e4b which I also used to validate the absence of changes):
```
$ git checkout darwin-drop-callps
$ go run cmpversion.go
2022/01/13 11:56:01 listed in 3.857042ms
2022/01/13 11:56:02 total time: 1.138989334s with 0 errors (and 1 err on Parent()) and 466 PIDs (466 without error)

$ git checkout master
$ go run cmpversion.go
2022/01/13 11:56:03 listed in 814.121459ms
2022/01/13 11:56:08 total time: 5.748625959s with 0 errors (and 209 err on Parent()) and 465 PIDs (465 without error)
```


PierreF:
I've dropped one test on Darwin on Pids(): now that we no longer parse ps output to get list of PIDs and use unix.SysctlKinfoProc() the test would ends testing unix.SysctlKinfoProc().

PierreF:
The continuation (or remplacment) of this PR in #1229. PR #1229 contains all change from this PR and more.

So if you want to review it as two smaller part, I kept this one open. If you prefer reviewing all change in one, I can close this PR.

PR #1229, opened 2022-01-13
Darwin drop callps step2

PierreF:
Continuation of #1226. This PR include #1226 (so we could merge this one directly and ignoring the first one).

Using libproc (CGO) we can speed-up retrieving information of most process. Some (likely all that belong to another user) fail with EPERM. For them I do fallback on using ps.

On my Mac, I got close to 5x faster run time: 3 seconds for 732 processes vs 14.5 seconds on master.

I've used https://gist.github.com/PierreF/d48d6c56bf5f06b471362d06eafe0111 to test performance and difference between this PR and master.

I didn't see significant difference. Most difference are either:
* fixes of existings bugs (like GIDs missing, cf comment in #1226)
* Precision changes (like in createtime that now includes milliseconds, or user/system time that are no longer truncated to 1/100th of seconds).
* Real change, since the two program don't run at the exact same time.

My system is a Mac M1 (arm64) using MacOS 12. I'll try it on a Mac Intel tomorow.


Lomanic:
Please squash the last two commits in one.

Could you be more explicit with your 3rd commit message? (at least mention the process package and darwin).

> Some (likely all that belong to another user) fail with EPERM. For them I do fallback on using ps.

That feels really hacky and we don't do the same for linux. Users should run their binary as setuid 0 (like ps) if they want to look at foreign programs.

PierreF:
Is there something I can do to help this PR getting merged ?

I should have fixed your comments, or is there something missing ?

shirou:
Since I don't have a mac, I can not confirm this PR works or not, and no one reported except noble kind @Lomanic (but I think he also does not have a real mac).

I will borrow a intel mac from my friend in this weekend.

shirou (approved):
I confirmed this PR works on my friend's mac, Big Sur Intel i7. All tests in process by `go test -v` have been become faster from 12sec to 3 sec. Great improvement! Thank you so much and sorry to late to review.

PR #1233, opened 2022-01-19
Let almalinux be part of the rhel family

Kogelvis:
As almalinux is also one of the redhat based OS'es, let it also be part of the rhel family.


shirou (approved):
Thank you!
In [almalinux](https://almalinux.org/) site, " AlmaLinux OS is 1:1 binary compatible with RHEL® and pre-Stream CentOS" so I agree with this is one of a RHEL family.

PR #1235, opened 2022-01-22
[process] implement ParentWithContext using PpidWithContext

scop:
Removes need for redundant ParentWithContext implementations. It had led to it being unsupported on FreeBSD and OpenBSD even though PpidWithContext was available for them, and different implementations for getting the parent info used in ParentWithContext and PpidWithContext on Darwin and Linux.

For Darwin and Linux, I just dropped the separate ParentWithContext implementation. Let me know if you would have preferred the internals of that to be moved over to PpidWithContext instead.


shirou (approved):

PR #1238, opened 2022-01-30
Spelling and grammar fixes

scop:


shirou (approved):
Ah, great catch! My English is still in its infancy...

PR #1242, opened 2022-02-05
[host][solaris] implement SensorsTemperaturesWithContext

scop:


scop:
This touches only a Solaris-only file, assuming the Windows CI failures are unrelated.

shirou (approved):
I haven't confirm but it seems no problem. Thank you so much!

PR #1243, opened 2022-02-05
FreeBSD sensors support

scop:
Includes invoker enhancement to make it possible to pass command env in the context, as the env affects sysctl decimal formatting here. Not that nice to hide this in context, but ~~

### DIFF
--- a/host/host_freebsd.go
+++ b/host/host_freebsd.go
@@ -4,12 +4,15 @@
 package host
 
 import (
+	"bufio"
 	"bytes"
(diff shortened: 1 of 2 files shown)


henrygd:
I'm curious if there was a reason why this wasn't merged at the time? Maybe the env handling?

Would there be any interest from maintainers in an updated PR to support FreeBSD sensors?

shirou:
@henrygd Of course! PRs are always welcome.

PR #1241, opened 2022-02-04
Fix broken retrieval of OpenBSD CPU metrics

chrissnell:
CPU metrics are broken on OpenBSD 7.0-CURRENT [as described in #1239](https://github.com/shirou/gopsutil/issues/1239).  I believe that the problem is the multiplication of the CPU incrementer by 2 when the code walks through the available CPUs to gather CPU time metrics.   The code makes uses of the KERN_CPTIME2 sysctl to fetch per-CPU time metrics. 

Looking at the kernel code behind this particular sysctl, [we have this.](https://github.com/openbsd/src/blob/4cd1828ed9eeebb6486d183eb7e4a99fc9464b70/sys/kern/kern_sysctl.c#L2420-L2443).  As you can see, it's calling a `CPU_INFO_FOREACH` macro and iterating through the results until it has the info for the requested CPU.  This macro is defined per-architecture and in every one that I found, it's just iterating through `cpu_info` structs [defined here](https://github.com/openbsd/src/blob/0a894fa656a20dcd3351d61ee4e14415e615ef83/sys/arch/amd64/include/cpu.h#L94).   

I can't find anything that leads me to believe that this Go code should be incrementing by anything other than 1.

In the entire OpenBSD codebase, there's only one use of the KERN_CPTIME2 sysctl and that's in `snmpd`'s source code, [which  iterates through the CPUs and it does this by incrementing the counter by 1](https://github.com/openbsd/src/blob/2207c4325726fdc5c4bcd0011af0fdf7d3dab137/usr.sbin/snmpd/timer.c#L63-L66).

This particular code ^ hasn't changed since 2016 and I'll bet it still works. :-)

Here's another clue.  As I mentioned, the guts of this syscall is just iterating through `cpu_info` struct.  I looked around for other uses of this struct and found code for the new OpenBSD debugger.   There are different implementations for different architectures, but for amd64, [it also increments through CPUs by 1.](https://github.com/openbsd/src/blob/5280e11119be648630599c383004733f6bed28e8/sys/arch/amd64/amd64/db_interface.c#L211)

Here is a PR to fix this.  I only have one OpenBSD machine to test on, a simple Dell Optiplex workstation.

PR #1244, opened 2022-02-06
fix cpu_openbsd.go once and for all

omar-polo:
Hello,

as reported in #1241 currently cpu_openbsd.go doesn't retrieve the correct values in all situations.  It's completely my fault for not implementing it correctly the first time, I was only thinking about amd64.

This is PR has a number of improvements, I suggest reading the diffs per-commit rather than a whole.

First of all, it drops support for OpenBSD <6.3, which was EOL'd more than three years ago!  This allows to semplify the code a bit and to "const-ify" some parameters.

The second commit is aestetic: it just fixes some typos.

The third commit changes the way we parse the data we get from `common.CallSyscall`.  Should make this library work on big endians.

The fourth commit is the real fix: it changes the way we obtain statistics from the CPU.  There were multiple things wrong there:

 - `KERN_CPTIME` returns an array of `long`s, while `KERN_CPTIME2` an array of uint64_t.  longs depends on the platform (on amd64 are 64bits, on i386 are 32bits).  
 - Also, instead of trying to guess which CPUs are online, use `KERN_CPUSTATS` that gives us the statistics and an handy flags field from which we can observe is the cpu is online or not.

I've tested this on amd64 (4 real cores, works correctly with both `hw.smt` enabled and not) and i386 (2 real cores, no smt), both running OpenBSD-CURRENT.

### DIFF
--- a/cpu/cpu_openbsd.go
+++ b/cpu/cpu_openbsd.go
@@ -4,17 +4,18 @@
 package cpu
 
 import (
-	"bytes"
 	"context"
-	"encoding/binary"
 	"fmt"
 	"runtime"
 	"syscall"
+	"unsafe"
 
 	"github.com/shirou/gopsutil/v3/internal/common"
 	"github.com/tklauser/go-sysconf"
 	"golang.org/x/sys/unix"
 )
+
+import "C"
 
 const (
 	// sys/sched.h
@@ -51,13 +52,8 @@ func smtEnabled() (bool, error) {
 		return false, err
 	}
 
-	var ret bool
-	br := bytes.NewReader(buf)
-	if err := binary.Read(br, binary.LittleEndian, &ret); err != nil {
-		return false, err
-	}
-
-	return ret, nil
+	smt := *(*uint32)(unsafe.Pointer(&buf[0]))
+	return smt == 1, nil
 }
 
 func Times(percpu bool) ([]TimesStat, error) {
@@ -89,7 +85,6 @@ func TimesWithContext(ctx context.Context, percpu bool) ([]TimesStat, error) {
 			j *= 2
 		}
 
-		cpuTimes := make([]int32, cpuStates)
 		var mib []int32
 		if percpu {
 			mib = []int32{ctlKern, kernCptime2, int32(j)}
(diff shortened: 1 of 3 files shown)


cpu/cpu_openbsd.go
```diff
@@ -45,99 +43,64 @@ func init() {
...
+func TimesWithContext(ctx context.Context, percpu bool) (ret []TimesStat, err error) {
+	if !percpu {
+		mib := []int32{ctlKern, kernCpTime}
 		buf, _, err := common.CallSyscall(mib)
 		if err != nil {
 			return ret, err
 		}
+		var x []C.long
+		// could use unsafe.Slice but it's only for go1.17+
+		x = (*[cpuStates]C.long)(unsafe.Pointer(&buf[0]))[:]
```
shirou, line 72: Sorry, It seems this PR uses CGO. we are aiming gopsutil for pure golang.

omar-polo: To be honest I was tempted to rewrite the whole thing in cgo for practical reason.  By design, OpenBSD doesn't have a strong ABI compatibility, things are changed, removed an added when needed.  (This not to say that OpenBSD breaks its ABI every day, but it's not uncommon.  See the removal for the support for <6.3 for example, where various constants changed value after `hw.smt` was added.)

What OpenBSD does is to provide *source* compatibility; that is you can always recompile the software after a change.  Rewriting the constants to use values pulled from C will mean that it would be almost no breakage of this library (and the software that depends on it) due to changes on base.

IIRC Go itself on OpenBSD switched to cgo by default a couple of years ago, as now it goes thru libc to do syscalls (the same happens on solaris too.)

anyway, in this pr i'm using cgo only to know the size of `long`, which varies between platform (and OpenBSD supports a huge amount of them.)  Usually it's 64 bit on 64 bit arches and 32 on 32 bit arches, but I'm not sure and I personally own only amd64 and i686.  What do we do in these cases?

omar-polo: personally I don't really like that approach -- limiting the support to i386 and amd64 for no real gain only to avoid cgo, when that's the default, the preferred way to get stuff done on this os and the only solution to avoid breakage in the future due to ABI changes.

I'll save this discussion for another thread thought, since it's the pattern already in use I'll update this pr to match existing practices in a couple of days :)

thanks

omar-polo:
OK, I've got rid of the cgo bits and polished it a bit.  I've tested on amd64 -CURRENT (with both smt enabled and disabled) and on i386 -CURRENT.

omar-polo:
P.S.: I *think* I could have used `uintptr` instead of the two `_386.go` and `_amd64.go` files, but I'm not sure if `sizeof(long)` == `size of uintptr` for all architectures, and I don't have an arm / aarch64 board at hand to verify, so I went for the additional files.

shirou (approved):
Thank you for your great contribution! I am very appreciate your continuous work.

omar-polo:
happy to have been useful and apologize again for my initial mistake :)

Cheers!

chrissnell-okta:
Can you cut a new tag on this repo so that I can try it with Telegraf?

chrissnell-okta:
> Can you cut a new tag on this repo so that I can try it with Telegraf?

Actually, I figured out how to build from this most recent commit and good news, telegraf works now!  Thanks @omar-polo 

@shirou still, if you can cut a release or a tag, I will pester the telegraf folks to update the dep.  Thanks!

shirou:
gopsutil has a [monthly release policy](https://github.com/shirou/gopsutil/blob/master/README.md?plain=1#L13). plz wait a few days.

PR #1246, opened 2022-02-08
Fix typing issues in createTimeWithContext on Freebsd/i386

skozlov404:
Before change:
```
$ GOOS=freebsd GOARCH=386 go build ./process
process\process_freebsd.go:118:26: cannot use k.Start.Sec * 1000 + k.Start.Usec / 1000 (type int32) as type int64 in return argument
```


shirou:
Thank you for reporting and this PR! We should add `GOARCH=386` in build_test job. (let me add in another PR)

PR #1251, opened 2022-02-18
Add support for reading smaps_rollup

rishabh-arya95:
Adding  support for reading memory  information for a process from smaps_rollup file which already contains the pre-summed information, will be more efficient than parsing the whole smaps file and summing up the values.

Reference: https://github.com/torvalds/linux/blob/master/Documentation/ABI/testing/procfs-smaps_rollup


shirou (approved):
> The format is almost identical to /proc/pid/smaps
and
> Additionally, the fields Pss_Anon, Pss_File and Pss_Shmem are not present in /proc/pid/smaps
but we are not using those values. so nothing change is required, just reduce parsing cost. I think it seems good PR, thank you so much!

PR #1254, opened 2022-02-22
[disk][freebsd] include serial number in IOCounters

scop:


shirou (approved):
Good catch! Thank you. (I agree about we can ignore the error return value in this case)

scop:
I also agree that it's ok to ignore the error here. But for full disclosure, doing that wasn't my invention, I just followed what was already being done in the corresponding Linux version.

PR #1256, opened 2022-02-22
[disk][solaris] implement SerialNumberWithContext

scop:


disk/disk_solaris.go
```diff
@@ -114,7 +116,35 @@ func UsageWithContext(ctx context.Context, path string) (*UsageStat, error) {
...
+				for i, f := range flds {
+					if i > 0 && i < flen-1 && f == "SN:" {
+						return flds[i+1], nil
+					}
+				}
+			}
+			return "", nil
+		}
+	}
+	if err = s.Err(); err != nil {
```
shirou, line 144: How about this?

```suggestion
	if err := s.Err(); err != nil {
```

scop: We already have `err` defined in scope here, and to me reusing it like in the original is cleaner than shadowing it with a new one, so I'd personally keep it as it is in my patch. But your call, do feel free to push the change if you like it better the way you suggested.

shirou (approved):
Thank you so much!

Issue #374, opened 2017-05-24
Find top processes by their resource consumption


mrLSD:
How to get top processes utilization CPU, Mem, Disk?
Currently via process.Pids() we can fetch all processes. Is it another way?

Additionally how to calculate CPU percentage for process?
For example memory: MemoryPercent()

miton18:
I try to get top 5 process too. Actually, I can get mem usage, but I can't find a good way for CPU usage, process.Percent(0) is equal to 0 because I call it in short times, and set a duration to 1 s for example, make me wait 1s * process count, and it's not possible....
```
for _, pid := range pids {
    prcs, err := processTool.NewProcess(pid)
    if err != nil {
	return err
    }
    ram, err := prcs.MemoryPercent()
    if err != nil {
	return err
    }
    cpu, err := prcs.Percent(0)
    if err != nil {
	return err
    }
    name, err := prcs.Name()
    if err != nil {
        return err
    }
}
```

Lomanic:
psutil doesn't include such kind of (specialized) helper neither. Here is a simple tutorial that shows how to get the 5 processes with highest memory consumption on the system with psutil. https://thispointer.com/python-get-list-of-all-running-processes-and-sort-by-highest-memory-usage/

PR #1265, opened 2022-03-02
Add process get pid tree

gekigek99:
I think it could be handy to be able to get the children tree pid of a specific process.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -518,6 +518,12 @@ func (p *Process) Children() ([]*Process, error) {
 	return p.ChildrenWithContext(context.Background())
 }
 
+// TreePid returns a slice of process tree pids
+// (together with original ppid).
+func (p *Process) TreePid() []int32 {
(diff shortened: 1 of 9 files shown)


Lomanic:
We already stated in the past that we didn't want to add helper functions to the (already vast) public API. https://github.com/shirou/gopsutil/issues/374#issuecomment-511950787

Plus this implementation is slow with a lot of unnecessary work, see https://github.com/giampaolo/psutil/blob/8091fa5999d2ac5c83e4aaf99a5560dc63dbee4b/scripts/pstree.py for a proper way of creating a process tree (a single loop to get each pid's parent). It's also needlessly duplicated on all platform-specific .go files when the function is the same everywhere.

gekigek99:
That python script gets all processes tho, my implementation was to retrieve the tree pid unders a specified process.

I wrote the same function in every os specific go file just because in process.go file, all functions are calling their counterpart with context.

PR #1267, opened 2022-03-04
fix(process): fix OpenFilesWithContext panic problem

hzhaop:
fix OpenFilesWithContext panic problem


shirou (approved):
Absolutely. Thank you so much!

PR #1268, opened 2022-03-04
refactor: remove unnecessary exec.LookPath calls

scop:
Executing the command does the lookup if needed and returns the same
error when not found, no need to do it separately.


shirou (approved):
I see. [Command uses LookPath internally](https://cs.opensource.google/go/go/+/refs/tags/go1.17.8:src/os/exec/exec.go;l=175), so I agree to this PR. Thank you!

PR #1269, opened 2022-03-04
[disk][linux] support new stats available with 4.18 and 5.5 kernels

scop:
https://www.kernel.org/doc/Documentation/ABI/testing/procfs-diskstats


shirou:
Thank you for your contribution. However, we don't want to add Linux specific information.

scop:
I don't think this is any more Linux specific than what we already have in `IOCountersStat`. Before this change, it currently has 11 counter-like members, only 6 of which are populated on Windows, 4 on OpenBSD, 7 on FreeBSD, etc.

Secondly, I see no reason why these members couldn't be populated on other OS's as well, if they provide such info -- to me discards and flushes seem fairly generic in nature, and not really something deeply implementation/OS specific.

shirou:
One more reason is [psutil doesn't have that fields](https://psutil.readthedocs.io/en/latest/#psutil.Process.io_counters).

scop:
To me not having something in psutil is not a reason not to improve gopsutil.

Issue #1159, opened 2021-10-21
[v3] Can not get mountinfo on android


shirou:
**Describe the bug**

#1133 changes getting mountinfo source from `/proc/self/mountinfo` to `/proc/1/mountinfo` in order to adapt container. However, from [this report](https://github.com/shirou/gopsutil/pull/1133#issuecomment-945991884) by @tmm1(thank you!) , Android can not read `/proc/1/mountinfo` because of `Permission denied`.

**Expected behavior**

success to get mountinfo.

**Environment (please complete the following information):**
 - [x] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]

android 

**Additional context**

Perhaps, we should add "fallback" to use `/proc/self/mountinfo` if error happened.

Lomanic:
#1133 broke more than just Android apps, but also apps running as unprivileged users on systems with procfs mounted with `hidepid=2`, see #1212.

cbos:
@shirou 

We face the same issue with Telegraf, see influxdata/telegraf#10357

We applied the CIS Hardening rules by running https://github.com/dev-sec/ansible-collection-hardening
Due to that hidepid is now applied. Since that moment the disk info about disk usage is not available anymore.
hidepid is a common security measure, see https://linux-audit.com/linux-system-hardening-adding-hidepid-to-proc/ for more info.

In the troubleshouting section of the disk plugin (https://github.com/influxdata/telegraf/tree/master/plugins/inputs/disk#troubleshooting) there is described to test this:
sudo -u telegraf cat /proc/self/mounts and that works fine.

So using /proc/self instead of /proc/1 would solve the problem.

As this project is used as upstream project, this block some functionality of Telegraf as well.
Hopefully it can be fixed soon.

shirou:
#1270 created. Could you check by using that PR?

cbos:
@shirou 
Thanks for fixing this. I quickly checked the code changes and that looks good to me as far as I can see.

@powersj 
What would be quick way I can validate this with telegraf?

powersj:
I ran:

```
GOPROXY=direct go get -u github.com/shirou/gopsutil/v3@feature/disk_add_fallback_to_self_mountinfo
```

Which upgraded the gopsutil dependency to the branch with the fix. I have put up a [PR here](https://github.com/influxdata/telegraf/pull/10785), which should have some artifacts attached to it shortly that you can then try out.

cbos:
@shirou @powersj 
[...]

PR #1272, opened 2022-03-09
[disk][linux] add HOST_PROC_MOUNTINFO

### LINKED ISSUE #1271: fallback mechanism for /proc/1/mountinfo to /proc/self/mountinfo can produce noisy logs
**Describe the bug**
The changes to `gopsutil` for reading `/proc/1/mountinfo` affected applications running under restricted environments that disallow access to `/proc/1/mountinfo` and the recent fix for #1159 (filed for android but other restricted environments are also affected (eg, snaps)) fixes application behavior to work under confinement (thanks!). However, depending on the system, the attempt to read `/proc/1/mountinfo` could cause a sandbox denial in the logs which can be quite noisy if using `gopsutil` as part of a monitoring solution that polls often.

I suggest introducing an environment variable to force the old behavior. Eg, use the current behavior on `master` but introduce a new `SELF_MOUNTINFO` environment variable to use `/proc/self/mountinfo` without first trying `/proc/1/mountinfo`. This allows people to set `SELF_MOUNTINFO=1` when `gopsutil` is running under these restricted environments.

I plan to submit a PR for this.

**Expected behavior**
`gopsutil` continues with current fallback behavior by default but allows reading only `/proc/self/mountinfo` so that `gopsutil` can function correctly without introducing noisy log entries.

```
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

$ uname -a
Linux parvati 5.4.0-100-generic #113-Ubuntu SMP Thu Feb 3 18:43:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

**Additional context**

jdstrand:
The changes to gopsutil for reading /proc/1/mountinfo affected applications running under restricted environments that disallows access to /proc/1/mountinfo. #1159 was filed for android but other restricted environments are also affected (eg, snaps)). The fix for #1159 addressed the application behavior to work under confinement for non-android as well. However, depending on the system, the attempt to read /proc/1/mountinfo could cause a sandbox denial in the logs which can be quite noisy if using gopsutil as part of a monitoring solution that polls often.

This introduces HOST_PROC_MOUNTINFO to force reading from the parent dir of the specified path instead of first trying /proc/1. When unset, retain the current behavior with fallback. This allows people, for example, to set HOST_PROC_MOUNTINFO=/proc/self/mountinfo when gopsutil is running under these restricted environments.

This change updates the private readMountFile() to use a root path instead of a root subpath, and adjusts PartitionsWithContext() to set the root path to /proc/1 initially and falling back to /proc/self. When HOST_PROC_MOUNTINFO is not empty, set the root path to the parent directory of HOST_PROC_MOUNTINFO.

Tested on linux with `HOST_PROC_MOUNTINFO` unset, set to /proc/self/mountinfo as well as other values and it behaves as expected (eg, when unset, has the current behavior with fallback, else tries the specified path without fallback). Also `make build_test` passed all tests.

NOTE: originally this used `SELF_MOUNTINFO`. Thanks to @shirou and @Lomanic for the feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,9 @@ environment variable.
 You can set an alternative location to `/dev` by setting the `HOST_DEV`
 environment variable.
 
+You can set an alternative location to `/proc/N/mountinfo` by setting the
+`HOST_PROC_MOUNTINFO` environment variable.
+
 ## Documentation
 
 see http://godoc.org/github.com/shirou/gopsutil
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -235,9 +235,10 @@ var fsTypeMap = map[int64]string{
 	ZFS_SUPER_MAGIC:             "zfs",                 /* 0x2FC12FC1 local */
 }
 
-// readMountFile reads mountinfo or mounts file under /proc/1 or /proc/self
+// readMountFile reads mountinfo or mounts file under the specified root path
+// (eg, /proc/1, /proc/self, etc)
 func readMountFile(root string) (lines []string, useMounts bool, filename string, err error) {
-	filename = common.HostProc(path.Join(root, "mountinfo"))
+	filename = path.Join(root, "mountinfo")
 	lines, err = common.ReadLines(filename)
 	if err != nil {
 		var pathErr *os.PathError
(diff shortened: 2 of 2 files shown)


shirou (approved):
Seems great. Thank you so much!

PR #1275, opened 2022-03-19
[mem][linux] added additional meminfo fields

quanweiZhou:
support HugePages_Rsvd, HugePages_Surp

Signed-off-by: quanweiZhou <quanweiZhou@linux.alibaba.com>


shirou (approved):
Thank you and sorry to late response. These are Linux specific values and we don't want to add such a platform specific information. However, `HugePagesTotal`, `HugePagesFree`, and `HugePageSize` are already added (in 2018). So, I don't like it very much, but I make a hard decision about adding those information to ensure consistency. Anyway, thank you for your contribution.

PR #1280, opened 2022-03-24
Fix telegraf 1.22.0 build on OpenBSD/arm64

geowiwi:
In v1.21.4, telegraf gained support for OpenBSD/arm64.
With v1.22.0 and the switch to gopsutil/v3 we now see:

../../../go/pkg/mod/github.com/shirou/gopsutil/v3@v3.22.2/cpu/cpu_openbsd.go:63:14: undefined: cpuTimes

PR adds support for OpenBSD/arm64 in v3/cpu.

### DIFF
--- a/cpu/cpu_openbsd_arm64.go
+++ b/cpu/cpu_openbsd_arm64.go
@@ -1,0 +1,10 @@
+package cpu
+
+type cpuTimes struct {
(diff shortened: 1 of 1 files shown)


shirou (approved):
Thank you!

PR #1281, opened 2022-03-28
Update meminfo fields

ronnocol:


shirou:
We don't want to add Linux specific information any more.

ronnocol:
> We don't want to add Linux specific information any more.

While I added several modern additions from /proc/meminfo; `AnonPages` has been around for a very very long time and is notably missing. Knowing the amount of anon memory in use is helpful in understanding the performance of many systems.

PR #1289, opened 2022-04-20
Wrap with a simple HTTP server

gkalele:
Add simple JSON REST API handlers to serve the structs as JSON.


Lomanic:
Thank you for your PR but providing a webserver to expose gopsutil data has never been in the library scope.

PR #1295, opened 2022-05-05
Add "Pid" and "Type" fields

dean2021:


shirou:
Thank you for your contribution. However, "Type" field is a different among platforms.

For example, [BOOT_TIME is 1 in freebsd](https://github.com/freebsd/freebsd-src/blob/373ffc62c158e52cde86a5b934ab4a51307f9f2e/include/utmpx.h#L59), but [BOOT_TIME is 2 on Linux](https://man7.org/linux/man-pages/man5/utmp.5.html). So we should define a table of those types and convert from platform dependent value to the independent value.

PR #1296, opened 2022-05-05
[disk][windows] Set EnableCounterForIoctl reg key to 1 in init() to automatically enable diskperf on Server editions

### LINKED ISSUE #1094: disk.IOCounters() fails in virtual windows
**Describe the bug**
Executing of disk.IOCounters() in Virtual Windows (WEMU or VMWare) player results in error:
"The system cannot find the path specified" 

**To Reproduce**
Compile the following code with the CLI: **GOOS=windows go build -tags windows -o file.exe file.go**

```go
	_, err := disk.IOCounters()
	if err != nil {
		panic(err)
	}
```

**Expected behavior**
Returned disk io counters

**Additional context**

- The code is compiled on Ubuntu 20.04 and executed on virtual Windows Server 2019 running in VMWare Player. Host is Ubuntu 20.04 
- Debugging of the source revealed that the path "\\.\c:" cannot be found. The failure is observed in this piece of code https://github.com/shirou/gopsutil/blob/ec072d99db0fecd5663e7127a3bd39c958522a23/disk/disk_windows.go#L167

Lomanic:
Fixes #1094

Tested on Windows Server 2016

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/shirou/gopsutil/v3/internal/common"
 	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


shirou (approved):
Great! Thank you so much!

PR #1301, opened 2022-05-16
[process][windows] Use WaitForSingleObject with a 0 delay in PidExistsWithContext

### LINKED ISSUE #1298: I have a Windows process that definitely exists, but I'm getting "process does not exist" error
**Describe the bug**
On Windows, when calling `NewProcess` passing a pid that exists for some processes I get "process does not exist" error, whereas the process definitely exists.

**To Reproduce**
This happens to some processes only, it works as exepcted for most of the processes.

```go
package main

import (
	"fmt"
)

func findPid(pid int32) {
	if pids, err := process.Pids(); err != nil {
		fmt.Printf("error calling Pids: %v\n", pid, err)
	} else {
		pidFound := false
		for _, c := range pids {
			if c == pid {
				pidFound = true
				break
			}
		}
		fmt.Printf("Pid %d found: %v\n", pid, pidFound)
	}

}

func callNewProcess(pid int32) {
	if p, err := process.NewProcess(pid); err != nil {
		fmt.Printf("error calling NewProcess for pid %d: %v\n", pid, err)
	} else {
		fmt.Printf("ok calling NewProcess for pid %d. Result: %v\n", pid, p)
	}
}

func main() {
	pid := int32(6756)
	findPid(pid)
	callNewProcess(pid)
	findPid(pid)
	callNewProcess(pid)

}
```

Output:

```
Pid 6756 found: true
error calling NewProcess for pid 6756: process does not exist
Pid 6756 found: true
error calling NewProcess for pid 6756: process does not exist
```

**Expected behavior**
Since the pid and the process exist I expect the NewProcess call to succeed.

 - Windows: Microsoft Windows [Version 10.0.14393]

**Additional context**
This is happening with some but not all VirtualBox Headless Frontend processes. I can see the process in the Process Explorer, I can also kill it form there if I like, as far as I can see it is no different from other processes. Process Explorer marks this process as "packed" and the command line for the process as displayed by the process explorer does not start with the executable name, which is a difference from most processes. Some pictures are below.

ps01
ps02

Lomanic:
Reference https://stackoverflow.com/a/6493793

Fixes #1298

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -285,8 +285,7 @@ func PidExistsWithContext(ctx context.Context, pid int32) (bool, error) {
 		}
 		return false, err
 	}
-	const STILL_ACTIVE = 259 // https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess
-	h, err := windows.OpenProcess(processQueryInformation, false, uint32(pid))
+	h, err := windows.OpenProcess(windows.SYNCHRONIZE, false, uint32(pid))
 	if err == windows.ERROR_ACCESS_DENIED {
 		return true, nil
 	}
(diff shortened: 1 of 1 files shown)


shirou (approved):
Great. Thank you so much!

PR #1302, opened 2022-05-18
get IOCounters for aix with perfstat

chbuescher:
get net interface statistics for aix from perfstat without calling netstat and parsing the output

### DIFF
--- a/cpu/cpu_aix.go
+++ b/cpu/cpu_aix.go
@@ -5,70 +5,12 @@ package cpu
 
 import (
 	"context"
-
-	"github.com/power-devops/perfstat"
 )
 
 func Times(percpu bool) ([]TimesStat, error) {
 	return TimesWithContext(context.Background(), percpu)
 }
 
-func TimesWithContext(ctx context.Context, percpu bool) ([]TimesStat, error) {
-	var ret []TimesStat
-	if percpu {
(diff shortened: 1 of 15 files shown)


shirou:
Thank you for the PR. However, [perfstat](https://github.com/power-devops/perfstat) looks using CGO. We would like to implement as much as possible in Pure Go. Are there any obvious advantages over the current implementation?

Lomanic:
We could have two files `net/net_aix_cgo.go` `net/net_aix_nocgo.go` redefining IOCountersWithContext according to build flags like on some other OSes, that would be preferable.

chbuescher:
I see the point using pure go, especially for cross compilation.
But using the perfstat calls should improve memory & cpu usage and make it more robust.
(language settings, changes of cli tools options or output format ..)
Btw, the aix cpu, load, disk & net implementations already are using perfstat

shirou:
Ah, sorry I forget perfstat is already added. So could you split into `net/net_aix_cgo.go` and `net/net_aix_nocgo.go` as @Lomanic said? (and also rebase to fix lint error)

chbuescher:
uh, magic, i thought i had to create a new pull request but it seems github did the merge for me.

shirou (approved):
Thank you for updating and great contribution!

PR #1308, opened 2022-06-06
cpu usage percent calculation error

sworgod:
`idle  = idle + iowait`, so `busy = user + system + nice + lrq + softirq + steal`.
I refer to the algorithm of the psutil project, and also look for relevant information.
The psutil project only calculates the user and system. I understand that the calculation results may be inaccurate.
Then I looked up the information.(http://blog.foool.net/2020/09/%E5%88%A9%E7%94%A8-proc%E7%B2%BE%E7%A1%AE%E8%AE%A1%E7%AE%97linux%E7%B3%BB%E7%BB%9F%E7%9A%84cpu%E5%88%A9%E7%94%A8%E7%8E%87/)
After my test, it should be like this.
My English is not very good.

### DIFF
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -99,9 +99,9 @@ func (c InfoStat) String() string {
 }
 
 func getAllBusy(t TimesStat) (float64, float64) {
-	busy := t.User + t.System + t.Nice + t.Iowait + t.Irq +
+	busy := t.User + t.System + t.Nice + t.Irq +
 		t.Softirq + t.Steal
-	return busy + t.Idle, busy
+	return busy + t.Idle + t.Iowait, busy
 }
 
 func calculateBusy(t1, t2 TimesStat) float64 {


shirou:
Thank you for point them out! You are true. According to [this psutil code](https://github.com/giampaolo/psutil/blob/master/psutil/__init__.py#L1650), if `iowait` is defined, then substrcat iowait from busy.


To adapt the psutil implementation , it seems better to use `Total()`. Therefore, how about the following code? (import runtime is also required to build)

```go
// Deprecated: Total returns the total number of seconds in a CPUTimesStat
// Please do not use this internal function.
func (c TimesStat) Total() float64 {
	total := c.User + c.System + c.Idle + c.Nice + c.Iowait + c.Irq +
		c.Softirq + c.Steal + c.Guest + c.GuestNice

	return total
}


func getAllBusy(t TimesStat) (float64, float64) {
	tot := t.Total()
	if runtime.GOOS == "linux" {
		tot -= t.Guest     // Linux 2.6.24+
		tot -= t.GuestNice // Linux 3.2.0+
	}

	busy := tot - t.Idle - t.Iowait

	return tot, busy
}
```

I also noticed we expose `Total()` but it is not implemented in psutil and should be used internally. But we can not delete exposed function in order to keep compatibility. So I want to add `Deperecated` comments.

shirou:
replaced by #1325. Thank you for your contribution!

PR #1314, opened 2022-06-17
implemented more AIX no-cgo functions

chbuescher:
added load values, memory & swap usage and cpu total usage


shirou (approved):
Have not checked it is working, but I don't see a problem with this PR. Thank you so much!

PR #1322, opened 2022-07-02
fix(net,linux): fix decodeaddress if Big Endian

### LINKED ISSUE #990: s390x support
I was building/executing tests for [vault](https://github.com/hashicorp/vault). There are some tests makes use of routines from https://github.com/shirou/gopsutil/blob/master/cpu/cpu_linux.go#L116 to read /proc/cpuinfo

I could build vault from source however thetwo tests are failing for s390x architecture ( with an error: Expected non-nil CPU value) as this arch's /proc/cpuinfo has different contents than intel x86.

Is there any plan to support s390x arch for https://github.com/shirou/gopsutil ?
Please let me know.

shirou:
fixes #990

Check endian when decode address.

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"net"
+	"unsafe"
 
 	"github.com/shirou/gopsutil/v3/internal/common"
 )
@@ -270,4 +271,11 @@ func getIOCountersAll(n []IOCountersStat) ([]IOCountersStat, error) {
 	}
 
 	return []IOCountersStat{r}, nil
+}
+
+// IsLittleEndian checks if the current platform uses little-endian.
+// copied from https://github.com/ntrrg/ntgo/blob/v0.8.0/runtime/infrastructure.go#L16 (MIT License)
+func IsLittleEndian() bool {
+	var x int16 = 0x0011
+	return *(*byte)(unsafe.Pointer(&x)) == 0x11
 }
--- a/net/net.go
+++ b/net/net.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"net"
-	"unsafe"
 
 	"github.com/shirou/gopsutil/v3/internal/common"
 )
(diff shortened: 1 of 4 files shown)


zhsj:
With this patch, `Test_Connections` pass on s390x, but a new test fails:

```
--- FAIL: TestDecodeAddress (0.00s)
    net_linux_test.go:173:
                Error Trace:    net_linux_test.go:173
                Error:          Not equal:
                                expected: "10.0.0.5"
                                actual  : "5.0.0.10"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -10.0.0.5
                                +5.0.0.10
                Test:           TestDecodeAddress
                Messages:       0500000A:0016
    net_linux_test.go:173:
                Error Trace:    net_linux_test.go:173
                Error:          Not equal:
                                expected: "127.0.0.1"
                                actual  : "1.0.0.127"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -127.0.0.1
                                +1.0.0.127
                Test:           TestDecodeAddress
                Messages:       0100007F:D1C2
FAIL
```

shirou:
@zhsj fix tests. Could you try again?

zhsj:
Now the tests have passed, except `TestCpuInfo`, but it's unrelated to this PR.

net/net.go
```diff
@@ -271,3 +272,10 @@ func getIOCountersAll(n []IOCountersStat) ([]IOCountersStat, error) {
 
 	return []IOCountersStat{r}, nil
 }
+
+// IsLittleEndian checks if the current platform uses little-endian.
+// copied from https://github.com/ntrrg/ntgo/blob/v0.8.0/runtime/infrastructure.go#L16 (MIT License)
+func IsLittleEndian() bool {
```
Lomanic, line 278: do we want this function to be exported here?

shirou: Absolutely no! I have moved to `internal/common` because it may be used from other package. Thank you for pointing out!!

PR #1325, opened 2022-07-09
fix(cpu): fix cpu total and busy calc

shirou:
The current CPU Total/Busy calculation is different from [psutil code](https://github.com/giampaolo/psutil/blob/master/psutil/__init__.py#L1650). This PR changes to make it the same as psutil.

This PR is based on #1308. Thank your @sworgod, and sorry for overwriting the same code. This PR is a credit to you. I appreciate it very much.

Note: I found `cpu.Total()` is exposed unintentionally. It will be removed in future release.

### DIFF
--- a/cpu/cpu.go
+++ b/cpu/cpu.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


sworgod:
You're welcome,my English is not very good.
不用谢，我也是在使用过程中发现了这个问题。

edigaryev:
> Note: I found `cpu.Total()` is exposed unintentionally. It will be removed in future release.

We use this function [in the Cirrus CI Agent](https://github.com/cirruslabs/cirrus-ci-agent/blob/b155a919e0f24053807ea192fc2073e234dad51f/internal/executor/metrics/source/cgroup/cpu/cpu_linux.go#L46-L54) and it's pretty useful.

Is there a reason to remove it? Or maybe there's an alternative that allows to easily calculate the sum of all of the `TimesStat` fields?

shirou:
Because [psutil does not have](https://psutil.readthedocs.io/en/latest/#system-related-functions) a method called Total. Also, Total is used to calculate Busy, but this value [depends on the version of Linux](https://github.com/shirou/gopsutil/blob/master/cpu/cpu.go#L107-L108). In other words, the meaning of `Total` may differ. So it is deprecated and should not be used.

However, since we have heard that you are using it now, this method will not disappear (but not guaranteed).

sworgod:
> Because [psutil does not have](https://psutil.readthedocs.io/en/latest/#system-related-functions) a method called Total. Also, Total is used to calculate Busy, but this value [depends on the version of Linux](https://github.com/shirou/gopsutil/blob/master/cpu/cpu.go#L107-L108). In other words, the meaning of `Total` may differ. So it is deprecated and should not be used.
> 
> However, since we have heard that you are using it now, this method will not disappear (but not guaranteed).

The `total`  method should not be canceled, Because I am also using it.
No need to guarantee identical functionality with psutil.

PR #1334, opened 2022-07-25
mem.VirtualMemory(): wrong page size on darwin arm64

kyz:
`mem.VirtualMemory()` returns values that are 4x larger/smaller than they should be on darwin arm64.

Generally speaking, `unix.Getpagesize()` returns the correct page size on macOS; 4K on Intel Macs and 16K on M1 Macs.

However, that is not the unit that `host_statistics()` values are measured in; they are  _kernel_ memory pages which are `vm_kernel_page_size` bytes long.

You can see this in the source code of `vm_stat`, which is the executable that the non-CGO version of the darwin VM stats runs and parses.

https://github.com/unofficial-opensource-apple/system_cmds/blob/master/vm_stat.tproj/vm_stat.c#L126-L132
> ```
>printf("Mach Virtual Memory Statistics: (page size of %llu bytes)\n", (mach_vm_size_t)vm_kernel_page_size);
>```

This PR replaces `unix.Getpagesize()` with `C.vm_kernel_page_size`

### DIFF
--- a/mem/mem_darwin_cgo.go
+++ b/mem/mem_darwin_cgo.go
@@ -5,6 +5,7 @@ package mem
 
 /*
 #include <mach/mach_host.h>
+#include <mach/vm_page_size.h>
 */
 import "C"
 
@@ -12,8 +13,6 @@ import (
 	"context"
 	"fmt"
 	"unsafe"
-
-	"golang.org/x/sys/unix"
 )
 
 // VirtualMemory returns VirtualmemoryStat.
@@ -34,7 +33,7 @@ func VirtualMemoryWithContext(ctx context.Context) (*VirtualMemoryStat, error) {
 		return nil, fmt.Errorf("host_statistics error=%d", status)
 	}
 
-	pageSize := uint64(unix.Getpagesize())
+	pageSize := uint64(C.vm_kernel_page_size)
 	total, err := getHwMemsize()
 	if err != nil {
 		return nil, err


shirou:
Thank you for the PR! 

> Running an x86_64 executable on an M1 Mac (under emulation): unix.Getpagesize() returns 4096 (incorrect)

Is this pattern the only incorrect? and "Emulation" means Rosetta 2?

kyz:
That is the case. The built-in emulation via Rosetta 2

shirou (approved):
I have confirmed this short code.

```
package main

/*
#include <mach/mach_host.h>
#include <mach/vm_page_size.h>
*/
import "C"

import (
        "fmt"
        "golang.org/x/sys/unix"
)

func main() {
        cPageSize := uint64(C.vm_kernel_page_size)
        uPageSize := unix.Getpagesize()

        fmt.Printf("C=%d, unix=%d\n", cPageSize, uPageSize)
}
```

using
```
[golang.org/x/sys](http://golang.org/x/sys) v0.0.0-20220728004956-3c1f35247d10 
```


on Mac 11 i7:
```
C=4096, unix=4096
```

on M1 mac (I run on [scaleway.com](http://scaleway.com/). Thank you for the sponsors!)
```
GOOS=darwin GOARCH=amd64 /usr/local/go/bin/go run main.go
C=16384, unix=4096

GOOS=darwin GOARCH=arm64 /usr/local/go/bin/go run main.go
C=16384, unix=16384
```

According to current code, `unix.Getpagesize` uses [sysctl hw.pagesize](https://github.com/golang/go/blob/9839668b5619f45e293dd40339bf0ac614ea6bee/src/runtime/os_darwin.go#L172). And Rosetta 2 uses [4kb](https://github.com/dotnet/runtime/issues/44897#issuecomment-734032437). I think that means in Rosetta 2, it is rewritten including sysctl, but `C.vm_kernel_page_size` is still the same.

Now, as gopsutil, which is the correct value ?
As a gopsutil, I think we should use `C.vm_kernel_page_size` as your PR because we want to know memory on a host, even if it works in the Rosetta 2.

So let me merge this great PR. Thank you so much!

PR #1335, opened 2022-07-28
host, process: delete darwin/386 code

tklauser:
The darwin/386 port was dropped in Go 1.15 and the imported version of
golang.org/x/sys also no longer supports the port. This module requires
at least Go 1.15 per go.mod, so it no longer builds on darwin/386
anyway.


shirou (approved):
Agree. Thank you so much! (I have added my commit about deleting from Makefile and README)

PR #1336, opened 2022-08-02
Add hostinfo Virtualization hyperv

johanburati:


shirou (approved):

I confirmed that there is `hv_utils` entry in `/proc/modules` by using Azure VM (Thank you sponsors!)

PR #1341, opened 2022-08-12
implement aix nocgo disk usage

chbuescher:
This time i wanted to get disk usage data for aix nocgo.
To get your disk_unix implementation working for aix, i implemented getFSType() for aix.
Getting the partition list with PartitionsWithContext() was a bit more work.
Because Getfsstat() is not available for aix, i had to implement a parser for the output of the mount command.

### DIFF
--- a/disk/disk_aix_nocgo.go
+++ b/disk/disk_aix_nocgo.go
@@ -5,14 +5,78 @@ package disk
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


shirou:
Thank you and sorry to late reply. I can not check this PR is working on AIX, but it seems no problem.

PR #1346, opened 2022-08-25
fix(process,windows): compare len(cwd) to an incorrect value

tienthanh1993:
[process/process_windows.go#L411](https://github.com/shirou/gopsutil/blob/master/process/process_windows.go#L411)  len(cwd) need compare to userProcParams.CurrentDirectoryPathNameLength instead of userProcParams.CurrentDirectoryPathAddress
```go
	if procIs32Bits {
		userProcParams, err := getUserProcessParams32(h)
		if err != nil {
			return "", err
		}
		if userProcParams.CurrentDirectoryPathNameLength > 0 {
			cwd := readProcessMemory(syscall.Handle(h), procIs32Bits, uint64(userProcParams.CurrentDirectoryPathAddress), uint(userProcParams.CurrentDirectoryPathNameLength))
			if len(cwd) != int(userProcParams.CurrentDirectoryPathAddress) {
				return "", errors.New("cannot read current working directory")
			}

			return convertUTF16ToString(cwd), nil
		}
	} else {
		userProcParams, err := getUserProcessParams64(h)
		if err != nil {
			return "", err
		}
		if userProcParams.CurrentDirectoryPathNameLength > 0 {
			cwd := readProcessMemory(syscall.Handle(h), procIs32Bits, userProcParams.CurrentDirectoryPathAddress, uint(userProcParams.CurrentDirectoryPathNameLength))
			if len(cwd) != int(userProcParams.CurrentDirectoryPathNameLength) {
				return "", errors.New("cannot read current working directory")
			}

			return convertUTF16ToString(cwd), nil
		}
	}
```

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -408,7 +408,7 @@ func (p *Process) CwdWithContext(_ context.Context) (string, error) {
 		}
 		if userProcParams.CurrentDirectoryPathNameLength > 0 {
 			cwd := readProcessMemory(syscall.Handle(h), procIs32Bits, uint64(userProcParams.CurrentDirectoryPathAddress), uint(userProcParams.CurrentDirectoryPathNameLength))
-			if len(cwd) != int(userProcParams.CurrentDirectoryPathAddress) {
+			if len(cwd) != int(userProcParams.CurrentDirectoryPathNameLength) {
 				return "", errors.New("cannot read current working directory")
 			}
 


shirou (approved):
Great catch! This is different from [64bit](https://github.com/shirou/gopsutil/blob/master/process/process_windows.go#L424). Thank you!

PR #1347, opened 2022-09-07
Return all Windows partitions 

atoulme:
Return all partitions on Windows and all errors rather than returning early.

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -79,7 +79,21 @@ func UsageWithContext(ctx context.Context, path string) (*UsageStat, error) {
 	return ret, nil
 }
 
+type errorCollector []error
+
+func (c *errorCollector) collect(e error) { *c = append(*c, e) }
+
+func (c *errorCollector) Error() (err string) {
+	err = ""
+	for i, e := range *c {
+		err += fmt.Sprintf("\tError %d: %s\n", i, e.Error())
+	}
+
+	return err
+}
+
 func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, error) {
+	collector := new(errorCollector)
 	var ret []PartitionStat
 	lpBuffer := make([]byte, 254)
 	diskret, _, err := procGetLogicalDriveStringsW.Call(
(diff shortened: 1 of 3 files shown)


atoulme:
@shirou any chance you'd be available to review?

shirou:
Thank you for contribution(and sorry to late response)! Is the purpose of this PR to get all partition information in case of errors?
If so, I have the following comments.

1. How about moving `errorCollector` under internal/common, since this code seems generic?
2. Is the purpose achieved by using [go-multierror library]([https://github.com/hashicorp/go-multierror) ?

atoulme:
Thanks for your reply!
1. Sure thing
2. That works too. I didn't know if it was a good idea to pull in a new library and tried to go for the most minimal approach instead. Would you like to use https://github.com/uber-go/multierr or https://github.com/hashicorp/go-multierror?

shirou:
Thank you.

You are correct that gopsutil is minimalist oriented. Therefore, I am happy not to use these libraries. However, if you could align the method names, etc. with these libraries (ex. `Append`, not `collect`) , it would make it easier for other users who have used these libraries before.

Lomanic:
I remember there already is a multierr-like thing in gopsutil though, maybe reuse or adapt that? https://github.com/shirou/gopsutil/blob/16b3aac6adb5ef7c2c30e1b1cce90c35069b829a/host/types.go

shirou:
I found [errors.Join() is now added](https://github.com/golang/go/commit/4a0a2b33dfa3c99250efa222439f2c27d6780e4a) to `errors` package. I don't know when it will be released. But, I would change what I mentioned before, it would be great if you could implement this error handling with the intention of replacing it with this `errors.Join` instead of `multierr`.

atoulme:
Please take a look. I have refactored Warnings to be part of internal/common.

atoulme:
@shirou please let me know?

shirou (approved):
LGTM! Thank you so much!

davidnewhall:
Now that `host.Warnings` has been moved into an internal package, how are we supposed to unwrap it? 
I used to do this, but my code quit working when I updated this library.
```go
	var warns *host.Warnings
	if !errors.As(err, &warns) {
		return fmt.Errorf("unable to get sensor temperatures: %w", err)
	}

	for i, w := range warns.List {
		fmt.Sprint("warning %v: %v", i+1, w)
	}
```

klauspost:
@atoulme @shirou 

For changes [like these](https://github.com/shirou/gopsutil/pull/1347/files#diff-a9b9e6000e198acc88830e834a2ded604b329832b310e84f9c8969b58244da9dR1) that break compatibility may I recommend you add `type Warnings = common.Warnings` with a deprecation notice?

atoulme:
First PR here. I had the assumption that given that this was an internal package, it wasn’t imported directly downstream. Sorry for the breakage.

shirou:
I have opened #1379 to fix. (type alias is better, thank you for your suggestion.)

PR #1356, opened 2022-09-23
fix some kylin linux bug

N1neSun:
For Kylin Linux Desktop V10, It has `/etc/debian_version` which will cause the error to get host info.


shirou (approved):
sorry, I haven't noticed this PR. I don't have Kylin but it seems ok for me.

Issue #1357, opened 2022-09-27
[bug] windows: load.Avg() returns zeroes all time


elfrucool:
**Describe the bug**
When I use `load.Avg()` on windows, the resulting structure of type `load.AvgStat` has its fields filled with zero, it happens even after waiting several minutes.

**To Reproduce**
```go

// go.mod:
// require github.com/shirou/gopsutil/v3 v3.22.8

import (
	"fmt"
	"os"
	"time"

	"github.com/shirou/gopsutil/v3/load"
)

func main() {
	avg, err := load.Avg()
	if err != nil {
		fmt.Fprintf(os.Stderr, "Error calculating load average: %v\n", err)
		os.Exit(1)
	}

	var countdown int = 60 * 16 // this covers 1, 5 and 15 minutes
	fmt.Printf("[%10d] average: %#v\n", countdown, avg)

	for range time.Tick(10 * time.Second) {
		countdown = countdown - 10
		fmt.Printf("[%10d] average: %#v\n", countdown, avg)
		if countdown <= 0 {
			break
		}
	}
}
```

**Expected behavior**
Non-zero values in the `load.AvgStat` structure.

**Environment (please complete the following information):**
 - [x] Windows: `Microsoft Windows [Versión 10.0.19044.1889]`
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [ ] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**
[Cross-compiling? Paste the command you are using to cross-compile and the result of the corresponding `go env`]

PR #1358, opened 2022-09-29
[issue-1357] fix by calling NtQuerySystemInformation

elfrucool:
I use `NtQuerySystemInformation` instead of `ProcessorQueueLengthCounter` because the latter returns zero all time.

Tested with this program:

```go
package main

import (
	"fmt"
	"os"
	"time"

	"github.com/shirou/gopsutil/v3/load"
)

func main() {
	var countdown int = 60 * 1

	avg, err := load.Avg()
	if err != nil {
		fmt.Fprintf(os.Stderr, "Error calculating load average: %v\n", err)
		os.Exit(1)
	}
	fmt.Printf("[%2d] average: %.2f %.2f %.2f\n", countdown, avg.Load1, avg.Load5, avg.Load15)

	for range time.Tick(2 * time.Second) {
		countdown = countdown - 2
		avg, err := load.Avg()
		if err != nil {
			fmt.Fprintf(os.Stderr, "Error calculating load average: %v\n", err)
			os.Exit(1)
		}
		fmt.Printf("[%2d] average: %.2f %.2f %.2f\n", countdown, avg.Load1, avg.Load5, avg.Load15)
		if countdown <= 0 {
			break
		}
	}
}
```

### DIFF
--- a/internal/procstats/procstats_windows.go
+++ b/internal/procstats/procstats_windows.go
@@ -1,0 +1,219 @@
+//go:build windows
+// +build windows
+
+package procstats
+
+import (
+	"unsafe"
+
+	"github.com/shirou/gopsutil/v3/internal/common"
+	"golang.org/x/sys/windows"
+)
+
+// represents a win32 thread status
+// see <winternl.h> for possible values
+type THREAD_STATE uint32
+
+const (
+	StateInitialized THREAD_STATE = iota
+	StateReady
+	StateRunning
+	StateStandby
+	StateTerminated
+	StateWait
+	StateTransition
+	StateUnknown
+)
+
+func (s THREAD_STATE) String() string {
+	switch s {
+	case StateInitialized:
+		return "StateInitialized"
+	case StateReady:
+		return "StateReady"
+	case StateRunning:
+		return "StateRunning"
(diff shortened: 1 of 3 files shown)


shirou:
related to #1357

Thank you for your contribution. However, the [psutil uses "Processor Queue Length"](
https://github.com/giampaolo/psutil/blob/33981759a67601b91d925a14403077e8d218c015/psutil/arch/windows/wmi.c), and it works, gopsutil does has some issue. So I want to fix that instead of use other Win API.

And please do not add new package. We want to keep API and gopsutil aim to be multi-platform.

elfrucool:
> related to #1357
> 
> Thank you for your contribution. However, the [psutil uses "Processor Queue Length"](https://github.com/giampaolo/psutil/blob/33981759a67601b91d925a14403077e8d218c015/psutil/arch/windows/wmi.c), and it works, gopsutil does has some issue. So I want to fix that instead of use other Win API.
> 
> And please do not add new package. We want to keep API and gopsutil aim to be multi-platform.

Removed the new package, Processor Queue Length call returns 0 on my computer `Microsoft Windows [Versión 10.0.19044.1889]` and other folks' too.

PR #1360, opened 2022-10-04
Fix memory leaks in process_darwin_cgo

bduffany:
`NumThreadsWithContext`, `TimesWithContext`, and `MemoryInfoWithContext` were calling `C.malloc` but not calling `C.free` on the returned pointer, causing a leak. This change fixes that.


shirou (commented):
Great catch! I can not check on darwin, but this PR does fix the leaks. Thank you!

PR #1361, opened 2022-10-11
[android][host] fix Info() failure due to forbidden /proc/stat and /proc/uptime

tmm1:
these are not accessible on Android

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


tmm1:
> We could use https://pkg.go.dev/golang.org/x/sys/unix#Sysinfo, maybe as a fallback to a read error.

Good idea. I have implemented this approach on the PR, and tested working with Android 11

tmm1:
@Lomanic wdyt?

tmm1:
Friendly ping

internal/common/common_linux.go
```diff
@@ -68,7 +69,17 @@ func BootTimeWithContext(ctx context.Context) (uint64, error) {
...
+		var info syscall.Sysinfo_t
+		err := syscall.Sysinfo(&info)
+		if err != nil {
+			return 0, err
+		}
+
+		currentTime := time.Now().UnixNano() / int64(time.Second)
+		t := currentTime - int64(info.Uptime)
+		return uint64(t), nil
+	} else if err != nil {
```
Lomanic, line 82: useless `else` when you `return` unconditionally in the preceding `if`

tmm1:
```suggestion
	}
	if err != nil {
```

tmm1:
Thanks. How this now?

tmm1:
ping @shirou

PR #1362, opened 2022-10-11
[ios][disk] IOKit is not available on iOS

tmm1:
fix iOS builds with cgo trying to build with IOKit

### DIFF
--- a/disk/disk_darwin_cgo.go
+++ b/disk/disk_darwin_cgo.go
@@ -1,5 +1,5 @@
-//go:build darwin && cgo
-// +build darwin,cgo
+//go:build darwin && cgo && !ios
+// +build darwin,cgo,!ios
 
 package disk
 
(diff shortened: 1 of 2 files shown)


shirou:
Thank you! I don't know iOS, but [IOKit is available on iOS 16.0](https://developer.apple.com/documentation/iokit?language=objc)?

tmm1:
You're right, they include part of IOKit on iOS now.

But the actual drive stats are still macOS only:

https://developer.apple.com/documentation/iokit/kioblockstoragedriverstatisticslatentwritetimekey/

https://github.com/shirou/gopsutil/blob/360136f40562e030fb6c558d2c15d03afe4bf970/disk/iostat_darwin.c#L96

tmm1:
@shirou wdyt?

shirou (approved):
I don't have iOS devices. If it is not possible to build, I would consider this change does not break compatibility. If it breaks, someone will open an issue. so I merge this. Thank you so much!

PR #1364, opened 2022-10-14
Initial support for NetBSD

khorben:
These two commits will hopefully help bring support for NetBSD to the project. I have not been able to verify if these changes are correct yet ("go test" did not yield anything apparently useful). Please review before merging, and do so only if you feel like it is not expected to break anything new.

For these changes, I have simply copied the OpenBSD files as NetBSD, as the two systems are usually very close to one another. As such I think at least some things should work once the module is confirmed to build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ can be skipped.
 - Linux i386/amd64/arm(raspberry pi)
 - Windows i386/amd64/arm/arm64
 - Darwin amd64/arm64
+- NetBSD amd64 (not tested)
 - OpenBSD amd64 (Thank you @mpfz0r!)
 - Solaris amd64 (developed and tested on SmartOS/Illumos, Thank you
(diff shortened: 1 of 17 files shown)


shirou:
Thank you for your contribution. I have not checked all of the files, but if these netbsd files are just copied from openbsd, how about simply add `netbsd` to the build tag in openbsd files, like this?

```
//go:build openbsd netbsd
```

khorben:
It's a good idea, it might be a bit misleading though.

Let me try to actually test and verify if the changes work over the week-end.

khorben:
For the `cpu` part, some changes were required for the code to work on NetBSD.
I will continue with the other parts of the code ASAP.

Lomanic:
Closing, superseded by #1530 which is partially based off this PR, thanks @khorben.

PR #1365, opened 2022-10-19
Pass on context to ConnectionsPidWithContext

ib-steffen:
`ctx` should be passed on to ConnectionsPidWithContext.


shirou (approved):
Indeed. Great catch! Thank you!

PR #1373, opened 2022-11-04
feat: Setting the model name for arm based CPUs

yaozbek:
```
- Added arm model and model name as map.
- The modelName is set again according to the model value when the model name is empty.
- Based on lscpu source code.. https://github.com/util-linux/util-linux/blob/master/sys-utils/lscpu-arm.c

---

Signed-off-by: Yalcin Ozbek <yalcinozbekceng@gmail.com>
```

The contents of the /proc/cpuinfo file on some of the arm devices we work with are as follows. ModelName value is set to empty.. However, the model name can be obtained with commands such as **lscpu.** The source code of lscpu can be viewed [here](https://github.com/util-linux/util-linux/blob/master/sys-utils/lscpu-arm.c)
This PR has been created so that the ModelName value is not set empty.

Example 1:
```
processor	: 0
BogoMIPS	: 400.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd03
CPU revision	: 4

processor	: 1
BogoMIPS	: 400.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd03
CPU revision	: 4

processor	: 2
BogoMIPS	: 400.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd03
CPU revision	: 4

processor	: 3
BogoMIPS	: 400.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd03
CPU revision	: 4
```

Example 2:
```
processor	: 0
BogoMIPS	: 243.75
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x3
CPU part	: 0xd0c
CPU revision	: 1

processor	: 1
BogoMIPS	: 243.75
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x3
CPU part	: 0xd0c
CPU revision	: 1
```

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -244,16 +244,6 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 			c.Model = value
 		case "model name", "cpu":
 			c.ModelName = value
-			if c.VendorID == "ARM" && c.ModelName == "" {
-				if v, err := strconv.ParseUint(c.Model, 0, 16); err == nil {
-					modelName, exist := armModelToModelName[v]
-					if exist {
-						c.ModelName = modelName
-					} else {
-						c.ModelName = "Undefined"
-					}
-				}
-			}
 			if strings.Contains(value, "POWER8") ||
 				strings.Contains(value, "POWER7") {
 				c.Model = strings.Split(value, " ")[0]
@@ -298,6 +288,16 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 	if c.CPU >= 0 {
 		finishCPUInfo(&c)
 		ret = append(ret, c)
+	}
+	if c.VendorID == "ARM" && c.ModelName == "" {
+		if v, err := strconv.ParseUint(c.Model, 0, 16); err == nil {
+			modelName, exist := armModelToModelName[v]
+			if exist {
+				c.ModelName = modelName
+			} else {
+				c.ModelName = "Undefined"
+			}
+		}
 	}
 	return ret, nil
 }
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -244,13 +244,13 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 			c.Model = value
 		case "model name", "cpu":
 			c.ModelName = value
-			if c.ModelName == "" {
+			if c.VendorID == "ARM" && c.ModelName == "" {
 				if v, err := strconv.ParseUint(c.Model, 0, 16); err == nil {
-					modelName, isThereModel := armModelToModelName[v]
-					if !isThereModel {
-						c.ModelName = "Undefined"
-					} else {
+					modelName, exist := armModelToModelName[v]
+					if exist {
 						c.ModelName = modelName
+					} else {
+						c.ModelName = "Undefined"
 					}
 				}
 			}
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -285,10 +285,6 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 			c.Microcode = value
 		}
 	}
-	if c.CPU >= 0 {
-		finishCPUInfo(&c)
-		ret = append(ret, c)
-	}
 	if c.VendorID == "ARM" && c.ModelName == "" {
 		if v, err := strconv.ParseUint(c.Model, 0, 16); err == nil {
 			modelName, exist := armModelToModelName[v]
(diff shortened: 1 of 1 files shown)


shirou (changes requested):
Good catch, thank you!

I have tested on an GCP ARM instance (Thank you, sponsors!), I got nothing on ModelName.
`/proc/cpuinfo` is as follows and does not contain the `model name` nor `cpu` line. So your code is not executed.

```
% cat /proc/cpuinfo 
processor       : 0
BogoMIPS        : 50.00
Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
CPU implementer : 0x41
CPU architecture: 8
CPU variant     : 0x3
CPU part        : 0xd0c
CPU revision    : 1
```

I think it would be better to move the code into ["model"](https://github.com/shirou/gopsutil/pull/1373/files#diff-931769434306cfcfbbd0e67c1f45baa95c9e5450656cd4a233ba64344055a0e3R244). (Although I am concerned about the order in which "model" and "modelname" appears in the cpuinfo file.)

And also, I don't have a knowledge about model id uniqueness, but is it only for ARM? If so, it would be safer to add a check to see if it is ARM.

yaozbek:
> Good catch, thank you!
> 
> I have tested on an GCP ARM instance (Thank you, sponsors!), I got nothing on ModelName. `/proc/cpuinfo` is as follows and does not contain the `model name` nor `cpu` line. So your code is not executed.
> 
> ```
> % cat /proc/cpuinfo 
> processor       : 0
> BogoMIPS        : 50.00
> Features        : fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
> CPU implementer : 0x41
> CPU architecture: 8
> CPU variant     : 0x3
> CPU part        : 0xd0c
> CPU revision    : 1
> ```
> 
> I think it would be better to move the code into ["model"](https://github.com/shirou/gopsutil/pull/1373/files#diff-931769434306cfcfbbd0e67c1f45baa95c9e5450656cd4a233ba64344055a0e3R244). (Although I am concerned about the order in which "model" and "modelname" appears in the cpuinfo file.)
> 
> And also, I don't have a knowledge about model id uniqueness, but is it only for ARM? If so, it would be safer to add a check to see if it is ARM.

- In all ARM devices I work with, the model is added before the model name in /proc/cpuinfo. This is the case with most amd devices I've worked with. Because the model name is mapped to the model. You can review the source code of lscpu.  [lscpu source code](https://github.com/util-linux/util-linux/blob/master/sys-utils/lscpu-arm.c#L34)
For this reason, it will be beneficial for us to have the relevant code in modelName. Also, even if a modelName is defined, we do not overwrite this defined value with the if control.
 
- You're right about the other point, these unique hexadecimal values ​​are for ARM only. For this reason, I added the condition that the vendorID should be ARM in the if block.

shirou:
My point is, there is at least one machine which does not have a `model name` in cpuinfo. This prevents the code itself from executing. How about to move the code to outside of the for-loop, like [here](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_linux.go#L223)?

yaozbek:
> My point is, there is at least one machine which does not have a `model name` in cpuinfo. This prevents the code itself from executing. How about to move the code to outside of the for-loop, like [here](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_linux.go#L223)?

you are right. I moved the code.

shirou (approved):
I have confirmed it works on GCP ARM instance. Thank you so much!

```
=== RUN   TestCpuInfo
    cpu_test.go:123: {"cpu":0,"vendorId":"ARM","family":"","model":"0xd0c","stepping":1,"physicalId":"","coreId":"0","cores":1,"modelName":"Neoverse-N1","mhz":0,"cacheSize":0,"flags":["fp","asimd","evtstrm","aes","pmull","sha1","sha2","crc32","atomics","fphp","asimdhp","cpuid","asimdrdm","lrcpc","dcpop","asimddp","ssbs"],"microcode":""}
```

PR #1374, opened 2022-11-05
feat(host, windows): add UBR (Update Build Revision) to kernel version

### LINKED ISSUE #1372: Include UBR (Update Build Revision) into Windows version number
The currently returned version string on Windows is missing the UBR (Update Build Revision).

In PlatformInformationWithContext read the 'UBR' key from 'SOFTWARE\Microsoft\Windows NT\CurrentVersion' and append it to the version.
Instead of "10.0.19044 Build 19044" we would have "10.0.19044.1949 Build 19044".

None.

**Additional context**
The version withour UBR is not very useful, since it only changes in the half-yearly updates. The UBR allows to track if the Windows installation is up to date.
https://learn.microsoft.com/en-us/windows/release-health/windows11-release-information

shirou:
fixes #1372

This PR adds UBR (Update Build Revision) to current Build Version.

# Notice: this PR breaks compatibility

This PR changes the format of the string returned by `host.KernelVersion()` as follows. I do not believe there are any applications that parses this string and uses numbers, but if there are such applications and they implement a naive parser, this might be a problem.

From

```
    host_test.go:181: KernelVersion(): 10.0.22000 Build 22000
```
to

```
    host_test.go:181: KernelVersion(): 10.0.22000.1098 Build 22000.1098
```

### DIFF
--- a/host/host_windows.go
+++ b/host/host_windows.go
@@ -188,6 +188,14 @@ func PlatformInformationWithContext(ctx context.Context) (platform string, famil
 		}
 	}
 
+	var UBR uint32 // Update Build Revision
+	err = windows.RegQueryValueEx(h, windows.StringToUTF16Ptr(`UBR`), nil, &valType, nil, &bufLen)
+	if err == nil {
+		regBuf := make([]byte, 4)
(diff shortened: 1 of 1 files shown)


d1ss0nanz:
That was fast!

Thank you.

Issue #1378, opened 2022-11-18
Windows error: gopsutil: unexpected processor queue length counter error, please file an issue on github: err


axetroy:
**Describe the bug**

print the error information

```
gopsutil: unexpected processor queue length counter error, please file an issue on github: err
```

**To Reproduce**
```go
// paste example code reproducing the bug you are reporting
```

I am not sure what api cause this error

**Expected behavior**

It should works without any error information.

**Environment (please complete the following information):**
 - [x] Windows: [paste the result of `ver`]
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [ ] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**
[Cross-compiling? Paste the command you are using to cross-compile and the result of the corresponding `go env`]

```
set GO111MODULE=
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\Admin\AppData\Local\go-build
set GOENV=C:\Users\Admin\AppData\Roaming\go\env  
set GOEXE=.exe
set GOEXPERIMENT=
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\Users\Admin\go\pkg\mod
set GONOPROXY=
set GONOSUMDB=
set GOOS=windows
set GOPATH=C:\Users\Admin\go
set GOPRIVATE=
set GOPROXY=https://proxy.golang.com.cn,direct   
set GOROOT=C:\Program Files\Go
set GOSUMDB=sum.golang.org
set GOTMPDIR=
set GOTOOLDIR=C:\Program Files\Go\pkg\tool\windows_amd64
set GOVCS=
set GOVERSION=go1.19.3
set GCCGO=gccgo
set GOAMD64=v1
set AR=ar
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=C:\Users\Admin\go\src\cdcode.51vip.biz\digimaple\cloudoc_im_pc_builder\go.mod
set GOWORK=
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
[...]

PR #1380, opened 2022-11-19
[load][windows]: add error detail and context handling.

shirou:
refs: #1378

This PR change `load_windows`.
1. add `err` value to log message
2. add `context.Context` handling to stop goroutine

### DIFF
--- a/load/load_windows.go
+++ b/load/load_windows.go
@@ -26,7 +26,7 @@ var (
 // TODO instead of this goroutine, we can register a Win32 counter just as psutil does
 // see https://psutil.readthedocs.io/en/latest/#psutil.getloadavg
 // code https://github.com/giampaolo/psutil/blob/8415355c8badc9c94418b19bdf26e622f06f0cce/psutil/arch/windows/wmi.c
-func loadAvgGoroutine() {
+func loadAvgGoroutine(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


load/load_windows.go
```diff
@@ -37,20 +37,30 @@ func loadAvgGoroutine() {
...
 		loadErr = err
+		loadAvgMutex.Lock()
 		loadAvg1M = loadAvg1M*loadAvgFactor1M + currentLoad*(1-loadAvgFactor1M)
 		loadAvg5M = loadAvg5M*loadAvgFactor5M + currentLoad*(1-loadAvgFactor5M)
 		loadAvg15M = loadAvg15M*loadAvgFactor15M + currentLoad*(1-loadAvgFactor15M)
 		loadAvgMutex.Unlock()
-		<-tick
+	}
+
+	f() // run first time
```
axetroy, line 56: go f() maybe better?

shirou: Thank you for the comment. However, I don't think so. This `f()` is defined because I do not want to write the same function twice, the first time and called by tick (I want `do-while`). This function itself is in a goroutine, and the time to execute `counter.GetValue()` is very little, so I do not think it is necessary to increase a goroutine.

PR #1381, opened 2022-11-22
pull request on adding missing statistic under solaris/illumos

sfzfs:
The current codes miss below statistic data under solaris/illumos:
1. the disk io statistic data as: nread, nwritten, reads, writes, rtime, wtime;
2. the free memory under global zone;
3. the net io statistic data as: rbytes64, ipackets64, idrops64, ierrors, obytes64, opackets64, odrops64, oerrors.

The new feature branch adds the above missing statistic data based on the psutil project (https://psutil.readthedocs.io/), it has been tested under solaris ( Oracle Solaris 11.4 X86) and illumos (OmniOS v11 r151044).

### DIFF
--- a/mem/mem_solaris.go
+++ b/mem/mem_solaris.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/shirou/gopsutil/v3/internal/common"
+	"github.com/tklauser/go-sysconf"
 )
 
 // VirtualMemory for Solaris is a minimal implementation which only returns
@@ -34,6 +35,13 @@ func VirtualMemoryWithContext(ctx context.Context) (*VirtualMemoryStat, error) {
 			return nil, err
 		}
 		result.Total = cap
+		freemem, err := globalZoneFreeMemory()
+		if err != nil {
+			return nil, err
+		}
+		result.Available = freemem
+		result.Free = freemem
+		result.Used = result.Total - result.Free
 	} else {
 		cap, err := nonGlobalZoneMemoryCapacity()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


mem/mem_solaris.go
```diff
@@ -34,6 +35,13 @@ func VirtualMemoryWithContext(ctx context.Context) (*VirtualMemoryStat, error) {
 			return nil, err
 		}
 		result.Total = cap
+		freemem, err := globalZoneFreeMemory()
```
shirou, line 38: How about to pass `ctx` instead of creating a new context in this function?

```suggestion
		freemem, err := globalZoneFreeMemory(ctx)
```

shirou (commented):
Thank you for the contribution! Unfortunately, I can not confirm it works on solaris/illumos, but it looks good for me, except one comment.

mem/mem_solaris.go
sfzfs: Hi Shirou, 
Thank for your review and comment; I have made the suggested change and submitted the commit. 
For the tested OS environments, below is what I did for the Solaris and illumos environments:
1. For the Solaris OS test, I have created the Solaris instance from the Oracle Solaris 11.4 image provided by the Oracle Cloud marketplace at https://cloudmarketplace.oracle.com/marketplace/en_US/listing/61750333 (OCI allows to create free Solaris instances with the public image);
2. For the illumos OS test, I used the omniOS-r151044 (recommended by the [illumos.org](http://illumos.org/)) from https://omnios.org/download.html to create the virtual machine and do the test. 
For both environments, I tested with the go1.15  illumos/amd64 (and go1.13 illumos/amd64) and tests result look good.

shirou (approved):
Great work! Thanks a lot!!

sfzfs:
Thank, Shirou! Will contribute the other missing features under Solaris/Illumos if needed.

PR #1382, opened 2022-11-23
Packages

saksham-ghimire:
Added support to view installed, upgraded, purged packages on Linux and view installed software on windows.


Lomanic:
Thank you for your contribution, but gopsutil already has a broad API to support on many different platforms with few maintainers and limited time, we don't plan to add more.

PR #1384, opened 2022-11-30
Revert "[disk][windows] Support LabelWithContext & SerialNumberWithContext - rebased "

shirou:
Reverts shirou/gopsutil#1367

it seems test failed after merged. https://github.com/shirou/gopsutil/actions/runs/3587986129/jobs/6038901787. revert for temporary.

PR #1386, opened 2022-12-01
Retract deleted tag v3.22.11

bogdandrutu:
This is to disallow to re-tag v3.22.11 since you will generate security issues/warnings (different checksum)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,3 +11,5 @@ require (
 	github.com/yusufpapurcu/wmi v1.2.2
 	golang.org/x/sys v0.2.0
 )
+
+retract v3.22.11


hansgylling:
Is the subject of this PR supposed to say "Retract deleted tag v3.22.11"? It's v3.22.11 that's retracted in the PR, not v3.22.10. Also tag v3.22.10 is still there, it's not deleted.

shirou:
Ah, sorry, I was not aware of the impact of this deletion.

The background is, I have merged https://github.com/shirou/gopsutil/pull/1384 yesterday, but test failed (That test was success when before merged). And v3.22.11 was tagged automatically while merged. So I have removed the v3.22.11 tag because it might be affected by the failure of this test.

I agree with making it "retracted", v3.22.11, not v3.22.10.

bogdandrutu:
@shirou fixed the title, sorry.. the go.mod is fine

shirou (approved):
Great! Thank you so much!

PR #1389, opened 2022-12-06
fill modelName for all cores in arm64 devices

sgthammer:
If there are more than one core for arm64 based devices, model name of CPU was not filled for all the cores except the last core. For instance, if there are 4 cores, only the last core's model name info is filled. As there is no model name line in **/proc/cpuinfo** file for arm64 devices(refer to: **arch/arm64/kernel/cpuinfo.c** in linux kernel code), it is better to move the code under case _"model", "CPU part"_. Also, all the core's model names will filled as this case is inside of loop. 

The code is tested for both amd64 and arm64 devices:

- Arm Cortex-A53
- 11th Gen Intel(R) Core(TM) i7-11850H @ 2.50GHz

### DIFF
--- a/cpu/cpu_linux.go
+++ b/cpu/cpu_linux.go
@@ -242,6 +242,17 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 			c.Family = value
 		case "model", "CPU part":
 			c.Model = value
+			// if CPU is arm based, model name is found via model number. refer to: arch/arm64/kernel/cpuinfo.c
+			if c.VendorID == "ARM" {
+				if v, err := strconv.ParseUint(c.Model, 0, 16); err == nil {
+					modelName, exist := armModelToModelName[v]
+					if exist {
+						c.ModelName = modelName
+					} else {
+						c.ModelName = "Undefined"
+					}
+				}
+			}
 		case "model name", "cpu":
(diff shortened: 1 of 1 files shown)


sgthammer:
@shirou, could you please review this change or forward it to someone who can review?

shirou:
Hmm. The `model name` has been discussed once [here](https://github.com/shirou/gopsutil/pull/1373#issuecomment-1304544753). I found at least one device does not have a `model` line in `cpuinfo`. Could you share your cpuinfo by using [gist](gist.github.com/)?

sgthammer:
Cpuinfo link: https://gist.github.com/sgthammer/6b1b841f313897909bd33d5ce3bf1e49

For arm device we need to focus on "CPU Part" keyword. Kernel provides this line for all arm64 devices. Please refer to cpuinfo.c file as I provided in upper description.

shirou (approved):
I have confirmed `CPU part` is always provided by [cpuinfo.c](https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/cpuinfo.c#L221) as you said, I can reproduce with 4 CPU machine on GCP (Thank you sponsors!), and also confirmed all cpus have `ModelName` with your PR without errors. Thank you so much!

PR #1392, opened 2022-12-07
Truncate the Getfsstat result to the count of items that were returned

### LINKED ISSUE #1390: Partitions() on macOS may return uninitialized elements
**Describe the bug**
When calling disk.Partitions() on macOS (observed only on 10.14, so far), there may be uninitialized elements returned at the end of the PartitionStat array.

This happens because we don't truncate the fs array to the number of items returned by the second call to unix.Getfsstat(). This count may be less than the count returned by the first call to getfsstat.

**To Reproduce**
```go
partitions, err := disk.Partitions(false)
// iterate over partitions
// last partition in the list has Device, Mountpoint, and Fstype as all unprintable characters ���
```

**Expected behavior**
All elements in partitions contain valid data.

 - [X ] Mac OS: [paste the result of `sw_vers` and `uname -a`
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G103

Darwin (hostname redacted) 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64

**Additional context**

brianryner8:
This count may be less than what was returned by the first call to Getfsstat.

Fixes #1390

### DIFF
--- a/disk/disk_darwin.go
+++ b/disk/disk_darwin.go
@@ -20,9 +20,11 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 		return ret, err
 	}
 	fs := make([]unix.Statfs_t, count)
-	if _, err = unix.Getfsstat(fs, unix.MNT_WAIT); err != nil {
+	count, err := unix.Getfsstat(fs, unix.MNT_WAIT)
+	if err != nil {
 		return ret, err
 	}
+	fs = fs[:count] // Actual count may be less than from the first call.
 	for _, stat := range fs {
 		opts := []string{"rw"}
 		if stat.Flags&unix.MNT_RDONLY != 0 {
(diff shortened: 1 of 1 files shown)


brianryner8:
Note, I think the freebsd and openbsd implementations are similarly affected, but I don't have an environment to test changes there.

brianryner8:
@shirou would you be able to review this please?

I'm also wondering if you'd accept the fix on the v2 branch? Or are no more releases planned there?

Thanks,

shirou:
- I have quite limited time, and that only on weekends. (Last weekend, I ran out of budgets in the middle of review.)
- v2 branch is not maintained and no more release(except security issue).

brianryner8:
Updated the comment to mention 10.14 specifically, but also include the possibility that it affects other past or future releases.

shirou (approved):
Thank you for your updates and great contribution!

PR #1393, opened 2022-12-07
host: use unix.ByteSliceToString

tklauser:
Use ByteSliceToString provided in golang.org/x/sys/unix to convert \0-terminated byte slices to strings.


shirou (approved):
[unix.ByteSliceToString](https://pkg.go.dev/golang.org/x/sys/unix#ByteSliceToString). Indeed. Thank you so much!

PR #1394, opened 2022-12-07
cpu: use windows.GetActiveProcessorCount

tklauser:
Use GetActiveProcessorCount and the ALL_PROCESSOR_GROUPS const provided in golang.org/x/sys/windows. The function is available on Windows 7 and later. Go requires Windows 7, see https://go.dev/doc/go1.11#ports

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -14,8 +14,7 @@ import (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


shirou (approved):
This PR means Go already stop under Win7 support. Great catch. Thank you!

PR #1400, opened 2022-12-28
Detect Docker also using /.dockerenv

scop:
Docker itself does that too, for example in https://github.com/moby/libnetwork/blob/1f3b98be6833a93f254aa0f765ff55d407dfdd69/drivers/bridge/setup_bridgenetfiltering.go#L161

/proc/self/cgroup doesn't necessarily contain "docker".


shirou (approved):
Good catch! Thank you!

PR #1403, opened 2023-01-04
[host] linux/ppc: ppc64 definition from ppc64le

donbowman:
Same as commit fc04d2d, but for ppc64 from ppc64le definition.


shirou (approved):
Sorry for late response. Thank you for great contribution!

Issue #1411, opened 2023-02-01
lvm group named dev fails to resolve 


powersj:
**Describe the bug**

LVM devices that start with a group named "dev" do not correctly resolve to the `/dev/dm` name when calling `disk.Partitions(true)`. The root cause is the [string replace](https://github.com/shirou/gopsutil/blob/0494d7f0b9a36ccda018c7d60da7b42522832cc1/disk/disk_linux.go#L344) call on `/dev/mapper/` devices. It replaces all occurrences of `/dev`, when I believe it should only replace the first instance of `/dev/`.

**To Reproduce**

Assume I have two LVM devices:

```
❯ lsblk
NAME                      MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
sdb                         8:16   1 119.5G  0 disk 
└─dev_dent_d-dev_dent_d   254:1    0    16G  0 lvm  /storage/scsi1
sdc                         8:32   1 119.5G  0 disk 
└─prod_dent_d-prod_dent_d 254:2    0    16G  0 lvm  /storage/scsi2
❯ ls -l /dev/mapper/
total 0
crw------- 1 root root 10, 236 Feb  1 06:09 control
lrwxrwxrwx 1 root root       7 Feb  1 07:25 dev_dent_d-dev_dent_d -> ../dm-1
lrwxrwxrwx 1 root root       7 Feb  1 07:39 prod_dent_d-prod_dent_d -> ../dm-2
```

```go
package main

import (
	"fmt"

	"github.com/shirou/gopsutil/v3/disk"
)

func main() {
	parts, err := disk.Partitions(true)
	if err != nil {
		panic(err)
	}
	for _, part := range parts {
		fmt.Println(part.String())
	}
}
```

Results in:

```
{"device":"/dev/mapper/dev_dent_d-dev_dent_d","mountpoint":"/storage/scsi1","fstype":"xfs","opts":["rw","relatime"]}
{"device":"/dev/dm-2","mountpoint":"/storage/scsi2","fstype":"xfs","opts":["rw","relatime"]}
```

Notice how the first device does not become `/dev/dm-1`.

**Expected behavior**

I expect the `/dev/mapper/dev_dent_d-dev_dent_d` to resolve to `/dev/dm-1`:
```
{"device":"/dev/dm-1","mountpoint":"/storage/scsi1","fstype":"xfs","opts":["rw","relatime"]}
[...]

PR #1412, opened 2023-02-01
fix(disk): correctly replace /dev in /dev/mapper

powersj:
Only replace the first instance of /dev in a /dev/mapper string. Otherwise, if an LVM group is named dev it will replace that as well.

fixes: #1411


powersj:
@shirou anything I can do to help this along?

shirou (approved):
Indeed. Thank you for pointing out, good catch!

(sorry I am late because I can only do gopsutil work on weekends.)

PR #1421, opened 2023-02-24
Recognize Cumulus Linux distro

jak3kaj:
Recognize the Cumulus Linux distro and add Cumulus Linux and Kylin to debian family.

### DIFF
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -210,6 +210,9 @@ func PlatformInformationWithContext(ctx context.Context) (platform string, famil
 		} else if lsb.ID == "Kylin" {
 			platform = "Kylin"
 			version = lsb.Release
+		} else if lsb.ID == "\"Cumulus Linux\"" {
+			platform = "cumuluslinux"
+			version = lsb.Release
 		} else {
 			if common.PathExistsWithContents("/usr/bin/raspi-config") {
 				platform = "raspbian"
@@ -287,7 +290,7 @@ func PlatformInformationWithContext(ctx context.Context) (platform string, famil
 	platform = strings.Trim(platform, `"`)
 
 	switch platform {
-	case "debian", "ubuntu", "linuxmint", "raspbian":
+	case "debian", "ubuntu", "linuxmint", "raspbian", "Kylin", "cumuluslinux":
 		family = "debian"
 	case "fedora":
 		family = "fedora"
--- a/host/host_linux.go
+++ b/host/host_linux.go
@@ -210,7 +210,7 @@ func PlatformInformationWithContext(ctx context.Context) (platform string, famil
 		} else if lsb.ID == "Kylin" {
 			platform = "Kylin"
 			version = lsb.Release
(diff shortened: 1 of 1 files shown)


host/host_linux.go
```diff
@@ -210,6 +210,9 @@ func PlatformInformationWithContext(ctx context.Context) (platform string, famil
 		} else if lsb.ID == "Kylin" {
 			platform = "Kylin"
 			version = lsb.Release
+		} else if lsb.ID == "\"Cumulus Linux\"" {
```
shirou, line 213: Is this `"` required? 
```suggestion
		} else if lsb.ID == "Cumulus Linux" {
```

jak3kaj: Yes, unfortunately it is required.  In my testing, it discovers `Debian` when I used the suggestion

``` go
} else if lsb.ID == "Cumulus Linux" {
``` 

Not many Linux Distributions have a space between their names in LSB release file, most just push together the words like `LinuxMint` and don't use double quotes.

Here are two samples of the /etc/lsb.release from Cumulus Linux switches running two different releases:

``` bash
cumulus@switch1:mgmt:~$ cat /etc/lsb-release
DISTRIB_ID="Cumulus Linux"
DISTRIB_RELEASE=4.4.0
DISTRIB_DESCRIPTION="Cumulus Linux 4.4.0"

cumulus@switch2:mgmt:~$ cat /etc/lsb-release
DISTRIB_ID="Cumulus Linux"
DISTRIB_RELEASE=5.2.1
DISTRIB_DESCRIPTION="Cumulus Linux 5.2.1"
```

jak3kaj: A better approach would be to accept your suggestion as well as updating the `getLsbStruct()` function, so it trims doublequotes from the strings:

``` go
            switch field[0] {
            case "DISTRIB_ID":
                ret.ID = field[1]
            case "DISTRIB_RELEASE":
                ret.Release = field[1]
            case "DISTRIB_CODENAME":
                ret.Codename = field[1]
            case "DISTRIB_DESCRIPTION":
                ret.Description = field[1]
```

to

``` go
            switch field[0] {
            case "DISTRIB_ID":
                ret.ID = strings.Trim(field[1], "\"")
            case "DISTRIB_RELEASE":
                ret.Release = field[1]
            case "DISTRIB_CODENAME":
                ret.Codename = strings.Trim(field[1], "\"")
            case "DISTRIB_DESCRIPTION":
                ret.Description = strings.Trim(field[1], "\"")
```

I'll update my fork if you'd prefer this method.

jak3kaj: Good suggestion.  I tested the raw string and that works as well.

Did you see my message about updating `getLsbStruct`?  That will remove the quotes from the `lsbStruct`, then I can just use a double quoted string and the Distro name strings will be more uniform.

``` go
} else if lsb.ID == "Cumulus Linux" {
```

jak3kaj: Yes, I understand.  I'll update with your raw string suggestion.

PR #1424, opened 2023-02-28
Groups in /proc/PID/status has type uint32.

chernomor:
Fix error of parser:
> error get info about worker process status 3150292: strconv.ParseInt:
> parsing "4294967293": value out of range

$ grep Groups /proc/self/status
Groups: 20001 [...] 4294967293


shirou (approved):
Sorry for the wait. This PR has been merged into the branch for v4. Thank you for your contribution!

Issue #1429, opened 2023-03-13
plugins\inputs\system\ps.go:220:26: undefined: host.Warnings


powersj:
**Describe the bug**
On FreeBSD, Darwin, and Windows, builds using `host.Warnings` now fail. It appears that this was removed in #1347, but then re-added only for Linux in #1379. Telegraf uses these warnings with the call to `host.SensorsTemperatures()` as it seems to almost always return an error. We need to unwrap the error to see if it contains only a warning:

```s
plugins/inputs/system/ps.go:220:26: undefined: host.Warnings
```

**To Reproduce**
```go
package main

import (
    "errors"
    "fmt"
    "os"

    "github.com/shirou/gopsutil/host"
)

func main() {
    temp, err := host.SensorsTemperatures()
    if err != nil {
 	var hostWarnings *host.Warnings
	if !errors.As(err, &hostWarnings) {
	    fmt.Println(err)
	    os.Exit(1)
	}
    }
    fmt.Println(temp)
}
```

**Expected behavior**

Sensors print out results, ignoring warnings:

```
[{"sensorKey":"nvme_composite","temperature":33.85,"sensorHigh":74.85,"sensorCritical":79.85} {"sensorKey":"nvme_composite","temperature":47.85,"sensorHigh":89.85,"sensorCritical":94.85} {"sensorKey":"k10temp_tctl","temperature":46.5,"sensorHigh":0,"sensorCritical":0} {"sensorKey":"k10temp_tccd1","temperature":36.25,"sensorHigh":0,"sensorCritical":0} {"sensorKey":"k10temp_tccd2","temperature":40.5,"sensorHigh":0,"sensorCritical":0}]
```

**Environment (please complete the following information):**
 - [x] Windows: [paste the result of `ver`]
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [x] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [x] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**

Yes we are cross-compiling, this was reported to us in Telegraf in https://github.com/influxdata/telegraf/issues/12841

PR #1431, opened 2023-03-15
fix: moves common.Warnings to common OS files

powersj:
This allows all OS types to unwrap the Warnings from errors, and not just linux in host or windows in disk.

fixes: #1429


shirou (approved):
Definitely. Thank you!

PR #1436, opened 2023-03-23
cpu: add frequency support for apple silicon m1/m2 cpus

### LINKED ISSUE #1000: cpu.Info() errors on Apple Silicon M1 (darwin/arm64)
Getting "no such file or directory", from here:

https://github.com/shirou/gopsutil/blob/340db113de4d1b4f14905e76ca5012d4d45a4fc9/cpu/cpu_darwin.go#L94-L98

I confirmed this sysctl is not present on M1:

```
tmm1@m1 ~ % sysctl hw.cpufrequency
tmm1@m1 ~ % echo $?
0
```

It only appears under Rosetta 2:

```
tmm1@m1 ~ % arch -x86_64 /bin/bash -c 'sysctl hw.cpufrequency'
hw.cpufrequency: 2400000000
```

shoenig:
This PR adds support for reading the frequency of Apple Silicon M1/M2 CPUs. We do so by reading the values out of the IOKit framework, as a few other projects have now demonstrated to be possible. This requires the use of CGO. The library provides a convenience IsAppleSilicon() guard to detect whether the values can be read.

Currently gopsutil does not support the big.LITTLE CPU architectures (i think?) - in fact the P and E cores have different max frequencies. For now, just read the P core frequency. The E core data is readily available if we want to read it in the future.

Closes #1000

Small example program

```go
package main

import (
        "fmt"

        "github.com/shoenig/go-m1cpu"

        "github.com/shirou/gopsutil/v3/cpu"
)

func main() {
        fmt.Println("is Apple Silicon:", m1cpu.IsAppleSilicon())
        fmt.Println("model name", m1cpu.ModelName())
        fmt.Println("pCore GHz", m1cpu.PCoreGHz())
        fmt.Println("eCore GHz", m1cpu.ECoreGHz())
        fmt.Println("pCore Hz", m1cpu.PCoreHz())
        fmt.Println("eCore Hz", m1cpu.ECoreHz())

        fmt.Println("----- gopsutil ----")

        infos, err := cpu.Info()
        if err != nil {
                panic(err)
        }

        for _, info := range infos {
                fmt.Println("info.Mhz", info.Mhz)
        }
}
```

```shell
go run main.go
is Apple Silicon: true
model name Apple M2 Pro
pCore GHz 3.504
eCore GHz 2.424
pCore Hz 3504000000
eCore Hz 2424000000
----- gopsutil ----
info.Mhz 3.504e+09
```

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/shoenig/go-m1cpu"
 	"github.com/tklauser/go-sysconf"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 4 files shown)


shirou (changes requested):
Excellent! This PR is very promising.

However, could you please fix the following?

- Create a new cpu_darwin_test.go and move all darwin tests there to fix the lint issue.
- [use strncmp instead of strcmp.](https://github.com/shoenig/go-m1cpu/blob/main/cpu.go#L74)

shoenig:
For `strncmp` I merged https://github.com/shoenig/go-m1cpu/pull/3/files and bumped the version here.

And I created `cpu_darwin_test.go` and added a specific `Test_CpuInfo_AppleSilicon` test case.

Let me know if there's anything else to do ~

shirou (approved):
Awesome! This PR is probably the PR that many people have been waiting for! Great thanks!

PR #1439, opened 2023-03-26
allow to pass context values to override environment variables

### LINKED ISSUE #1438: Offer a way to avoid using os.GetEnv
When using gopsutil programmatically, we want to avoid relying on environment variables. We have use cases where we want to find information about processes in the container and outside the container, for example.
We want to keep the current default behavior of reading from os.GetEnv. However, we also want a way to override optionally this approach by passing a map as part of the invocation.

3 approaches to this:
* set os.GetEnv as a function defined as a public var in the gopsutil common package, and allow users to access and override it. This is not thread-safe and probably just a quick hack to get by.
* Add a map of values to all calls to gopsutil as an additional argument. Keep the current API, making sure it passes in os.GetEnv as its default parameter value.
* Use a context parameter. The context parameter may optionally have a key that points to a custom function, that can be used instead of the default os.GetEnv.

Still looking.

**Additional context**
N/A

atoulme:
Fixes #1438

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -324,11 +324,11 @@ func PathExistsWithContents(filename string) bool {
 }
 
 // GetEnvWithContext retrieves the environment variable key. If it does not exist it returns the default.
-// The context may optionally contain a map superseding os.Env.
+// The context may optionally contain a map superseding os.EnvKey.
 func GetEnvWithContext(ctx context.Context, key string, dfault string, combineWith ...string) string {
 	var value string
-	if env, ok := ctx.Value(common.Env).(map[string]string); ok {
-		value = env[key]
+	if env, ok := ctx.Value(common.EnvKey).(common.EnvMap); ok {
+		value = env[common.EnvKeyType(key)]
 	}
 	if value == "" {
 		value = os.Getenv(key)
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -25,16 +25,14 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/shirou/gopsutil/v3/common"
 )
 
 var (
 	Timeout    = 3 * time.Second
 	ErrTimeout = errors.New("command timed out")
 )
-
-type envKey string
-
-var Env = envKey("string")
 
 type Invoker interface {
 	Command(string, ...string) ([]byte, error)
@@ -329,7 +327,7 @@ func PathExistsWithContents(filename string) bool {
 // The context may optionally contain a map superseding os.Env.
 func GetEnvWithContext(ctx context.Context, key string, dfault string, combineWith ...string) string {
 	var value string
-	if env, ok := ctx.Value(Env).(map[string]string); ok {
+	if env, ok := ctx.Value(common.Env).(map[string]string); ok {
 		value = env[key]
 	}
 	if value == "" {
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -32,6 +32,10 @@ var (
 	ErrTimeout = errors.New("command timed out")
 )
 
+type envKey string
+
+var Env = envKey("string")
+
 type Invoker interface {
 	Command(string, ...string) ([]byte, error)
 	CommandWithContext(context.Context, string, ...string) ([]byte, error)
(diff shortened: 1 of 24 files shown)


internal/common/common.go
```diff
@@ -341,34 +366,73 @@ func GetEnv(key string, dfault string, combineWith ...string) string {
 	}
 }
 
+// Deprecated: use HostProcWithContext instead
```
rmfitzpatrick, line 369: I'm not sure these changes warrant deprecating the existing methods since that could introduce lint issues for users unexpectedly and then require passing noop/background contexts for all usages.

dloucasfx (approved):

dloucasfx:
@shirou when you get a chance, can you please review this PR ?
Thanks

internal/common/common.go
```diff
@@ -32,6 +32,10 @@ var (
 	ErrTimeout = errors.New("command timed out")
 )
 
+type envKey string
+
+var Env = envKey("string")
```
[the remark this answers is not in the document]
atoulme: the internal folder is actually used by other users already. I tried to limit as much as possible the exposure of the internal folder to outside.
See https://github.com/shirou/gopsutil/pull/1347#issuecomment-1332145721 for a reference showing internal usage import.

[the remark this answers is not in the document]
rmfitzpatrick: I'm referring to external users who can't import internal packages

internal/common/common.go
atoulme: @shirou wdyt?

internal/common/common.go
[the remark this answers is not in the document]
atoulme: Oh I see the problem, this won't be available outside gopsutil otherwise so it needs to be exported elsewhere.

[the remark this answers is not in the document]
atoulme: OK this is addressed in f71f245

shirou:
Sorry to late response. I finally had time to review it. As a premise, I agree with the policy about the method of putting values into the `context.Context`.

So, I have a more implementation-oriented comment. [common.Env](https://github.com/shirou/gopsutil/pull/1439/files#diff-189fed0d432e11286e00bdea5788347c8a77c35559a3d4d2d698b1fc22c45) will become a  `map[string]string`, which means that any value can be put in it, and the user does not know what value can be put in it. This amplifies the bad points of `context`.

Therefore, how about to restrict the types in the `common` package as follows. (If you have a better name idea, please change as you like)

```
type EnvKeyType string
const (
    HostProcEnvKey EnvKeyType = "HOST_PROC"
    HostSysEnvKey EnvKeyType = "HOST_SYS"
    HostEtcEnvKey EnvKeyType = "HOST_ETC"
      (snip)
)
type EnvMap map[EnvKeyType]string
```

[GetEnvWithContext](https://github.com/shirou/gopsutil/pull/1439/files#diff-e0a57b903e60761f673a4485b4a747f500f4155828e182cc) retrieve side may became a bit more complicated, but I think this method is clearer and more extensible. More to the point, in the future, I would also like to add options other than `HOST_*` here.

How do you think?

internal/common/common.go
shirou: Thank you for the suggestion. Yes, I agree about we should not add `Deprecated` at this time.

atoulme:
I'm on it.

internal/common/common.go
atoulme: ok removed.

atoulme:
Ready for another review @shirou , thanks!

internal/common/common_linux.go
```diff
@@ -30,8 +30,13 @@ func DoSysctrl(mib string) ([]string, error) {
 	return values, nil
 }
 
+// Deprecated: use NumProcsWithContext instead
```
shirou, line 33: Could you remove this `Deprecated` also?

internal/common/common_test.go
```diff
@@ -160,3 +163,44 @@ func TestGetSysctrlEnv(t *testing.T) {
...
+
+func TestGetEnvWithNoContext(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("windows doesn't have etc")
+	}
+	old := os.Getenv("HOST_ETC")
+	os.Setenv("HOST_ETC", "/bar")
+	defer func() {
+		os.Setenv("HOST_ETC", old)
+	}()
```
shirou, line 185: How about to use [testing.Setenv](https://pkg.go.dev/testing#T.Setenv) instead?

internal/common/common_test.go
```diff
@@ -160,3 +163,44 @@ func TestGetSysctrlEnv(t *testing.T) {
...
+
+func TestGetEnvWithContextOverride(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("windows doesn't have etc")
+	}
+	old := os.Getenv("HOST_ETC")
+	os.Setenv("HOST_ETC", "/bar")
+	defer func() {
+		os.Setenv("HOST_ETC", old)
+	}()
```
shirou, line 200: ditto

internal/common/common_test.go
atoulme: Done.

internal/common/common_linux.go
atoulme: Done.

atoulme:
@shirou code review feedback addressed, thanks!

shirou (approved):
I would like to merge this as I think this is probably the best way to go.
Thank you very much for your patience with my review.

I believe this change will give gopsutil great potential for extensibility in the future. I would like to thank you again!

Issue #1444, opened 2023-04-04
Add Security Policy


gabibguti:
**Is your feature request related to a problem? Please describe.**
A Security Policy is important as it provides clear guidance on how to report potential vulnerabilities and inform the vulnerabilities disclosure window for this repo.

I see you already have Security Advisory enabled, and the potential vulnerabilities could be reported through Issues tab, but a Security Policy file would be relevant to communicate how long would it take for a vulnerability to disclosed. In other words, to be confirmed or fixed.

**Describe the solution you'd like**
Add a Security Policy file that points to Security Advisory to report potential vulnerabilities and informs the window for disclosing a vulnerability.

**Describe alternatives you've considered**
None.

**Additional context**
I'm Gabriela and I work on behalf of Google and the OpenSSF suggesting supply-chain security changes :)

If you agree, I can open a PR!

shirou:
I see. What kind of security policy do you think? I would be very grateful if you could give us an example. However, as you know, we are short of manpower for this project.

gabibguti:
@shirou I would suggest a posture similar to this example: https://github.com/Cyan4973/xxHash/security/policy

That would be:
- Doing security patches for vulnerabilities found only in the latest release
- Pointing users to report the vulnerabilities through GitHub's Security Advisories
- Inform users of when reported vulnerabilities will be confirmed, fixed and disclosed to the public (this can be on a best-effort basis time, not necessarily 90 days as in the example)

PR #1450, opened 2023-04-13
Add Security Policy

gabibguti:
Resolves https://github.com/shirou/gopsutil/issues/1444

Here's a Security Policy suggestion! Let me know what do you think and if this seems reasonable for maintenance :)


shirou (approved):
Thank you so much!

PR #1453, opened 2023-04-24
refactor(process): compare error with `errors.Is`

Juneezee:
Starting from Go 1.13, `errors.Is` is the preferable way to compare error equality [^1].

[^1]: https://go.dev/blog/go1.13-errors


shirou (approved):
Exactly! Thank you so much!

PR #1454, opened 2023-04-27
process: implement NumFDs for Windows

braydonk:
This PR implements the `NumFDs` method for Windows by borrowing the system info query from `OpenFiles` and using it to count any Handles belonging to the process.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -549,7 +549,37 @@ func (p *Process) NumCtxSwitchesWithContext(ctx context.Context) (*NumCtxSwitche
 }
 
 func (p *Process) NumFDsWithContext(ctx context.Context) (int32, error) {
-	return 0, common.ErrNotImplementedError
+	buffer := make([]byte, 1024)
+	var size uint32
+
+	st := common.CallWithExpandingBuffer(
+		func() common.NtStatus {
+			return common.NtQuerySystemInformation(
+				common.SystemExtendedHandleInformationClass,
+				&buffer[0],
+				uint32(len(buffer)),
+				&size,
+			)
+		},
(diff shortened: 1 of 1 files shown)


shirou:
Thank you for your PR. It works on my Windows 10, at least return some values. However, after some searching, I found [GetProcessHandleCount](https://stackoverflow.com/questions/15357961/retrieve-the-number-of-opened-file-descriptors-using-the-windows-api). With my knowledge I do not know which is better than your PR. Could you please provide us with some URLs or other references about this PR? Thank you so much!

braydonk:
Hi @shirou thanks for taking a look!

I think you're right, since this PR I've figured out a lot better ways to get this information. In this library in particular, I believe `GetProcessHandleCount` would indeed be the better choice given that it is designed to operate one process at a time.

I can make that change. In the accompanying issue #1455 I asked a question about the naming for this API; the name `NumFDs` actually wouldn't be very descriptive for this particular use case, because on Windows, handles held by a process are [a lot more than just file descriptors](https://learn.microsoft.com/en-us/windows/win32/sysinfo/object-categories). Is there a way that we could name this API more clearly? It looks like the current design is to have the same interface on all platforms, or is there precedent for breaking that somewhere I'm not seeing?

shirou:
`NumFDs` is comes from [psutil](https://psutil.readthedocs.io/en/latest/#psutil.Process.num_fds). psutil has a `num_handles` [for Windows](https://psutil.readthedocs.io/en/latest/#psutil.Process.num_handles). However, in go, we should use same name for all os, so I choose `NumFD`. 

And we can not change the name because of compatibility.

PR #1457, opened 2023-04-30
Add support for Pressure Stall Information on Linux

rare-magma:
This PR adds support for Pressure Stall Information (PSI) on Linux:
https://facebookmicrosites.github.io/psi/docs/overview

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ will provide useful information.
 - iptables nf_conntrack (linux only)
   - system wide stats on netfilter conntrack module
   - sourced from /proc/sys/net/netfilter/nf_conntrack_count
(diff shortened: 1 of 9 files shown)


shirou:
Thank you for the PR. However, gopsutil aims to be multi-platform. Therefore, we do not want to add any more Linux-specific information. If you want to add it, we need a working implementation on at least two of the four platforms Linux/Windows/Darwin/FreeBSD (either an actual implementation or some URL about how to get the information is fine)

rare-magma:
@shirou thanks for reviewing the PR. 
I understand now it is out of scope for the project since PSI is a set of metrics available on Linux only as far as I know. 
I'll proceed with closing the PR.

PR #1462, opened 2023-05-09
fix some comments

cuishuang:


shirou (approved):
Good point. Thank you so much!

Issue #1471, opened 2023-05-24
mem: linux WriteBack is incorrectly capitalized


powersj:
**Describe the bug**

As reported in https://github.com/influxdata/telegraf/issues/13299, a user noticed that the memory writeback values were not correctly reported since upgrading to v3. It looks like [gopsutil v3](https://github.com/shirou/gopsutil/blame/7c610c50377c48e230fd092e65e656b03b97b069/mem/mem_linux.go#L157) looks for `WriteBack` (note the capital `B`), while [v2 looked for `Writeback`](https://github.com/shirou/gopsutil/blob/36cc0d3c0b09a295f954f0b8f9cb6e4b65719175/mem/mem_linux.go#L155).

Looking at the output of `/proc/meminfo` on my own sysem and looking at the [kernel.org `/proc/` docs](https://www.kernel.org/doc/html/latest/filesystems/proc.html) it does look like `Writeback` is correct.

```s
❯ grep -i "writeback" /proc/meminfo 
Writeback:             0 kB
WritebackTmp:          0 kB
```

**To Reproduce**
```go
package main

import (
	"fmt"

	"github.com/shirou/gopsutil/mem"
)

func main() {
	memory, _ := mem.VirtualMemory()
	fmt.Println(memory.Writeback)
}
```

**Expected behavior**

Correctly collect the `Writeback` and `WritebackTmp` values.

**Environment (please complete the following information):**

 - [ ] Windows: [paste the result of `ver`]
 - [x] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [ ] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**

My guess, is that since the gopsutil [`VirtualMemoryStat` struct](https://github.com/shirou/gopsutil/blob/master/mem/mem.go#L55-L57) uses "WriteBack" some copy and paste or find and replace may have switched how gopsutil was looking for this value.

```s
❯ cat /etc/os-release
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
[...]

PR #1472, opened 2023-05-24
fix(mem): Correct capitalization of linux writeback

powersj:
fixes: #1471


shirou (approved):
Oh... So no one had noticed this before. Thank you very much.

Issue #465, opened 2017-12-06
network interface card speed value


monotone:
Hi,

I found the python psutil package have the `psutil.net_if_stats()` to get the network interface card speed.
```bash
>>> import psutil
>>> psutil.net_if_stats()
{'eth0': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_FULL: 2>, speed=100, mtu=1500),
 'lo': snicstats(isup=True, duplex=<NicDuplex.NIC_DUPLEX_UNKNOWN: 0>, speed=0, mtu=65536)}
```

How gopsutil get this?

shirou:
Unfortunately, current gopsutil does not have this function. I will implement but I can not say a schedule. Of curse, PR is always welcome!

monotone:
I have found in the python psutil package source code.

And try to use this C code for test. 

 It seems like if `ethtool eth0` work ,this code will work right. Otherwise it will failed.

```c
#include <errno.h>
#include <net/if.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/ioctl.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

#include <linux/ethtool.h>
#include <linux/sockios.h>

int main() {
    int ret;
    int sock = 0;
    struct ethtool_cmd ethcmd;
    struct ifreq ifr;

    sock = socket(AF_INET, SOCK_DGRAM, 0);
    if (sock == -1) {
        perror("socket AF_INET failed");
        goto error;
    }
    strncpy(ifr.ifr_name, "eth0", sizeof(ifr.ifr_name));

    // duplex and speed
    memset(&ethcmd, 0, sizeof ethcmd);
    ethcmd.cmd = ETHTOOL_GSET;
    ifr.ifr_data = (void *)&ethcmd;
    ret = ioctl(sock, SIOCETHTOOL, &ifr);

    if (ret == -1) {
        perror("ioctl SIOCETHTOOL failed");
        goto error;
    }

    printf("duplex=%d,speed=%d\n", ethcmd.duplex, ethcmd.speed);

error:
    if (sock != -1) close(sock);
    return 0;
}

```

shirou:
Thanks, but gopsutil aims to pure golang library.
I found https://github.com/safchain/ethtool. This is for only linux, but it should be a first step.

monotone:
yeah. Golang use `syscall` function to call the function `ioctl` of C.  

[...]

PR #1473, opened 2023-05-24
interface speed for windows, linux, darwin and freebsd

a-pichler:
Implemented interface speed for windows, linux, darwin, and freebsd as discussed here #465 

Windows is self-explanatory since the speed data was already in mibIfRow2

Linux reads the speed directly out of the /sys/class/net files for each interface

Darwin and Freebsd call ifconfig and parse the output (since the other data is also obtained by calling netstat and parsing it i figured that would be the best way)

### DIFF
--- a/net/net.go
+++ b/net/net.go
@@ -22,6 +22,8 @@ type IOCountersStat struct {
 	Dropout     uint64 `json:"dropout"`     // total number of outgoing packets which were dropped (always 0 on OSX and BSD)
 	Fifoin      uint64 `json:"fifoin"`      // total number of FIFO buffers errors while receiving
 	Fifoout     uint64 `json:"fifoout"`     // total number of FIFO buffers errors while sending
+	TransmitSpeed uint64 `json:"transmitSpeed"` // transmission speed of interface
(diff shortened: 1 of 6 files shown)


shirou:
Thank you for your contribution and updating PR.

However, when I run on my FreeBSD 12.4, `ifconfig` returns this value
```
vtnet1: flags=8802<BROADCAST,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=6c07bb<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,TSO4,TSO6,LRO,VLAN_HWTSO,LINKSTATE,RXCSUM_IPV6,TXCSUM_IPV6>
        ether fo:ba:12:34:56
        media: Ethernet 10Gbase-T <full-duplex>
        status: active
```
`Ethernet 10Gbase-T` is not match the regex, therefore, the test is failed. 

Could you update the PR?

1. Could you survey any other way to get interface speed other than `ifconfig` ?
2. Update regex to match `Ethernet 10Gbase-T`.
3. return error if nothing match.

I will check your PR on Windows later. Thank you for your patience.

Hipska:
Any update on this?

Issue #1474, opened 2023-05-28
Potential leak in sleep.go


hhheiaenia:
I notice the ctx.Done() needs a cancel function to unblock the statement of ctx.Done in https://github.com/shirou/gopsutil/blob/master/internal/common/sleep.go#LL13C1-L13C1,
but I never find the cancel function in its invocation.  So if this may leak the goroutine where the ctx.Done() locates? 

Maybe what I said is not right, thank you.

shirou:
After some searching, I thought your point was correct. Thank you for the suggestion, and I have opened a PR for fix it, #1475. Could you check it out? Thank you!

PR #1475, opened 2023-05-28
[common]: fix potential leak on Sleep.

shirou:
fix: #1474

References:

- [A story of a memory leak in GO: How to properly use time.After()](https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/)
- [Use "time.After" may cause a memory leak](https://github.com/eclipse/paho.mqtt.golang/issues/518)
- [Potential memory leak](https://github.com/hellofresh/health-go/issues/89)


hhheiaenia:
Thank you for fixing the issue, I think we've done this work.

PR #1486, opened 2023-06-29
cpu_percent and memory_percent Readme

elfranne:
CPUPercent and MemoryPercent works fine on Linux and Windows. Is it just a missing update to the table or am I missing some info ?


shirou (approved):
I think the documentation is simply not updated. Thank you!

PR #1488, opened 2023-07-01
[process][posix] Realign process.Name() with python psutil to return same value on python3 scripts processes

### LINKED ISSUE #1485: Why get processName from `/proc/(pid)/comm` should be at most 14 characters ?
**Describe the bug**
Why get processName from `/proc/(pid)/comm` should be at most 14 letters ?

Refs https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L795

**To Reproduce**
1. My python script named `admin_worker.py`,  just example
```python
#!/usr/local/easyops/python/bin/python
# -*- coding: utf-8 -*-

from time import sleep

while True:
  print("process...")
  sleep(1)
```

2. Start `admin_worker.py`
```shell
$ chmod +x admin_worker.py
$ ./admin_worker.py
```

3. In Linux, running my go program
```go
// `pid` should be found by `ps` command in machine.

proce, _ := process.NewProcess(int32(pid))
name, _ := proce.Name()
fmt.Println(name)
```
I find the `name` is `/usr/local/easyops/python/bin/python`

**Expected behavior**
I expect the name is `admin_worker.py`, because i use `cat /proc/(pid)/comm` in machine, i get my expected value. 

os-release
```
```

uname -a
```
```

**Additional context**

No.

Lomanic:
e2c79a1 started to blindly set the process name to the full path (instead of the basename) of the cmdline exectuable
if the process name from the process comm was truncated on linux. Python psutil never did that, and this is just wrong
for python (or any executable interpreted script) where the process name is not the interpreter binary but the script
itself.

A new test to check process name value against psutil value is added here, which would hopefully catch any potential
future changes in psutil.

Reverts #542

Fixes #1485

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -82,8 +82,6 @@ func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 			extendedName := filepath.Base(cmdName)
 			if strings.HasPrefix(extendedName, p.name) {
 				name = extendedName
-			} else {
-				name = cmdName
 			}
 		}
 	}
--- a/process/process_freebsd.go
+++ b/process/process_freebsd.go
@@ -55,8 +55,6 @@ func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 			extendedName := filepath.Base(cmdlineSlice[0])
 			if strings.HasPrefix(extendedName, p.name) {
 				name = extendedName
-			} else {
-				name = cmdlineSlice[0]
 			}
 		}
 	}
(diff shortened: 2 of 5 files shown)


shirou (approved):
Awesome! I confirmed it works on my FreeBSD 12.4-RELEASE-p1.

```
% go test -run Test_Process_Name_Against_Python -v
=== RUN   Test_Process_Name_Against_Python
    process_test.go:435: pyName python3.9
--- PASS: Test_Process_Name_Against_Python (0.16s)
PASS
ok      github.com/shirou/gopsutil/v3/process   0.166s
```

And I am very impressed with the testing idea in this PR, which is to compare directly with psutil results. Perhaps we should extend this idea to other tests.

Lomanic:
Thanks for the kind words and testing on freebsd, I was only able to test on linux. Interestingly your test shows that process comm (name) is set to the interpreter binary basename on freebsd as opposed to the script basename on linux.

Is it OK to merge or do you think we should test on more platforms?

> Perhaps we should extend this idea to other tests.

Indeed. We could have automated tests against python psutil on various platfoms, set up on CI even.

shirou:
merged. Thank you so much!

Issue #1494, opened 2023-07-25
Information related to Model, Family and VendorID are unavailable for POWER9 and above.


kishen-v:
**Bug Description**
The fields which contain information related to model, family and VendorID are unavailable for POWER9 and above generations of processors. The following is the output observed when cpu.Info() is executed.

>[{"cpu":0,"vendorId":"","family":"","model":"","stepping":2,"physicalId":"","coreId":"0","cores":1,"modelName":"POWER9 (architected), altivec supported","mhz":2500,"cacheSize":0,"flags":null,"microcode":""} }

**Expected behavior**
The expected behaviour is to have the fields related to model, family and vendor ID to be set for POWER9 and above. 

>[{"cpu":0,"vendorId":"IBM","family":"POWER","model":"POWER9","stepping":2,"physicalId":"","coreId":"0","cores":1,"modelName":"POWER9 (architected), altivec supported","mhz":2500,"cacheSize":0,"flags":null,"microcode":""}]

The following lines is where the change needs to be made.

https://github.com/shirou/gopsutil/blob/c43eaea6a71b0092b3be18dcb1bbf7b7751d1274/cpu/cpu_linux.go#L260-L263

**Environment (please complete the following information):**

 - [ ] Windows: [paste the result of `ver`]
 - [x] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
```
[root@node-test ~]# uname -a
Linux node-test.power-iaas.cloud.ibm.com 4.18.0-408.el8.ppc64le #1 SMP Mon Jul 18 16:06:55 UTC 2022 ppc64le ppc64le ppc64le GNU/Linux
```
```
[root@node-test gops]# lscpu
Architecture:        ppc64le
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  8
Core(s) per socket:  1
Socket(s):           1
NUMA node(s):        1
Model:               2.2 (pvr 004e 0202)
Model name:          POWER9 (architected), altivec supported
```
 - [ ] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**
None.

PR #1495, opened 2023-07-25
Enable setting of vendor and related information for all Power versions

kishen-v:
This PR contains the changes to set VendorID, Model and Family for all generations of POWER.
Fixes: https://github.com/shirou/gopsutil/issues/1494


shirou (approved):
Indeed. Thank you for your effective contribution to the future as well!

Issue #1283, opened 2022-04-01
BootTimeWithContext is being called for every pid causing high CPU consumption


dloucasfx:
**Description** 

User reported that our process is causing high CPU consumption, up to 500% on a linux machine with 128 CPU cores.
Profiling the process (see below image) shows that over 51% of the CPU time used by the process was spent in `gopsutil`, mainly executing the `BootTimeWithContext` func 

<img width="1741" alt="Screen Shot 2022-04-01 at 11 04 04 AM" src="https://user-images.githubusercontent.com/59848726/161299276-6f0c088d-0391-428f-9523-3708ca210086.png">

Although I was able to slightly optimize ReadLinesOffsetN [gopsutil/common.go at master · shirou/gopsutil](https://github.com/shirou/gopsutil/blob/master/internal/common/common.go#L119)  

From this:
```
~/Work/source/scratch  go test -bench=. -count 10                                                                                                                                                 ✔  9s  10:25:37 
goos: darwin
goarch: amd64
pkg: scratch
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkReadingProcStat-12        18934             57342 ns/op
BenchmarkReadingProcStat-12        20154             56708 ns/op
BenchmarkReadingProcStat-12        20383             56275 ns/op
BenchmarkReadingProcStat-12        21481             96737 ns/op
BenchmarkReadingProcStat-12        17793             61745 ns/op
BenchmarkReadingProcStat-12        20980             64534 ns/op
BenchmarkReadingProcStat-12        21688             60939 ns/op
BenchmarkReadingProcStat-12        19582             56650 ns/op
BenchmarkReadingProcStat-12        21484             55046 ns/op
BenchmarkReadingProcStat-12        21628             53929 ns/op
PASS
ok      scratch 20.430s

```
To this:

```
 ~/Work/source/scratch  go test -bench=. -count 10                                                                                                                                                ✔  21s  10:34:57 
goos: darwin
goarch: amd64
pkg: scratch
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkReadingProcStatOptimized-12               28608             41201 ns/op
BenchmarkReadingProcStatOptimized-12               25236             45126 ns/op
BenchmarkReadingProcStatOptimized-12               30998             37839 ns/op
BenchmarkReadingProcStatOptimized-12               30225             39085 ns/op
[...]

Issue #1070, opened 2021-05-10
get boot time each time cost too much cpu


hz-kelpie:
i use process.NewProcess(pid) 1000 processes per second
but common.BootTimeWithContext() runs every time,it depends on syscall.Syscall, it's blocked
maybe there's some way I don't know to cache the Process.starttime? 
thx~

Lomanic:
We can't cache host.BootTime as it can jump if time gets adjusted (like with ntpd)

Tested on a Raspberry Pi

```bash
pi@raspberrypi:~ $ grep btime /proc/stat
btime 1621271964
pi@raspberrypi:~ $ sudo systemctl disable systemd-timesyncd
pi@raspberrypi:~ $ sudo systemctl stop systemd-timesyncd
pi@raspberrypi:~ $ sudo systemctl disable fake-hwclock.service
pi@raspberrypi:~ $ sudo systemctl stop fake-hwclock.service
pi@raspberrypi:~ $ sudo reboot
# after reboot
pi@raspberrypi:~ $ grep btime /proc/stat
btime 1550139115
pi@raspberrypi:~ $ sudo systemctl enable systemd-timesyncd
pi@raspberrypi:~ $ sudo systemctl start systemd-timesyncd
pi@raspberrypi:~ $ grep btime /proc/stat
btime 1621272415
# btime jumped
```

See also #857/#837

hz-kelpie:
Thanks for your response, I think you are right
can i skip the step getting boot time , It's bit of a waste of time

hz-kelpie:
Because this method has a lock 
I can't reduce the CPU even with concurrency

ellocano:
Meet the same question. "process.NewProcessWithContext()" blocked in my situation when using pprof to monitor go routine stacks. And I found another problem,  "process.NewProcessWithContext()" seems not using context (1th param) as a timeout control threshold. Then block may happen.
![image](https://user-images.githubusercontent.com/26717749/146735876-d0ce5aad-ecff-4743-b872-d2c74264fe38.png)

cforce:
@hz-kelpie  Is this unnecessary  step getting boot time meanwhile dropped from the codebase and merged? Can this issue then get solved?

PR #1497, opened 2023-08-07
Saving boot time during ProcessesWithContext call (linux).

bjandras:
On Linux, the system boot time is used when calculating process creation times (expressed as number of seconds since the epoch). Obtaining the boot time involves parsing the /proc/stat file, which can incur siginificant overhead on systems with many CPU cores.

Permanently caching the boot time is not an option since it can change when the system clock is updated (e.g. ntpd).

In this changeset we will save the boot time during one invocation of the ProcessesWithContext function (listing all running processes).

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -30,6 +30,10 @@ const prioProcess = 0 // linux/resource.h
 
 var clockTicks = 100 // default value
 
+type bootTimeContextKeyType string
+
+const bootTimeContextKey = bootTimeContextKeyType("host_boot_time")
+
 func init() {
 	clkTck, err := sysconf.Sysconf(sysconf.SC_CLK_TCK)
 	// ignore errors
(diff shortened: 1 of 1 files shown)


bjandras:
This addresses issues #1070 and #1283.

To my knowledge, obtaining the boot time from /proc/stat is the only reliable
method. Alternative of subtracting the uptime from the current time is not
reliable since it requires obtaining these two times simultaneously.

bjandras:
Simple benchmark (just calling `process.Processes()` for `testing.B.N` times:

before:
```
goos: linux
goarch: amd64
pkg: testtest
cpu: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
BenchmarkBTime-4             196           6110134 ns/op
PASS
ok      testtest        2.272s
```

with change:
```
goos: linux
goarch: amd64
pkg: testtest
cpu: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
BenchmarkBTime-4             445           2677522 ns/op
PASS
ok      testtest        1.480s
```

(depends heavily on the number of processes running)

atoulme (approved):

bjandras:
Requested changes have been implemented.

shirou:
Ah, I am sorry that I have misled you.  What I meant was that caching in context should be done by the Application that uses gopsutil, not in the gopsutil, because it is a Library. To use a different term, we are unable to accept this PR.

bjandras:
@shirou I don't understand -- caching is done by the application, the library only uses the cached value when available.

shirou:
We do not want to add more context values. And also, bootTime can jump if time gets adjusted, described in [this comment](https://github.com/shirou/gopsutil/issues/1070#issuecomment-842512782).

PR #1498, opened 2023-08-08
mem: linux: fix vmstat field names

chouquette:
The field names are read from /proc/vmstat were capitalized as their output fields by mistake


shirou (approved):
Ah, you are absolutely right. Why have I not noticed this before? I appreciate your contribution very much.

mwdd146980:
Hey @shirou, would it be possible for you to provide an ETA on when the next release with this fix will get released? One of our customers was negatively affected by this bug and is asking for one.

shirou:
gopsutil has a [Monthly release policy](https://github.com/shirou/gopsutil/#tag-semantics). So next release will be end of August.

PR #1501, opened 2023-08-10
fix: fixed windows disk package leaks

ozanh:
- fixed goroutine leak in PartitionsWithContext
- closed registry handle in init

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"sync"
 	"syscall"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


shirou (commented):
Let me check the intent of this PR: is the intent to change the `quitChan` [to wait for the `quitChan` to close on defer with select](https://github.com/shirou/gopsutil/pull/1501/files#diff-871245bc6af04873bd2e01dd7ed020576064cf5124d49ca9df0d3c9cafc76057R157-R158) in order to prevent goroutine leaks?

If so, then indeed the goroutine continues to remain in the current implementation, and this PR seems to solve that.

ozanh:
Hi @shirou 
When the context is done before or during PartitionsWithContext, the goroutine will still be alive because it cannot send to the channel. I intended to prevent this kind of leak scenarios. Also, I used a quit channel explicitly instead of using context's Done channel in the gouroutine to make it more robust.

It could be resolved using a buffered channel as well however my fix seemed to be an idiomatic approach with channels.

Lastly, `all` parameter is not used in `PartitionsWithContext`. I don't know the background of this intention but if it seems to be a bug I can also create another PR for that.

shirou (approved):
Thank you for your explanation. I understood your intention and I agree with you. Thank you very much.

As for the `all` argument, we don't use it in the Windows implementation, but we use it on other platforms such as Linux.

PR #1507, opened 2023-08-15
Add darwin/arm64 (v2)

btoews:
This adds darwin/arm64 support in v2.


shirou:
Thank you for your contribution. But could you tell us what is the `(v2)` meaning? gopsutil v2 is not under maintenance.

btoews:
I was trying to use something on an m1 mac that depends on v2. I updated that thing to use v3 instead :-)

Issue #1139, opened 2021-09-30
Unable to get disks' serial number with IOCounters on linux


jbliao:
**Describe the bug**
`IOCountersStat.SerialNumber` always have no value on linux

I use telegraf to monitor disk which use this go module to get disk info, and the serial of disk always be null.

**To Reproduce**
```go
package main

import (
    "github.com/shirou/gopsutil/v3/disk"
    "fmt"
)

func main() {
    info, err := disk.IOCounters("/dev/sda")
    if (err != nil) {
        fmt.Print(err)
    }
    fmt.Println(info)

    serialNumber, err := disk.SerialNumber("/dev/sda")
    if (err != nil) {
        fmt.Print(err)
    }
    fmt.Println(serialNumber)

    serialNumber, err = disk.SerialNumber("sda")
    if (err != nil) {
        fmt.Print(err)
    }
    fmt.Println(serialNumber)

}                                 
```
result:
```
map[sda:{"readCount":398790,"mergedReadCount":1,"writeCount":613726,"mergedWriteCount":360,"readBytes":20676746240,"writeBytes":55764192768,"readTime":3302034,"writeTime":2927377,"iopsInProgress":0,"ioTime":6360176,"weightedIO":8757288,"name":"sda","serialNumber":"","label":""}]
INTEL_SSDSC2BW240A4_CVDA4152****<CENSORED>***
no such file or directory
```

**Expected behavior**
`disk.IOCounters` should contain serial number of disks.

**Environment (please complete the following information):**
Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]

* `cat /etc/os_release`
```
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
* `uname -a`
```
Linux truenas.local 5.10.42+truenas #1 SMP Mon Aug 30 21:54:59 UTC 2021 x86_64 GNU/Linux
```

**Additional context**
Actually this line will fail every time when it was called from IOCounters
https://github.com/shirou/gopsutil/blob/8d7a3abddb09aa082668d23e1e50cf719bbf875d/v3/disk/disk_linux.go#L444

[...]

PR #1508, opened 2023-08-17
fix IOCounters() SerialNumber enumeration

gdvalle:
Addresses https://github.com/shirou/gopsutil/issues/1139.

I think @jbliao's analysis was correct. We end up basename-ing the device names passed in, so `/dev/sda` -> `sda`, and then we try to look up the serial number using `sda`, but the first thing that happens is a `unix.Stat("sda")`, which errors silently here.

Because the API is this way, just prepend a hardcoded `/dev/` to the device name so this lookup can succeed. I think this has a minimal chance for breakage.

Alternatively, we could maybe check for a leading `/` in the `SerialNumber()` func, and prepend `/dev/` there. Probably also pretty safe, and does improve the API usability.

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -474,7 +474,11 @@ func IOCountersWithContext(ctx context.Context, names ...string) (map[string]IOC
 		}
 		d.Name = name
 
-		d.SerialNumber, _ = SerialNumberWithContext(ctx, name)
+		// Names passed in can be full paths (/dev/sda) or just device names (sda).
+		// Since `name`` here is already a basename, re-add a hardcoded /dev path.
+		// This is not ideal, but we may break the API by changing how SerialNumberWithContext
+		// works.
+		d.SerialNumber, _ = SerialNumberWithContext(ctx, "/dev/"+name)
 		d.Label, _ = LabelWithContext(ctx, name)
 
 		ret[name] = d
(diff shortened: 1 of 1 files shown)


disk/disk_linux.go
```diff
@@ -474,7 +474,11 @@ func IOCountersWithContext(ctx context.Context, names ...string) (map[string]IOC
 		}
 		d.Name = name
 
-		d.SerialNumber, _ = SerialNumberWithContext(ctx, name)
+		// Names passed in can be full paths (/dev/sda) or just device names (sda).
+		// Since `name`` here is already a basename, re-add a hardcoded /dev path.
+		// This is not ideal, but we may break the API by changing how SerialNumberWithContext
+		// works.
+		d.SerialNumber, _ = SerialNumberWithContext(ctx, "/dev/"+name)
```
shirou, line 481: Great Catch! But how about to use `common.HostDevWithContext` instead of hardcoded `/dev/` ?

```suggestion
		d.SerialNumber, _ = SerialNumberWithContext(ctx, common.HostDevWithContext(ctx, name))
```

gdvalle: Ahh, wonderful, I missed that little helper!

shirou (approved):
Great, Thank you so much!

PR #1509, opened 2023-08-18
refactor CpuUtilTotalStat as CpuUtilStat

yakuter:
This PR fixes a bug (wrong function name) in the AIX support of the CPU package. 
Screenshot 2023-08-18 at 11 40 34

### DIFF
--- a/cpu/cpu_aix_cgo.go
+++ b/cpu/cpu_aix_cgo.go
@@ -5,6 +5,7 @@ package cpu
 
 import (
 	"context"
+	"time"
 
 	"github.com/power-devops/perfstat"
 )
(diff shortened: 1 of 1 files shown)


shirou:
I am unable to confirm this using AIX's actual machine. However, `cpu.Times()` returns the "number of CPUs consumed at the moment" based on the Linux definition. Therefore, I think using `perfstat.CpuUtilTotalStat()` is correct.
Also, when `percpu` is true, `perfstat.CpuStat()` is used, which does not wait one second. Wouldn't we get different meanings when `percpu` is true and false?

I am not familiar with AIX so sorry if I am wrong.

yakuter:
Hi @shirou thank you for the comment. There is no function called **CpuUtilTotalStat** in the project. I think it is already a type. So it should be either CpuUtilStat or CpuTotalStat. As long as I get the CPU info in AIX true, they are all OK for me :)

shirou:
Hmm, it seems `CpuUtilTotalStat` is defined [here](https://github.com/power-devops/perfstat/blob/62379fc7944be7c242947c4ba8e211c2abafca73/cpustat.go#L108). And also I can build with `GOOS=aix GOARCH=ppc64`.

yakuter:
Ohh I searched in the wrong project, you are right. Thank you @shirou , that should be great 🙏

shirou:
I am closing this PR. If you have another issue, please feel free to reopen it. Thank you for using gopsutil.

PR #1512, opened 2023-08-25
Add iota for incremental data

mo1ein:


shirou:
Thanks for the contribute. However, these values are taken from Platform header files, etc. Therefore, if this PR hides the actual values by using iota, the difference from the original header file cannot be seen.

For example, linux has the following header file in `resource.h`. The number ``11`` has a meaning, but if the value is no longer written by iota, it will be difficult to know when a difference occurs.
````
(...)
  /* Maximum number of pending signals.*/
  __RLIMIT_SIGPENDING = 11,
#define RLIMIT_SIGPENDING __RLIMIT_SIGPENDING

  /* Maximum bytes in POSIX message queues.
  __RLIMIT_MSGQUEUE = 12, #define RLIMIT_SIGPENDING
#define RLIMIT_MSGQUEUE __RLIMIT_MSGQUEUE
(...)
```

PR #1515, opened 2023-08-27
[common][linux]: improve readlines performance.

shirou:
This PR will improve the following two points. This idea is based on #1514. Thank you so much.

1. Use `os.ReadFile()` and `strings.Split()` instead of `ReadLinesOffsetN()` on `common.ReadLines()`
2. Read `/proc/stat` from tail.

For 1, previously `ReadLines()` uses `ReadLinesOffsetN()`. but this function is simply too complicated when reading the entire contents of a file. So just use stdlib.
For 2, The `btime` line is usually found at the bottom part of the `stat` file. Therefore, I tried to read them in reverse order. It has little effect with my 4 CPUs, but I suspect it will work with more CPUs.

```
goos: linux
goarch: amd64
pkg: github.com/shirou/gopsutil/v3/process
cpu: AMD Ryzen 7 5800HS with Radeon Graphics
            │  before.txt  │              after.txt               │
            │    sec/op    │    sec/op     vs base                │
Processes-4   19.15m ± 12%   12.91m ± 11%  -32.61% (p=0.000 n=10)

            │  before.txt   │              after.txt               │
            │     B/op      │     B/op      vs base                │
Processes-4   3.901Mi ± 10%   2.642Mi ± 2%  -32.27% (p=0.000 n=10)

            │  before.txt  │              after.txt              │
            │  allocs/op   │  allocs/op   vs base                │
Processes-4   21.79k ± 10%   13.51k ± 2%  -38.01% (p=0.000 n=10)
```

### BenchmarkBootTimeWithManyCPUs

This PR adds `BenchmarkBootTimeWithManyCPUs` with #1514 stat file which has 80 CPU stat lines. This result seems a significant improvement in performance.

```
goos: linux
goarch: amd64
pkg: github.com/shirou/gopsutil/v3/internal/common
cpu: AMD Ryzen 7 5800HS with Radeon Graphics
                       │ before.txt  │              after.txt              │
                       │   sec/op    │   sec/op     vs base                │
BootTimeWithManyCPUs-4   19.39µ ± 4%   13.74µ ± 5%  -29.15% (p=0.000 n=10)

                       │  before.txt   │              after.txt               │
                       │     B/op      │     B/op      vs base                │
BootTimeWithManyCPUs-4   12.609Ki ± 0%   9.789Ki ± 0%  -22.37% (p=0.000 n=10)

                       │ before.txt  │             after.txt              │
                       │  allocs/op  │ allocs/op   vs base                │
BootTimeWithManyCPUs-4   103.00 ± 0%   10.00 ± 0%  -90.29% (p=0.000 n=10)
```

### TODO

We should add err [check on process_linux.go](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L1075). But it might be break application. So I will open a new PR in order to revert easily.


atoulme:
I see different results on my machine with 80 cores. Here they are:
```
goos: linux
goarch: arm64
pkg: github.com/shirou/gopsutil/v3/process
             │ old1515.txt │             new1515.txt             │
             │   sec/op    │   sec/op     vs base                │
Processes-80   249.4m ± 2%   297.3m ± 2%  +19.19% (p=0.000 n=10)
```
This is with go 1.21.0.

shirou:
I have tested with AWS c5a.16xlarge (CPU=64).  and I got no total performance improvement. hmm...

```
goos: linux
goarch: amd64
pkg: github.com/shirou/gopsutil/v3/process
cpu: AMD EPYC 7R32
             │  base.txt   │             after.txt              │
             │   sec/op    │   sec/op     vs base               │
Processes-64   37.55m ± 1%   38.32m ± 0%  +2.06% (p=0.002 n=10)

             │   base.txt   │               after.txt               │
             │     B/op     │     B/op       vs base                │
Processes-64   9.266Mi ± 0%   14.602Mi ± 0%  +57.59% (p=0.000 n=10)

             │  base.txt   │              after.txt              │
             │  allocs/op  │  allocs/op   vs base                │
Processes-64   64.68k ± 0%   25.49k ± 0%  -60.59% (p=0.000 n=10)
```

Issue #1516, opened 2023-08-28
cpu Mhz with current, max and min.


alexwbaule:
**Is your feature request related to a problem? Please describe.**
psutils from python, mas current, max and min.

**Describe the solution you'd like**
extend "Mhz" struct, with this 3 values.

**Describe alternatives you've considered**
Extend the InfoStat like this.

```
type InfoStat struct {
	CPU        int32    `json:"cpu"`
	VendorID   string   `json:"vendorId"`
	Family     string   `json:"family"`
	Model      string   `json:"model"`
	Stepping   int32    `json:"stepping"`
	PhysicalID string   `json:"physicalId"`
	CoreID     string   `json:"coreId"`
	Cores      int32    `json:"cores"`
	ModelName  string   `json:"modelName"`
	Mhz        Mhz  `json:"mhz"`
	CacheSize  int32    `json:"cacheSize"`
	Flags      []string `json:"flags"`
	Microcode  string   `json:"microcode"`
}

type Mhz struct {
      current  float64  `json:"current"` //from cpuinfo or cpufreq/cpuinfo_cur_freq 
      max       float64  `json:"max"` //from cpufreq/cpuinfo_max_freq 
      min        float64  `json:"min"`  //from cpufreq/cpuinfo_min_freq 
}

```

**Additional context**
If i can do a PR, i do. The only problem, is that i only has linux to test.

PR #1517, opened 2023-08-28
Implement Mhz to current, min and max.

alexwbaule:
Implement of https://github.com/shirou/gopsutil/issues/1516

Please, is my first PR here, and i only have Linux to test it.


shirou:
Thank you for your contribution. However, unfortunately, this PR breaks current compatibility. So we cannot accept while in current major version, v3. I have added `v4` label, but v4 release date is not discussed yet.

alexwbaule:
Hi @shirou , can we merge this ?

shirou:
As I commented, this PR breaks current compatibility. So we cannot merge until v4 is released.

kkartaltepe:
It seems the current version is v4 now can this be considered again? I was also interested in this data.

alexwbaule:
@shirou , need something to merge ?

renard:
@alexwbaule I +1 this feature because it eases the ability to figure out any glitches during a benchmark. However I'd recommend to keep a flat structure such as:

```
MhzCurrent float64 `json:"mhz_current"`
MhzMin float64 `json:"mhz_min"`
MhzMax float64 `json:"mhz_max"`
```
instead of using a `MHz` structure. 

So far no other collected metric use any sub-structure. It would be better IMHO to keep it that way.

cpu/cpu.go
```diff
@@ -40,12 +40,18 @@ type InfoStat struct {
...
 	Cores      int32    `json:"cores"`
 	ModelName  string   `json:"modelName"`
-	Mhz        float64  `json:"mhz"`
+	Mhz        Mhz      `json:"mhz"`
 	CacheSize  int32    `json:"cacheSize"`
 	Flags      []string `json:"flags"`
 	Microcode  string   `json:"microcode"`
 }
 
+type Mhz struct {
```
ccoVeille, line 49: I would use MHz

And while you are changing the type and so the output format, I would change the Mhz variable name in the struct

But here,I'm only a random Gopher.

The maintainers of this project might have another opinion

cpu/cpu_freebsd.go
```diff
@@ -103,7 +103,9 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 	if u32, err = unix.SysctlUint32("hw.clockrate"); err != nil {
 		return nil, err
 	}
-	c.Mhz = float64(u32)
+	c.Mhz.Current = float64(u32)
+	c.Mhz.Min = 0
+	c.Mhz.Max = 0
```
ccoVeille, line 109: At some point if it's for test purpose that you are resetting it

cpu/cpu_freebsd.go
```diff
@@ -103,7 +103,9 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
...
```
ccoVeille, line 108: Please consider writing this


```suggestion
	c.Mhz = Mhz{Current: float64(u32)}
```

Abd everywhere you did the replacement

cpu/cpu_linux.go
```diff
@@ -142,23 +141,35 @@ func finishCPUInfo(ctx context.Context, c *InfoStat) {
...
-	value, err = strconv.ParseFloat(lines[0], 64)
-	if err != nil {
-		return
-	}
-	c.Mhz = value / 1000.0 // value is in kHz
-	if c.Mhz > 9999 {
-		c.Mhz = c.Mhz / 1000.0 // value in Hz
+	c.Mhz.Current = fillMhz(ctx, "cur", c)
+	c.Mhz.Min = fillMhz(ctx, "min", c)
+	c.Mhz.Max = fillMhz(ctx, "max", c)
```
ccoVeille, line 146: I would pass


```suggestion
	c.Mhz.Current = fillMhz(ctx, "cpuinfo_cur_freq", c)
	c.Mhz.Min = fillMhz(ctx, "cpuinfo_min_freq", c)
	c.Mhz.Max = fillMhz(ctx, "cpuinfo_max_freq", c)
```

And then I would use

```suggestion
	c.Mhz = Mhz{Current: fillMhz(ctx, "cpuinfo_cur_freq", c), Min: fillMhz(ctx, "cpuinfo_min_freq", c), Max: fillMhz(ctx, "cpuinfo_max_freq", c)}
```

cpu/cpu_linux.go
```diff
@@ -142,23 +141,35 @@ func finishCPUInfo(ctx context.Context, c *InfoStat) {
...
+
+}
+
+func fillMhz(ctx context.Context, value string, c *InfoStat) float64 {
+	var lines []string
+	var err error
+	var line float64
+	var mhz float64 = 0
+
+	if value == "min" || value == "max" || value == "cur" {
```
ccoVeille, line 156: Add a switch here and return 0 if the value are not the one you expect

ccoVeille (commented):
It's a breaking change but why not

PR #1519, opened 2023-08-31
Add support for reading AnonHugePages from /proc/meminfo

kevinconaway:
This commit adds support for reading the `AnonHugePages` field from `/proc/meminfo`.

The values in this field allow monitoring the [THP](https://www.kernel.org/doc/Documentation/vm/transhuge.txt) usage by systems that use this type of memory


shirou (approved):
We have added `HugePagesRsvd` and `HugePagesSurp` in #1275. As I stated in the PR, we do not want to add any more Linux-specific information. However, since information related to HugePages has already been added, I think we should unfortunately also add this `AnonHugePages` field. Anyway, I am very appreciative of your contribution.

PR #1530, opened 2023-09-23
Add some support for NetBSD

darkgeek:
Hi,

I want to add support for `NetBSD` in this PR, where I implement some functions in `disk/mem/cpu/host`. Though not all functions are implemented, it's a decent start. The `cpu` part is largely inspired by the #1364 .

I use a `Nano Pi R5S` aarch64 board, where `NetBSD 10` is running, to test my code, and all the functions I implement can work well there.

So can you check this PR to see if it's ok to merge? Thanks.


darkgeek:
I see some github action checks above failed, all of which are macos related. After digging into the check details, I couldn't find any relation between the errors and my commits😂️.

shirou (approved):
hmm. [dependabot PR also failed](https://github.com/shirou/gopsutil/actions/runs/6293637804/job/17084418589?pr=1531) with same error, "invalid nice: -10". perhaps some of GitHub image have changed? I will check later.

Anyway, thank you for your great contribution!
I can not confirm it works or not on NetBSD machine, but it seems almost same as FreeBSD, then it should work.

PR #1544, opened 2023-10-30
fix(linux): validate cpu fields length before accessing index

JanDeDobbeleer:
Two fixes in this one:
- validate the fields length before assuming we have everything. This is related to https://github.com/JanDeDobbeleer/oh-my-posh/issues/4397
- do not print things to stdout on Windows. This is related to https://github.com/JanDeDobbeleer/oh-my-posh/issues/2947


shirou (approved):
Great patch. Thank you for your contribution!

PR #1551, opened 2023-11-09
feat: use lsof for net_connections on FreeBSD

wolf31o2:
Use net.ConnectionsPidWithContext on FreeBSD, similarly to how it is done on Darwin. This uses common.CallLsofWithContext underneath the hood, which will use lsof under the hood, if available.

Tested on FreeBSD 13.2-RELEASE


shirou (approved):
Thank you so much! I have confirmed this awesome PR works on my FreeBSD-12.4.

PR #1552, opened 2023-11-16
[process] Add support for cgroups

atoulme:
Add support for cgroups - only on Linux, all other OSes do not expose this information.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -26,6 +26,7 @@ type Process struct {
 	Pid            int32 `json:"pid"`
 	name           string
 	status         string
+	cgroup         string
 	parent         int32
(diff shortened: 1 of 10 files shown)


astencel-sumo (approved):
Makes sense to me 👍

atoulme:
@shirou wdyt?

shirou:
I don't want to add Linux specific information anymore. However, I am thinking about [v4](https://github.com/shirou/gopsutil/discussions/1553) to have a room about this kind of information. If this implementation works well, I would like to add these information in it.

atoulme:
I see, we will work on this separately then.

PR #1561, opened 2023-12-13
add deepin and uos distro

keeword:
Make host to recognize deepin and uos distro.


shirou (approved):
I have not checked it works, but it looks good to me. Thank you for your contribution!

PR #1569, opened 2023-12-21
Avoid repeated regexp compilations

scop:


shirou (approved):
Good catch! Thank you!

PR #1570, opened 2023-12-21
Avoid some uses of regexps

scop:
Untested, noticed while looking into something else.


shirou (approved):
Indeed. Thank you so much!

PR #1573, opened 2023-12-25
feat(disk): look for filesystem labels from udev on Linux

scop:


shirou (approved):
As for `Label`, if there is no `block/%s/dm/name`, then `udev` is used, which does not seem to break the existing implementation. Your continued contributions are very much appreciated.

PR #1578, opened 2024-01-09
add support for sw64

weidongkl:


shirou:
Could you tell us what is the `sw64` arch ? It seems [no such an architecture on Golang](https://go.dev/doc/install/source#environment), at least Go 1.22.

weidongkl:
> Could you tell us what is the `sw64` arch ? It seems [no such an architecture on Golang](https://go.dev/doc/install/source#environment), at least Go 1.22.

https://zh.wikipedia.org/wiki/%E7%94%B3%E5%A8%81
```bash
$ go env GOHOSTARCH
sw64
$ go env GOARCH
sw64
$ go env GOHOSTOS
linux
$ go env GOTOOLDIR
/usr/lib/golang/pkg/tool/linux_sw64
$ 
```

PR #1580, opened 2024-01-11
Windows, read all PIDs if there are more than 1024 PIDs.

jnewmano:
Convert bytes read to number of uint32s that were read.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -253,7 +253,7 @@ func pidsWithContext(ctx context.Context) ([]int32, error) {
 		if err := windows.EnumProcesses(ps, &read); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


shirou (approved):
Indeed. Thank you for your contribution!

Lomanic:
This amends #454, confirmed with the test I described there (replacing the 1024 magic numbers with a small value, like 32, and adding some debug in the loop logging `psSize` to see it increase). I'm not sure to understand why my test worked at the time, in the meantime the w32 dependency was removed, [but it doesn't look suspicious](https://github.com/shirou/w32/blob/bb4de0191aa41b5507caa14b0650cdbddcd9280b/psapi.go#L20-L26)? Anyway, thanks for fixing this long standing bug @jnewmano

PR #1584, opened 2024-01-18
ensure host platform are files and have contents

brycekahle:
In a containerized deployment, it is common to mount several files from `/etc`. Within the container, those files will be created regardless if they exist on the host or not. In those instances, the existing code would erroneously return empty platform information.


shirou (approved):
I see. I was not aware of that possibility. Thank you very much!

Issue #1576, opened 2024-01-02
Adopt a hierarchy for getting distro name/version on Linux 


joshuar:
**Is your feature request related to a problem? Please describe.**
On Linux, there are many difference files that can provide distribution details. gopsutil should adopt a hierarchy for checking those files, using the information from the first file found. 

**Describe the solution you'd like**
While gopsutil does currently adopt this kind of approach, the order of checks could be improved. My suggestion would be something like:

- `/etc/os-release`
- `/etc/lsb-release`
- `/etc/*[version|release]`

Above, preference the general standards/guideline files first, then the distro-specific files. `/etc/os-release` appears to be adopted widely, and often contains much more specific information (regarding distribution variants) than the other sources. While `/etc/lsb-release` usage appears to be waning and discouraged, it is still an important source for this information. The remaining files such as  `/etc/redhat-release`, `/etc/SuSE-release`, `/etc/debian_version`, `/etc/arch-release`, `/etc/gentoo-release`, `/etc/slackware-version`, `/etc/frugalware-release`, `/etc/altlinux-release`, `/etc/mandriva-release`, `/etc/meego-release`, `/etc/angstrom-version`, `/etc/mageia-release` may or may not contain variant-specific information. 

For example, I would image a lot of variants based on Debian do not edit `/etc/debian_version` and so relying on this first will report the wrong information (the debian base version, not the variant version). Likewise, if I was going to build my own distribution based on Fedora, I would likely leave `/etc/redhat-release` untouched and place my distribution information in the `/etc/os-release` file.

**Describe alternatives you've considered**
For my own usage, my alternative would be to just write my own code to apply the above order.

**Additional context**
Preferencing `/etc/os-release` over `/etc/debian_release` for example should fix #1469 and #224 and likely avoid similar issues in the future.

PR #1585, opened 2024-01-18
use VERSION_ID from os-release

brycekahle:
`VERSION_ID` is more appropriate for scripts and other usages, since `VERSION` can contain spaces and codenames

from `os-release` manpage:
```
VERSION=
           A string identifying the operating system version, excluding
           any OS name information, possibly including a release code
           name, and suitable for presentation to the user. This field
           is optional.

           Examples: "VERSION=17", "VERSION="17 (Beefy Miracle)"".

 VERSION_ID=
           A lower-case string (mostly numeric, no spaces or other
           characters outside of 0–9, a–z, ".", "_" and "-") identifying
           the operating system version, excluding any OS name
           information or release code name, and suitable for processing
           by scripts or usage in generated filenames. This field is
           optional.

           Examples: "VERSION_ID=17", "VERSION_ID=11.04".
```

### DIFF
--- a/internal/common/common_linux.go
+++ b/internal/common/common_linux.go
@@ -329,7 +329,7 @@ func GetOSReleaseWithContext(ctx context.Context) (platform string, version stri
 		switch field[0] {
 		case "ID": // use ID for lowercase
 			platform = trimQuotes(field[1])
-		case "VERSION":
+		case "VERSION_ID":
 			version = trimQuotes(field[1])
 		}
 	}


shirou (approved):
When I first looked at this PR, I thought that this change would have a very broad impact.

However, after reading [the comments in the linked PR](https://github.com/DataDog/gopsutil/pull/44#issue-1433558759), I realized that it wouldn't really matter, since the priority for `os-release` is lower than for files like `lsb-release` and distribution files(e.g. `debian_version`).

One thing to note is that there is an issue #1576. In this case, os-release has the highest priority, so the impact will be significant. However, I approve this PR because I think this `version` value should actually return the `VERSION_ID`.

Your contribution is very much appreciated.

PR #1588, opened 2024-01-26
memory: expose virtual totals / avail values on windows

sni:
There are 2 more memory values i'd like to use. They are there already, just not exposed by the module yet.

- https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-memorystatusex


shirou:
Thank you for your contribute. However, we don't want to add platform specific fields anymore. We have a [v4 plan](https://github.com/shirou/gopsutil/discussions/1553) which introduces platform specific structs. We will be able to add fields on it, but not developed yet.

sni:
that's great, thank you very much.

PR #1589, opened 2024-01-26
temperature sensors: add low value from ..._min file

sni:
add an optional low value from the ./*_min file.

```
temp[1-*]_min	Temperature min value.
		Unit: millidegree Celsius
		RW
```


sni:
perfect, thanks for your work.

PR #1594, opened 2024-02-11
add support for OpenBSD/riscv64

jmatthew:
This is the output of mktypes.sh on a StarFive VisionFive 2 board running OpenBSD -current, slightly adjusted so it builds and the tests mostly work. While I was there, I updated the arch list for OpenBSD in README.md.


shirou (approved):
I don't have such a machine, but this PR looks very promising. Thanks a lot for your contribution!

PR #1607, opened 2024-03-05
Failed to obtain virtual volume in windows

gorpher:
[fix: disk Partition error in windows](https://github.com/shirou/gopsutil/commit/8a1e07f1c8c896a2b1a06d5d260bdf29f8fe05f4)


shirou:
The content of this PR appears to ignore the error, rather than the title "virtual volume in Windows". I am very sorry, but could you please explain why this PR makes it possible to use Virtual Volume, including the source of the information?

PR #1612, opened 2024-03-21
Total Processes in `MiscStat` Corrected

### LINKED ISSUE #1606: ProcTotal Inaccurate in `MiscStats`
**Describe the bug**

PR #682 added the ability to get the total number of processes via the `MiscStats` struct. It gets this information by reading the `/proc/loadavg` file which looks like:

```
0.35 0.22 0.18 2/1095 5671
```

The [proc man page](https://www.man7.org/linux/man-pages/man5/proc.5.html) describes that file as:

> The first three fields in this file are load average figures giving the number of jobs in the run queue (state R) or waiting for disk I/O (state D) averaged over 1, 5, and 15 minutes.  They are the same as the load average numbers given by [uptime(1)](https://www.man7.org/linux/man-pages/man1/uptime.1.html) and other programs.  **The fourth field consists of two numbers separated by a slash (/). The first of these is the number of currently runnable kernel scheduling entities (processes, threads).  The value after the slash is the number of kernel scheduling entities that currently exist on the system.**  The fifth field is the PID of the process that was most recently created on the system.

The bolded text is the key part. In my example above, it's showing 1095 as the total number of kernel scheduled entities. But note it uses that term instead of "processes". It even indicates in parentheses that "kernel scheduled entities" is both processes and threads. But by the name `ProcTotal` and the fact that all other numbers in `MiscStat` are processes only (not threads) this is implying I have 1095 processes when I do not. Thread count varies greatly. If I run it just a few minutes later it's showing 1127:

```
0.22 0.27 0.22 1/1127 5877
```

Also if we actually count the processes in my system we find we only have a bit over 300:

```sh
%> $ ps -A | wc -l
307
```

**Expected behavior**

I would expect `ProcsTotal` to return the total number of processes not processes and threads. I don't know if there is something else that can be read in `/proc` that gives just the total number of processes (I couldn't find anything) or if the attribute needs to be deprecated and removed for being in-accurate.

I came across this due to [OpenTelemtry using this struct](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processesscraper/processes_scraper_unix.go#L45-L73) to help measure the number of processes in the system and getting wildly high numbers. It actually [calculates this number itself by reading all directories under `/proc` that are number](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L1151-L1174). But then it only uses that as a starting place and offsets it with what this struct is providing [saying](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processesscraper/processes_scraper_unix.go#L45-L47):

>  Processes are actively changing as we run this code, so this reason the above loop will tend to underestimate process counts. getMiscStats is a single read/syscall so it should be more accurate.

Based on that I don't think the OpenTelemetry strategy will be sufficient to get an accurate process count.

```sh
$ cat /etc/os-release
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os
```

```sh
$ uname -a
Linux personal 6.6.10-76060610-generic #202401051437~1704728131~22.04~24d69e2 SMP PREEMPT_DYNAMIC Mon J x86_64 x86_64 x86_64 GNU/Linux
```

eric1234:
The `ProcsTotal` in the `MiscStat` structure was very inaccurate. It was reading a value which is the total number of kernel scheduling entities. This includes both processes and threads significantly overcounting.

This instead uses an existing method already in common to count the number of processes via the /proc filesystem where any directory is a number. This should still be a single syscall to read that directory entry.

I don't see any existing automated test with this functionality so I did not add any but I did do my own manual testing by writing the following program:

```go
package main

import (
	"fmt"
	"github.com/shirou/gopsutil/v3/load"
)

func main() {
	stats, _ := load.Misc()
	fmt.Printf("Count: %v\n", stats.ProcsTotal)
}
```

With the previous code it was returning a value of over 1200. After the change the count was around 320 and matched `ps -A | wc -l`.

This fixes #1606.

### DIFF
--- a/load/load_linux.go
+++ b/load/load_linux.go
@@ -107,21 +107,13 @@ func MiscWithContext(ctx context.Context) (*MiscStat, error) {
 
 	}
 
-	procsTotal, err := getProcsTotal(ctx)
+	procsTotal, err := common.NumProcsWithContext(ctx)
 	if err != nil {
 		return ret, err
 	}
 	ret.ProcsTotal = int(procsTotal)
 
 	return ret, nil
-}
-
-func getProcsTotal(ctx context.Context) (int64, error) {
-	values, err := readLoadAvgFromFile(ctx)
-	if err != nil {
-		return 0, err
-	}
-	return strconv.ParseInt(strings.Split(values[3], "/")[1], 10, 64)
 }
 
 func readLoadAvgFromFile(ctx context.Context) ([]string, error) {


shirou (approved):
Great work. Thank you so much!

PR #1616, opened 2024-03-27
Refresh boot time

braydonk:
This PR adds a function to `process` and `host` to manually refresh the cached boot time value. This is very helpful in scenarios where you are reading all the processes on the system at a given time, want the performance gains of using the boot time cache, but want to update the cached boot time when you read all the processes.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -69,6 +69,12 @@ func EnableBootTimeCache(enable bool) {
 	enableBootTimeCache = enable
 }
 
+// RefreshBootTimeCache manually refreshes the cached BootTime value.
+func RefreshBootTimeCache(ctx context.Context) error {
+	_, err := common.BootTimeWithContext(ctx, false)
+	return err
+}
+
 func Info() (*InfoStat, error) {
 	return InfoWithContext(context.Background())
 }
--- a/process/process.go
+++ b/process/process.go
@@ -178,6 +178,12 @@ func EnableBootTimeCache(enable bool) {
 	enableBootTimeCache = enable
 }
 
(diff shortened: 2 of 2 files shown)


shirou:
I think the same could be accomplished with the following code on a library user side. I don't think it is necessary to add a new API.

```go
host.EnableBootTimeCache(false)
defer host.EnableBootTimeCache(true)

BootTimeWithContext(ctx)
```

braydonk:
> I think the same could be accomplished with the following code on a library user side.

That would be better if it was possible, but [BootTimeWithContext is internal](https://github.com/shirou/gopsutil/blob/e230f528f075f78e713f167c28b692cc15307d19/internal/common/common_linux.go#L63). So I don't see any way on the user side to only read boot time once.

braydonk:
I will work on an alternative PR that instead factors boot time into the common package and provides entrypoints so that your suggestion is possible.

shirou:
`host.BootTimeWithContext` is a [public function](https://pkg.go.dev/github.com/shirou/gopsutil@v3.21.11+incompatible/host#BootTimeWithContext). Here is the complete code.

```go
package main

import (
	"context"
	"fmt"

	"github.com/shirou/gopsutil/v3/host"
)

func main() {
	host.EnableBootTimeCache(false)
	defer host.EnableBootTimeCache(true)

	fmt.Println(host.BootTimeWithContext(context.Background()))
}
```

braydonk:
Closing in favour of #1618

braydonk:
Ah okay so using the `host` function for boot time refreshes the same cache. I'll close the other PR too then.

PR #1618, opened 2024-04-02
process: Add exported BootTimeWithContext

braydonk:
This PR adds a publicly accessible `BootTimeWithContext` in the `linux` part of the process package. There is already the boot time cache to deal with boot time being read over and over again for every single process, and this exported function will allow the user to manually refresh the cached boot time if necessary by doing the following:

```go
process.EnableBootTimeCache(false)
process.BootTimeWithContext(ctx)
process.EnableBootTimeCache(true)
```


braydonk:
Closing as this is already possible using the host version: https://github.com/shirou/gopsutil/pull/1616#issuecomment-2032112231

PR #1619, opened 2024-04-04
Increasing precision for virtual memory used percent on windows

rayjanoka:
Hey! Nice library!!

I found that the `virtualMemory.UsedPercent` returned by windows is just a whole number and we lose some precision there. I computed it instead using `total - available / total * 100` so we can get the same precision we get on linux.

```
virtMemTest - Old 28.000000 : New 28.606437
virtMemTest - Old 28.000000 : New 28.594666
virtMemTest - Old 28.000000 : New 28.637374
```

### DIFF
--- a/mem/mem_windows.go
+++ b/mem/mem_windows.go
@@ -48,7 +48,7 @@ func VirtualMemoryWithContext(ctx context.Context) (*VirtualMemoryStat, error) {
 		Total:       memInfo.ullTotalPhys,
 		Available:   memInfo.ullAvailPhys,
 		Free:        memInfo.ullAvailPhys,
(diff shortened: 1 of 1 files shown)


shirou:
Good point. Thanks you for opening this PR. 

However, I noticed that [psutil](https://github.com/giampaolo/psutil/blob/master/psutil/arch/windows/mem.c#L24-L44) uses [PERFORMANCE_INFORMATION](https://learn.microsoft.com/en-us/windows/win32/api/psapi/ns-psapi-performance_information) instead of [MEMORYSTATUSEX](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-memorystatusex). So I think we should change that. After the change, the precision issue will be solved.

Issue #1622, opened 2024-04-08
`ExeWithContext` returns an error with a Go binary on newer versions of macOS


MDrakos:
**Describe the bug**
The `ExeWithContext` function [here](https://github.com/shirou/gopsutil/blob/master/process/process_darwin_nocgo.go#L20) expects a specific number of `ftxt` entries before it extract the executable file path. It _appears_ that on newer versions of macOS this output has changed.

**To Reproduce**
Passing the following code a running Go binary creates an error. I haven't tested with other running processes.

```bash
❯ CGO_ENABLED=0 go run main.go 64030
Could not get executable of process: missing txt data returned by lsof
exit status 1
```

```go
package main

import (
	"errors"
	"fmt"
	"os"
	"strconv"

	"github.com/shirou/gopsutil/v3/process"
)

func main() {
	if err := run(); err != nil {
		fmt.Println(err)
		os.Exit(1)
	}
}

func run() error {
	if len(os.Args) != 2 {
		return errors.New("Usage: process <pid>")
	}

	pidArg := os.Args[1]
	pid, err := strconv.Atoi(pidArg)
	if err != nil {
		return fmt.Errorf("Could not parse pid: %w", err)
	}

	proc, err := process.NewProcess(int32(pid))
	if err != nil {
		return fmt.Errorf("Could not create process: %w", err)
	}

	_, err = proc.Exe()
	if err != nil {
		return fmt.Errorf("Could not get executable of process: %w", err)
	}

	return nil
}
```

When I run the `lsof` command I can see only one `ftxt` line in the output.

**Expected behavior**
The `proc.Exe()` call does not fail and returns the executable file path.

**Environment (please complete the following information):**
 - [ ] Windows: [paste the result of `ver`]
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [x] Mac OS: [paste the result of `sw_vers` and `uname -a`
 ```
~
❯ sw_vers
ProductName:		macOS
ProductVersion:		14.4.1
BuildVersion:		23E224

~
❯ uname -a
Darwin M1-MacBook-Pro-6.local 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000 arm64
```
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
[...]

PR #1640, opened 2024-05-06
Add fallback for lsof output

MDrakos:
This PR is to fix address the issue outlined here: https://github.com/shirou/gopsutil/issues/1622


shirou (approved):
It looks great to me. Thank you very much for your valuable contribution!

PR #1646, opened 2024-05-12
remove duplicate code in mktypes.sh

zhanluxianshen:
/kind cleanup

remove duplicate code in mktypes.sh


shirou (approved):
good catch! Thanks!

zhanluxianshen:
Thanks

Issue #1425, opened 2023-03-02
Missing support for riscv64 FreeBSD


clausecker:
The package is missing support code for riscv64 FreeBSD.  Please add such code:

```
# github.com/shirou/gopsutil/disk                          
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:108:39: undefined: sizeOfDevstat                                
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:113:14: undefined: sizeOfDevstat                                
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:126:29: undefined: DEVSTAT_READ                                 
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:127:29: undefined: DEVSTAT_WRITE                                
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:128:24: undefined: DEVSTAT_READ                                 
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:129:24: undefined: DEVSTAT_WRITE                                
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:130:34: undefined: DEVSTAT_READ                                 
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:131:34: undefined: DEVSTAT_WRITE                                
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:141:9: undefined: Bintime                                       
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:148:32: undefined: Devstat                                      
vendor/github.com/shirou/gopsutil/disk/disk_freebsd.go:131:34: too many errors   
```

Lomanic:
If you have a riscv64 FreeBSD host with a Go compiler, please run `mktypes.sh` and open a merge request with the generated files.

clausecker:
The script doesn't run when executed on a freshly cloned checkout.

```
$ sh mktypes.sh 
/home/fuz/src/gopsutil/host/types_freebsd.go:15:10: fatal error: 'freebsd_headers/utxdb.h' file not found
#include "freebsd_headers/utxdb.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
cd: v3: No such file or directory
```

Lomanic:
[...]

PR #1652, opened 2024-05-18
Add type definitions for riscv64 FreeBSD

clausecker:
For #1425.  This enables gopsutil to work on riscv64 FreeBSD.

I also fixed an include directive so mktypes.sh would succeed.


clausecker:
Currently fixing some bugs with the patch set.  The Tv field in Utmpx should be a Timeval (which is a structure), but it's an uint64 in all the files for some reason.  This is wrong and will lead to wrong results.  Apparently this was changed in 2fd3f03f8455356d7e0034fbeed6356dc15c2499 and I don't know why.

clausecker:
Also note that the utmp structure no longer exists, it appears to be a legacy of FreeBSD 9.

shirou:
Thank you for your contribution! So are you still trying to edit this PR related to `Tv` in `Utmpx`? Then, we will wait to approve.

> Also note that the utmp structure no longer exists, it appears to be a legacy of FreeBSD 9.

Yes, it is just for a [compatibility](https://github.com/shirou/gopsutil/blob/master/host/types_freebsd.go#L45). But, perhaps, we can remove because FreeBSD 9 was released 10 years ago.

clausecker:
> So are you still trying to edit this PR related to Tv in Utmpx?

I'm not sure what to do here.  The `Tv` field is wrong on the other 64 bit architectures, too.  It is always a structure, not a 64 bit integer.  If I was to fix the behaviour of this field for riscv64, I would have to touch all other architectures, too.

> Yes, it is just for a [compatibility](https://github.com/shirou/gopsutil/blob/master/host/types_freebsd.go#L45). But, perhaps, we can remove because FreeBSD 9 was released 10 years ago.

Maybe that would be a good idea indeed.

shirou:
When I run build test by `GOOS=freebsd GOARCH=riscv64 go test ./...`, build was failed on `host`.

```
host/host_freebsd_riscv64.go:26:8: undefined: _cgopackage
host/host_freebsd.go:73:31: u.Tv undefined (type Utmpx has no field or method Tv)
```

I found #709 and remember why `freebsd_headers` is included and Tv is not a struct.

Regarding the `u.Tv undefined` error in utmpx.Tv, it seems that `uint64_t` in `utxdb.h` cannot be resolved, resulting in Pad_cgo_0. I have tried various approaches to solve this issue, but it also results in Pad_cgo_0 on amd64. Do you have any idea why this is happening? I believe it was not an issue when it was generated in the past.

PR #1655, opened 2024-05-26
Fix boot time not returning stat file value

govrin:
When `BootTimeWithContext` is called while cache is disabled, the value retrieved from reading the stat file isn't used. Also,  I moved the call to `time.Now()` closer to where the `uptime` stat is read, to reduce the time between them. This still isn't perfect.


shirou (approved):
This PR means "the value of `useStatFile` is not being used" and that seems correct. It only starts being used from the second time onward when caching is enabled. Indeed, now that you mention it... thank you!

The line for `currentTime` follows the principle of "defining it close to where it is used," but you might be right that it should be moved closer as there are a few steps in between.

govrin:
golangci-lint said something about the file not being `gci`-ed, but when I ran it locally it didn't show this error. I also tried to run `gci` with the correct settings and it didn't fix the file. Can you run the CI again?

shirou:
Great. Thanks a lot!

PR #1656, opened 2024-06-02
fix memory leak

thechampagne:


shirou (approved):

PR #1670, opened 2024-07-03
fix: docker_memory in centos_7.5 & docker_1.13.1

coderZoe:
bugfix：docker_memory in centos_7.5 & docker_1.13.1


shirou (approved):
Thank you for pointing it out. [kernel.org cgrou-1 document](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt) says `rss_huge` and `mapped_file` are used.

Issue #1687, opened 2024-08-09
The names passed into IOCountersWithContext func in disk_windows.go is not used to filter devices


zmyzheng:
**Describe the bug**

The IOCountersWithContext func in disk_windows.go accepts a list of device names as the filter, however, the windows imlementation of this func does not implements the filter.

**To Reproduce**
```go
stats, err := disk.IOCounters(deviceNames...)
```

**Expected behavior**
The IOCounter returned should filter device names

**Environment (please complete the following information):**
 - [X] Windows: [paste the result of `ver`] windows server 2022
 - [ ] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]
 - [ ] Mac OS: [paste the result of `sw_vers` and `uname -a`
 - [ ] FreeBSD: [paste the result of `freebsd-version -k -r -u` and  `uname -a`]
 - [ ] OpenBSD: [paste the result of `uname -a`]

**Additional context**

PR #1688, opened 2024-08-09
fix(IOCountersWithContext): Fix the issue of not using names to fil ter devices in func IOCountersWithContext of the windows implementation

zmyzheng:
The IOCountersWithContext func in disk_windows.go accepts a list of device names as the filter, however, the windows imlementation of this func does not implements the filter. This PR fixes this issue. 

Note: this issue only happens in the windows env
Issue #1687

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -218,14 +218,17 @@ func IOCountersWithContext(ctx context.Context, names ...string) (map[string]IOC
 			if err != nil {
 				return drivemap, err
 			}
-			drivemap[path] = IOCountersStat{
-				ReadBytes:  uint64(diskPerformance.BytesRead),
-				WriteBytes: uint64(diskPerformance.BytesWritten),
-				ReadCount:  uint64(diskPerformance.ReadCount),
-				WriteCount: uint64(diskPerformance.WriteCount),
-				ReadTime:   uint64(diskPerformance.ReadTime / 10000 / 1000), // convert to ms: https://github.com/giampaolo/psutil/issues/1012
-				WriteTime:  uint64(diskPerformance.WriteTime / 10000 / 1000),
-				Name:       path,
+
+			if len(names) == 0 || common.StringsHas(names, path) {
+				drivemap[path] = IOCountersStat{
+					ReadBytes:  uint64(diskPerformance.BytesRead),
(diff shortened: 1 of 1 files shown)


disk/disk_windows.go
```diff
@@ -218,14 +218,17 @@ func IOCountersWithContext(ctx context.Context, names ...string) (map[string]IOC
...
-			drivemap[path] = IOCountersStat{
-				ReadBytes:  uint64(diskPerformance.BytesRead),
-				WriteBytes: uint64(diskPerformance.BytesWritten),
-				ReadCount:  uint64(diskPerformance.ReadCount),
-				WriteCount: uint64(diskPerformance.WriteCount),
-				ReadTime:   uint64(diskPerformance.ReadTime / 10000 / 1000), // convert to ms: https://github.com/giampaolo/psutil/issues/1012
-				WriteTime:  uint64(diskPerformance.WriteTime / 10000 / 1000),
-				Name:       path,
+
+			if len(names) == 0 || common.StringsHas(names, path) {
```
shirou, line 222: If `names` is empty slice, I think `common.StringsHas` always return `false`. So no need to check `len(names) == 0`.

```suggestion
			if common.StringsHas(names, path) {
```

shirou (changes requested):
Ah, great catch! I have requested a little change, but I will merge just after you fix it. Thank you so much!

disk/disk_windows.go
zmyzheng: Hi @shirou , thanks for the comment, but my understanding is the if clause should return `true` rather than `false` if `names` is empty. In other words, the logic will skip filtering by `names` and return all disks when  `names` is empty. This is also how the linux version and darwin version is implemented today:
- Darwin: https://github.com/shirou/gopsutil/blob/master/disk/disk_darwin_cgo.go#L38
- Linux: https://github.com/shirou/gopsutil/blob/master/disk/disk_linux.go#L409

shirou: Ah, I apologize. I was mistaken (reviewing when tired is not a good idea). This feature is a "filter", so if there are no conditions, it should add everything. You are indeed correct.

shirou (approved):

PR #1691, opened 2024-08-14
Fix parsing of /proc/pid/smaps when path is empty

pgimalac:
As per the [documentation](https://www.kernel.org/doc/html/latest/filesystems/proc.html) of `/proc`, the path in `/proc/<pid>/maps` and `/proc/<pid>/smaps` is the 6th field of the line, and can be empty:

> The “pathname” shows the name associated file for this mapping. [...] if empty, the mapping is anonymous.

Rather than using the last field returned by `strings.Field`, using the 6th element of the slice if it exists is more correct, as it handles properly the case where the path is empty.

This PR also adds a test with a fake `smaps` file to ensure parsing is correct.


shirou (approved):
Yes, I agree with you. Thank you for your contribution!

Issue #1636, opened 2024-05-02
Crash in createTimeWithContext when run on FreeBSD arm64


sdalu:
**Describe the bug**
When running telegraph 1.31.0~553d972c, with gopsutil v3.24.4, it crash in the createTimeWithContext.

~~~
2024-05-02T13:09:40Z E! FATAL: [inputs.procstat] panicked: runtime error: invalid memory address or nil pointer dereference, Stack:
goroutine 147 [running]:
github.com/influxdata/telegraf/agent.panicRecover(0x4d410370)
	/go/src/github.com/influxdata/telegraf/agent/agent.go:1202 +0x74
panic({0x67aa400, 0xc587b20})
	/usr/local/go/src/runtime/panic.go:770 +0xfc
github.com/shirou/gopsutil/v3/process.(*Process).createTimeWithContext(0x4d0a0368, {0x8232a44, 0xc9983c0})
	/go/pkg/mod/github.com/shirou/gopsutil/v3@v3.24.4/process/process_freebsd.go:121 +0x4c
github.com/shirou/gopsutil/v3/process.(*Process).CreateTimeWithContext(0x4d0a0368, {0x8232a44, 0xc9983c0})
	/go/pkg/mod/github.com/shirou/gopsutil/v3@v3.24.4/process/process.go:310 +0x74
github.com/shirou/gopsutil/v3/process.NewProcessWithContext({0x8232a44, 0xc9983c0}, 0x3744)
	/go/pkg/mod/github.com/shirou/gopsutil/v3@v3.24.4/process/process.go:218 +0x78
github.com/shirou/gopsutil/v3/process.NewProcess(...)
	/go/pkg/mod/github.com/shirou/gopsutil/v3@v3.24.4/process/process.go:203
github.com/influxdata/telegraf/plugins/inputs/procstat.newProc(0x3744)
	/go/src/github.com/influxdata/telegraf/plugins/inputs/procstat/process.go:38 +0x30
github.com/influxdata/telegraf/plugins/inputs/procstat.(*Procstat).gatherOld(0x4ccc6e48, {0x824a858, 0x4d40cae0})
	/go/src/github.com/influxdata/telegraf/plugins/inputs/procstat/procstat.go:209 +0x848
github.com/influxdata/telegraf/plugins/inputs/procstat.(*Procstat).Gather(0x4ccc6e48, {0x824a858, 0x4d40cae0})
[...]

PR #1694, opened 2024-08-20
Fix panic on OpenBSD and FreeBSD systems if KinfoProc size has an unexpected size

fivitti:
Hello!

We observed the below panic in our application:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x714230]

goroutine 8 [running]:
github.com/shirou/gopsutil/process.(*Process).NameWithContext(0xc0000d0a80, {0xcb9678, 0x11e8860})
        /usr/local/stork/tools/golang/gopath/pkg/mod/github.com/shirou/gopsutil@v3.21.11+incompatible/process/process_openbsd.go:52 +0x30
github.com/shirou/gopsutil/process.(*Process).Name(...)
        /usr/local/stork/tools/golang/gopath/pkg/mod/github.com/shirou/gopsutil@v3.21.11+incompatible/process/process.go:347
isc.org/stork/agent.(*processWrapper).GetName(0xc0003f2e70?)
```

The application was running on the OpenBSD system.

The line where the panic occurs is `process/process_freebsd.go#L49` and `process/process_openbsd.go#L53`:

```go
name := common.IntToString(k.Comm[:])
```

Access to the' k's member is the only thing that can cause the `nil pointer dereference` error. It panics if `k` is nil.

`k` is instantiated a few lines above:

```go
k, err := p.getKProc()
if err != nil {
	return "", err
}
name := common.IntToString(k.Comm[:])
```

The `p.getKProc()` is system-specific. Its implementation for the FreeBSD (`process/process_freebsd.go#L325`) and OpenBSD (`process/process_openbsd.go#L341`) systems includes a bug.

```go
	buf, length, err := common.CallSyscall(mib)
	if err != nil {
		return nil, err
	}
	if length != sizeOfKinfoProc {
		return nil, err
	}
```

If the `common.CallSyscall(mib)` function returns an empty buffer without any error or the buffer has an unexpected length, the `getKProc` returns two `nil`s (the first `nil` is provided explicitly, and the `err` is `nil`).

Unfortunately, the `NameWithContext` expects that the first value is not `nil` when the `err` is `nil`. The code panics because the above case is not handled.

The problem affects only OpenBSD and FreeBSD systems. Darwin (MacOS)-specific implementation doesn't have this bug.

This pull request solves the problem by returning a non-nil error if the returned buffer has an unexpected length.

Regards!

### DIFF
--- a/process/process_freebsd.go
+++ b/process/process_freebsd.go
@@ -6,6 +6,7 @@ package process
 import (
 	"bytes"
 	"context"
+	"errors"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -329,7 +330,7 @@ func (p *Process) getKProc() (*KinfoProc, error) {
 		return nil, err
 	}
 	if length != sizeOfKinfoProc {
-		return nil, err
+		return nil, errors.New("unexpected size of KinfoProc")
 	}
 
 	k, err := parseKinfoProc(buf)
--- a/process/process_openbsd.go
+++ b/process/process_openbsd.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
@@ -343,7 +344,7 @@ func (p *Process) getKProc() (*KinfoProc, error) {
 		return nil, err
 	}
 	if length != sizeOfKinfoProc {
-		return nil, err
+		return nil, errors.New("unexpected size of KinfoProc")
 	}
 
 	k, err := parseKinfoProc(buf)


shirou:
(This might be related to #1636)

Indeed. I hadn’t noticed that at all. Thank you for pointing it out!

However, instead of creating and returning a new error on the caller side, I think it would be better to have `common.CallSyscall` return the correct error directly. Specifically, I suggest changing https://github.com/shirou/gopsutil/blob/master/internal/common/common_freebsd.go#L65 to `return b, length, errors.New("returned buffer size is null")` (or a more appropriate English message, if you have one). What do you think?

fivitti:
@shirou 

I'm unsure if it is a proper way to solve the problem. The `CallSyscall` function is a general purpose. It might be used to call commands that don't return any data, just a status code. Calling this kind of command would produce an unexpected error if the proposed change is applied. 

Additionally, it doesn't fully cover the problem discussed in this MR. We need to check not only if the returned buffer is not empty but also if it has an expected length. Otherwise, we could parse an improper or incomplete response.

I think it would be best to keep it as is. But if you, as a project maintainer, have an opposite opinion, I'm ready to change the code as requested because you have a more comprehensive view of this library's design than I do.

shirou (approved):
Thank you for your feedback. I understand that given the general-purpose nature of the `CallSyscall` function, my proposed change might cause unexpected errors. I also agree on the importance of checking the returned buffer's length on the caller side.

It’s clear now that my suggestion is insufficient, so I support proceeding with the current implementation. I appreciate you sharing your perspective as a project maintainer. I look forward to continuing our collaboration to achieve better design in the future. Thank you so much!

PR #1706, opened 2024-09-07
[process][darwin][freebsd][linux][openbsd] Make process.Children not reliant on pgrep

### LINKED ISSUE #1698: [process][unix] `err != nil` when calling process.Children() for a process without child processes
**Describe the bug**
Calling `Children()` on a process without child processes returns a non-nil err

the source code uses `common.CallPgrepWithContext` which [calls `pgrep -P $PID`](https://github.com/shirou/gopsutil/blob/74cb4037306917993ef750349472769e7c298e66/internal/common/common_unix.go#L45). If input pid is without child processes, `pgrep` exit code is `1`, the function call returns error with `exits status 1`

**To Reproduce**
```go
// paste example code reproducing the bug you are reporting
```

**Expected behavior**
The function should return ErrorNoChildren?

**Additional context**

Lomanic:
pgrep -P $PID exits with status of 1 (and nothing in stdout nor stderr) both if a process doesn't exist or it doesn't have child processes, so we don't use it anymore on these OSes. We sort PIDs as pgrep did.

Also remove the ErrorNoChildren error when there are no child processes, this is erroneous (simply check for the length of the returned slice, plus this is not an error per se), this was only returned on linux anyway.

Fixes #1698

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -18,7 +18,7 @@ import (
 
 var (
 	invoke                 common.Invoker = common.Invoke{}
-	ErrorNoChildren                       = errors.New("process does not have children")
+	ErrorNoChildren                       = errors.New("process does not have children") // Deprecated: ErrorNoChildren is never returned by process.Children(), check its returned []*Process slice length instead
 	ErrorProcessNotRunning                = errors.New("process does not exist")
 	ErrorNotPermitted                     = errors.New("operation not permitted")
 )
--- a/internal/common/common_unix.go
+++ b/internal/common/common_unix.go
@@ -40,23 +40,3 @@ func CallLsofWithContext(ctx context.Context, invoke Invoker, pid int32, args ..
 	}
 	return ret, nil
 }
-
-func CallPgrepWithContext(ctx context.Context, invoke Invoker, pid int32) ([]int32, error) {
-	out, err := invoke.CommandWithContext(ctx, "pgrep", "-P", strconv.Itoa(int(pid)))
-	if err != nil {
-		return []int32{}, err
-	}
-	lines := strings.Split(string(out), "\n")
-	ret := make([]int32, 0, len(lines))
-	for _, l := range lines {
-		if len(l) == 0 {
-			continue
-		}
-		i, err := strconv.ParseInt(l, 10, 32)
-		if err != nil {
-			continue
-		}
-		ret = append(ret, int32(i))
-	}
-	return ret, nil
-}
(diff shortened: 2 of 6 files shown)


Lomanic:
Tested only on linux and openbsd

process/process.go
```diff
@@ -18,7 +18,6 @@ import (
 
 var (
 	invoke                 common.Invoker = common.Invoke{}
-	ErrorNoChildren                       = errors.New("process does not have children")
```
shirou, line 21: I understand that `ErrorNoChildren` is a practically impossible error definition because when there are no children, `pgrep` exits with status 1. Therefore, it would be fine to remove this definition. However, since this definition has already been published, removing it would cause compilation failures.

So, how about adding a comment like `Deprecated: This error is practically non-occurring`? This way, it will be noted in the [documentation](https://go.dev/wiki/Deprecated) and [flagged by some linters](https://staticcheck.dev/docs/checks/#SA1019).

shirou (commented):
Thank you so much! I have confirmed that the expected behavior occurs on FreeBSD 13.3 and Ubuntu 20.04. This means, in the current `master` branch implementation, where an error `exit status 1` would be returned, `nil` is now returned.

I have added a comment regarding the removal of the error definition. I would appreciate it if you could consider it.

process/process.go
Lomanic: Do you suggest to still keep the variable definition, but to never return it when there are no child processes found? Or to still return it when there are no child processes?

Keep in mind the API contract was already broken as this error was never documented in the Children() function comment and more importantly it was only returned on linux.

shirou: > Do you suggest to still keep the variable definition, but to never return it when there are no child processes found? Or to still return it when there are no child processes?

I am thinking of keeping this variable defined, and an empty list and `nil` would be returned. It means `ErrorNoChildren` is never used.


> Keep in mind the API contract was already broken as this error was never documented in the Children() function comment and more importantly it was only returned on Linux.

You are indeed correct that the error is not mentioned in the `Children()` function comment. However, unfortunately, [pkg.go.dev](https://pkg.go.dev/github.com/shirou/gopsutil/v4@v4.24.8/process#pkg-variables) displays the Linux definition by default, and this variable is included in that documentation. There may be users who are only looking at this variable and use it.

Lomanic: Sure, done.

shirou (approved):
LGTM! Thank you so much for your awesome work!

PR #1707, opened 2024-09-10
[common] add HOST_PROC_MOUNTINFO to EnvMap

atoulme:
Add one more key to EnvMap for the environment variable HOST_PROC_MOUNTINFO.


shirou (approved):
Ah, we forget to add. Thank you so much!

PR #1712, opened 2024-09-19
implement process:NumFDs for Windows

al-online:
I found that gopsutil does not have a method to get the number of handles for a process, so I implemented it myself and would like to submit a pull request.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -549,6 +549,8 @@ func (p *Process) NumCtxSwitchesWithContext(ctx context.Context) (*NumCtxSwitche
 	return nil, common.ErrNotImplementedError
 }
 
+// NumFDsWithContext returns the number of handles for a process on Windows,
+// not the number of file descriptors (FDs).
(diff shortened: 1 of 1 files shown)


shirou:
I have confirmed that it works on my Windows machine, and it seems to be functioning correctly. Thank you.

However, file descriptors (FD) on Unix-like systems and handles on Windows are not exactly the same. Because of this, psutil provides a separate function called [num_handles()](https://psutil.readthedocs.io/en/latest/#psutil.Process.num_handles).

It would be possible for gopsutil to prepare a platform-specific struct like ExWindows, but that would add complexity.

The differences between FDs and handles are not that significant, so it seems like we just need to be mindful of how they are used. Therefore, how about adding a comment like, "On Windows, this returns the number of handles, not number of FDs" on the `NumFDsWithContext` in process_windows.go?

shirou (approved):
Thank you for the update and your great contribution!

PR #1713, opened 2024-09-23
Fix resolution for process create time limited to seconds

joshiste:
Due to the operation order, the create time of a process has a resolution of only a second.
By reordering the operations, we can increase the precision


shirou (approved):
You're absolutely right! Good catch!

PR #1715, opened 2024-09-26
Fix error message typo in sensors_linux

henrygd:
Sorry to open a PR just for a typo. This changes "tempreteure" to "temperature."


shirou (approved):

PR #1720, opened 2024-10-03
feat(mem): Add windows commit stats

### LINKED ISSUE #1719: Expose MemCommit info for Windows
Windows basically doesn't allow to overcommit memory, and applications most times start to fail when there is enough physical memory available, so this is a critical metric for the platform.

Expose `CommitTotal` and `CommitLimit` metrics.
It probably makes sense to also expose `ullTotalPageFile` and `ullAvailPageFile`, but I am less concerned about that.
This data is already being [requested from the system, but not exposed](https://github.com/shirou/gopsutil/blob/master/mem/mem_windows.go#L58).

[A clear and concise description of any alternative solutions or features you've considered.]

**Additional context**
Example of the difference between Used/Commited memory in Windows Task Manager:
image
Stats doc:
https://learn.microsoft.com/en-us/windows/win32/api/psapi/ns-psapi-performance_information#members
https://learn.microsoft.com/en-us/windows/win32/memory/memory-performance-information#system-memory-performance-information

n4nn31355:
I'm kinda puzzled if I should avoid calling `procGetPerformanceInfo` and stick to computing commit stats from `memoryStatusEx.ullTotalPageFile` and `memoryStatusEx.ullAvailPageFile` or expose pagefile stats and compute it in final application.

However, I chose this implementation because it is more aligned with the terminology used by Windows Task Manager and Powershell's `Get-Counter`, both of which refer to "commit" rather than "pagefile."

Closes #1719

### DIFF
--- a/mem/ex_windows.go
+++ b/mem/ex_windows.go
@@ -11,7 +11,10 @@ import (
 
 // ExVirtualMemory represents Windows specific information
 // https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-memorystatusex
+// https://learn.microsoft.com/en-us/windows/win32/api/psapi/ns-psapi-performance_information
 type ExVirtualMemory struct {
+	CommitLimit  uint64 `json:"commitLimit"`
+	CommitTotal  uint64 `json:"commitTotal"`
 	VirtualTotal uint64 `json:"virtualTotal"`
 	VirtualAvail uint64 `json:"virtualAvail"`
 }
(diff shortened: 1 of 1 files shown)


shirou (approved):
I apologize for the delayed response. I've been too busy over the past two months.

I have confirmed that your PR works correctly on my Windows 11. Additionally, I verified that at least `commitLimit` matches the following PowerShell command. We appreciate your contribution!

```
$MemCounters = (Get-Counter -ListSet Memory).Paths
Get-Counter -Counter $MemCounters
```

PR #1722, opened 2024-10-05
fix(mem): possible memory leak on Windows

uubulb:
Call PdhCloseQuery after a PdhOpenQuery call


shirou (approved):

There might be similar issues in other functions on Windows as well, so I plan to investigate later and submit a PR if necessary (or PR is always welcome!)

Issue #1726, opened 2024-10-07
sensors.TemperaturesWithContext() crashes on darwin arm64


sni:
**Describe the bug**
calling sensors.TemperaturesWithContext(ctx) crashes with sigsegv

**To Reproduce**
```go
sensors.TemperaturesWithContext(context.TODO())
```

**Expected behavior**
I would not expect it to crash obviously

**Environment (please complete the following information):**
 - [X] Mac OS: [paste the result of `sw_vers` and `uname -a`

Happens on the github darwin runner which reports as:
```
Current runner version: '2.319.1'
Operating System
  macOS
  14.6.1
  23G93
Runner Image
  Image: macos-14-arm64
  Version: 20240922.1

using go:
go version go1.22.7 darwin/arm64
```

**Additional context**

I am a but curious why it calls runtime.cgocall() in the end if this version should be without cgo?

Here is the relevant part of the crash output:

```
SIGSEGV: segmentation violation
PC=0x195daf738 m=8 sigcode=2 addr=0x0
signal arrived during cgo execution

goroutine 22 gp=0x140002501c0 m=8 mp=0x1400006c408 [syscall]:
runtime.cgocall(0x102c775f0, 0x140004b6ea0)
/Users/runner/hostedtoolcache/go/1.22.7/arm64/src/runtime/cgocall.go:157 +0x44 fp=0x140003101c0 sp=0x14000310180 pc=0x102955944
    github.com/ebitengine/purego.RegisterFunc.func4({0x14000494438?, 0x1?, 0x1?})
/Users/runner/go/pkg/mod/github.com/ebitengine/purego@v0.8.0/func.go:320 +0x938 fp=0x14000310640 sp=0x140003101c0 pc=0x102c74208
    reflect.callReflect(0x14000469d70, 0x14000310c18, 0x14000310a78, 0x14000310a80)
/Users/runner/hostedtoolcache/go/1.22.7/arm64/src/reflect/value.go:782 +0x3e8 fp=0x14000310a20 sp=0x14000310640 pc=0x1029fb7a8
    reflect.callReflect(0x14000469d70, 0x14000310c18, 0x14000310a78, 0x14000310a80)
<autogenerated>:1 +0x28 fp=0x14000310a50 sp=0x14000310a20 pc=0x102a0bcd8
    reflect.makeFuncStub()
[...]

PR #1727, opened 2024-10-07
sensors: avoid passing nil pointer to CFArrayGetCount

uubulb:
Add nil check for `matchingsrvs` before calling `CFArrayGetCount`, as some virtual environments (e.g., Github's action runners) lack sensor data.

Fix https://github.com/shirou/gopsutil/issues/1726

### DIFF
--- a/sensors/sensors_darwin_arm64.go
+++ b/sensors/sensors_darwin_arm64.go
@@ -96,6 +96,10 @@ func (ta *temperatureArm) getProductNames() []string {
 	ta.ioHIDEventSystemClientSetMatching(uintptr(system), uintptr(ta.sensors))
 	matchingsrvs := ta.ioHIDEventSystemClientCopyServices(uintptr(system))
 
+	if matchingsrvs == nil {
+		return nil
+	}
+
 	count := ta.cfArrayGetCount(uintptr(matchingsrvs))
 
 	var i int32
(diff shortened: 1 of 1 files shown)


vicenterzl:
I need this fix too.

sensors/sensors_darwin_arm64.go
```diff
@@ -96,6 +96,10 @@ func (ta *temperatureArm) getProductNames() []string {
 	ta.ioHIDEventSystemClientSetMatching(uintptr(system), uintptr(ta.sensors))
 	matchingsrvs := ta.ioHIDEventSystemClientCopyServices(uintptr(system))
 
+	if matchingsrvs == nil {
+		return nil
```
shirou, line 100: Returning nil may cause another panic on `dumpNameValues()`. How about return empty value?
```suggestion
		return []string{}
```

sensors/sensors_darwin_arm64.go
```diff
@@ -130,6 +134,10 @@ func (ta *temperatureArm) getThermalValues() []float64 {
...
```
shirou, line 138: ditto
```suggestion
		return []float64{}
```

sensors/sensors_darwin_arm64.go
shirou: I know that in Go, an empty slice can be represented as nil. However, due to the subtle differences, I thought that an empty slice was more explicit and therefore preferable. But after reviewing it again, I now see that a nil slice is [considered better](https://go.dev/wiki/CodeReviewComments#declaring-empty-slices), so I retract my previous opinion. Thank you for your comment.

shirou (approved):
Thank you for your contribution and comment!

PR #1743, opened 2024-11-10
enable testifylint linter

mmorel-35:
This enables [testifylint](https://github.com/Antonboom/testifylint) which is a linter that checks for usage of github.com/stretchr/testify.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,6 +35,7 @@ linters:
     - nolintlint
     - predeclared
     - revive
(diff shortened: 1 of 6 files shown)


shirou (approved):
Apologies for the delayed response. At first, I thought it might not be necessary, but the current gopsutil test code is written by various contributors and lacks consistency. Therefore, I reconsidered and believe this linter could be useful. Thank you for your contribution.

PR #1757, opened 2024-12-19
[openbsd] set ConnectionsPid to return NotImplemented

tuftedocelot:
Seems to fix building latest telegraf on OpenBSD


shirou:
There are several functions similar to ConnectionsPid, such as

- Connections
- ConnectionsMax
- ConnectionsWithoutUids
- ConnectionsPid
- ConnectionsPidWithoutUids
- ConnectionsPidMax
- ConnectionsPidMaxWithoutUids

I think these functions also should be added in openbsd.

tuftedocelot:
Something more like this? These were all taken from net/net_unix.go more or less verbatim

PR #1761, opened 2024-12-21
enable all rules of testifylint linter

mmorel-35:
Related to #1743

@shirou , I realised that there is a misspell in the configuration.

And all the excluded rules are already applied so they can be enabled

Only one Equal was left.


ccoVeille (approved):

shirou (approved):
Ah, thank you for your follow up.

PR #1762, opened 2024-12-21
enable all rules of perfsprint linter

mmorel-35:
#### Description

[perfsprint](https://golangci-lint.run/usage/linters/#perfsprint) checks that fmt.Sprintf can be replaced with a faster alternative.


ccoVeille (approved):

shirou (approved):
Thank you!

PR #1763, opened 2024-12-21
since Go 1.18 'interface{}' can be replaced by 'any'

mmorel-35:
#### Description

Use revive rule 'use-any' to ensure that 'any' is used instead of 'interface{}'

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -57,6 +57,9 @@ linters-settings:
             recommandations:
               - io
               - os
+  revive:
+    rules:
+      - name: use-any
   testifylint:
     disable:
       - expected-actual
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -57,6 +57,9 @@ linters-settings:
             recommandations:
               - io
               - os
+  revive:
+    rules:
+      - name: use-any
```
ccoVeille, line 62: Unfortunately, this is not the way to do with revive.

Or at least, it's not doing exactly what you want

Take a look here

https://github.com/ccoVeille/golangci-lint-config-examples/blob/4480ec8754d46e9993333847e6597141e00d408e/03-safe%2F.golangci.yml#L68-L139

ccoVeille (changes requested):
I approved by error

PR #1767, opened 2024-12-30
fix: memory leak and string corruption in ARM Mac temperature sensors

chuangbo:
This PR addresses several issues in the temperature sensor implementation for ARM Mac:
1. Memory leak caused by unreleased system resources
2. Increasing port counts in Activity Monitor
3. String corruption where the first character becomes "\x00"

## Changes

1. Resource Management:
   - Added proper resource cleanup using `defer` statements
   - Refactored code to share the HID system client instead of creating new ones
   - Fixed memory leaks by ensuring all CoreFoundation objects are properly released

2. String Handling:
   - Fixed buffer allocation for string conversion
   - Properly handle null terminators in C string to Go string conversion
   - Added correct string length calculations

3. Code Structure:
   - Reduced resource allocation by sharing system client between functions
   - Enhanced code readability with better comments
   
## Testing
Tested on Apple Silicon Mac with the following improvements:
- Memory usage remains stable over time
- Port count in Activity Monitor stays constant
- No more string corruption issues

## How to Reproduce
You can reproduce these issues using the following code:

```go
package main

import (
	"context"
	"time"

	"log/slog"

	"github.com/shirou/gopsutil/v4/sensors"
)

func main() {
	slog.SetLogLoggerLevel(slog.LevelDebug)
	ticker := time.NewTicker(10 * time.Millisecond)

	for range ticker.C {
		temps, err := sensors.TemperaturesWithContext(context.Background())
		if err != nil {
			slog.Debug("Sensor error", "err", err)
		}
		slog.Debug("Temperature", "sensors", temps)
	}
}
```

go run demo.go | grep "x00"

Memory leaks screen recordings: https://github.com/user-attachments/assets/0168feba-5637-4f3f-8a45-a56a3b5cbfdb

String corruption screenshot:
String corruption

@uubulb would you mind taking a look? Many thanks!

### DIFF
--- a/sensors/sensors_darwin_arm64.go
+++ b/sensors/sensors_darwin_arm64.go
@@ -41,11 +41,16 @@ func TemperaturesWithContext(ctx context.Context) ([]TemperatureStat, error) {
 	}
 
 	ta.matching(0xff00, 5)
-	thermalNames := ta.getProductNames()
-	thermalValues := ta.getThermalValues()
+	defer ta.cfRelease(uintptr(ta.sensors))
+
+	// Create HID system client
+	system := ta.ioHIDEventSystemClientCreate(common.KCFAllocatorDefault)
+	defer ta.cfRelease(uintptr(system))
+
+	thermalNames := ta.getProductNames(system)
+	thermalValues := ta.getThermalValues(system)
 	result := dumpNameValues(thermalNames, thermalValues)
 
-	ta.cfRelease(uintptr(ta.sensors))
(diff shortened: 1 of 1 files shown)


shirou (approved):
Thank you very much. The reproduction code and video were very clear. Unfortunately, I don't own a mac, so I can't verify the behavior myself, but it seems that @uubulb has also confirmed it, and the code itself looks fine. Thank you for your contribution!

PR #1725, opened 2024-10-07
add unit test for sensors.TemperaturesWithContext

sni:
sensors.TemperaturesWithContext() crashes with a segfault. This PR adds a test case which simply calls sensors.TemperaturesWithContext() to make sure it does not crash.

It is probably not so easy to have more useful test case because sensors are different everywhere and even errors might be ok.

```
SIGSEGV: segmentation violation
PC=0x195daf738 m=8 sigcode=2 addr=0x0
signal arrived during cgo execution

goroutine 22 gp=0x140002501c0 m=8 mp=0x1400006c408 [syscall]:
runtime.cgocall(0x102c775f0, 0x140004b6ea0)
/Users/runner/hostedtoolcache/go/1.22.7/arm64/src/runtime/cgocall.go:157 +0x44 fp=0x140003101c0 sp=0x14000310180 pc=0x102955944
    github.com/ebitengine/purego.RegisterFunc.func4({0x14000494438?, 0x1?, 0x1?})
/Users/runner/go/pkg/mod/github.com/ebitengine/purego@v0.8.0/func.go:320 +0x938 fp=0x14000310640 sp=0x140003101c0 pc=0x102c74208
    reflect.callReflect(0x14000469d70, 0x14000310c18, 0x14000310a78, 0x14000310a80)
/Users/runner/hostedtoolcache/go/1.22.7/arm64/src/reflect/value.go:782 +0x3e8 fp=0x14000310a20 sp=0x14000310640 pc=0x1029fb7a8
    reflect.callReflect(0x14000469d70, 0x14000310c18, 0x14000310a78, 0x14000310a80)
<autogenerated>:1 +0x28 fp=0x14000310a50 sp=0x14000310a20 pc=0x102a0bcd8
    reflect.makeFuncStub()
/Users/runner/hostedtoolcache/go/1.22.7/arm64/src/reflect/asm_arm64.s:48 +0x58 fp=0x14000310c10 sp=0x14000310a50 pc=0x102a07ce8
    github.com/shirou/gopsutil/v4/sensors.(*temperatureArm).getProductNames(0x14000311058)
/Users/runner/go/pkg/mod/github.com/shirou/gopsutil/v4@v4.24.9/sensors/sensors_darwin_arm64.go:99 +0x184 fp=0x14000310d20 sp=0x14000310c10 pc=0x102d0f624
    github.com/shirou/gopsutil/v4/sensors.TemperaturesWithContext({0x1038c5be0?, 0x103a985e0?})
[...]

PR #1768, opened 2024-12-30
refactor(darwin): optimize string allocation

uubulb:
Introduce a new CStr type to simplify string conversion, making it less error-prone.
Also added some comments to improve readability and unit test for sensors package.

Supersedes https://github.com/shirou/gopsutil/pull/1725


shirou (approved):
Amazing. `CStr` makes the existing code much more readable. Thank you!

PR #1769, opened 2024-12-31
feat(process_windows): Support collecting PageFault count in Windows

zmyzheng:
This PR adds the statistics collection for PageFault count in Windows. Different from Linux which specifies Major Page Fault and Minor Page Fault, Windows only provides the total PageFault count.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# ensure that line endings for Windows builds are properly formatted
+# see https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#how-to-use
+# at "Multiple OS Example" section
+*.go text eol=lf
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -1,5 +1,15 @@
-on: [push, pull_request]
 name: Build Test
+
+on:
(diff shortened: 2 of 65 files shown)


shirou:
Sorry for keeping you waiting. After conducting a [survey](https://github.com/shirou/gopsutil/issues/1770), it appears that all operating systems except Windows distinguish between Major and Minor faults. I believe adding "Total" solely for Windows might lead to inconsistency.  

Therefore, how about adding `mem.PageFaultCount` to "Major" and including a comment like, "Since Windows does not distinguish between Major and Minor faults, all faults are treated as Major"?

zmyzheng:
> Sorry for keeping you waiting. After conducting a [survey](https://github.com/shirou/gopsutil/issues/1770), it appears that all operating systems except Windows distinguish between Major and Minor faults. I believe adding "Total" solely for Windows might lead to inconsistency.
> 
> Therefore, how about adding `mem.PageFaultCount` to "Major" and including a comment like, "Since Windows does not distinguish between Major and Minor faults, all faults are treated as Major"?

Sounds good. Thanks @shirou for the investigation. I updated the code based on your suggestion.

PR #1778, opened 2025-01-18
chore: explicit revive rules

mmorel-35:
* List explicitly the activated rules of revive.
* fixes redundant-import-alias as there is only one
* moves G115 exclusion to a dedicated configuration of gosec


ccoVeille (approved):

shirou (approved):
I have not checked the revive rules one by one, but it LGTM and seems good starting point. Thank you for your contribution!

PR #1779, opened 2025-01-18
chore: call actions/checkout before actions/setup-go to benefit from native cache

mmorel-35:
actions/setup-go is able to cache with the same configuration as what is done with an extra configuration.
This simplifies the worflow to benefit from this automatic configuration


shirou (approved):
Good catch! I haven't noticed the cache is enabled by default. and also go version hard-coded in lint. Thank you so much!

PR #1784, opened 2025-01-30
chore: enable use-any from revive

mmorel-35:
#### Description 

[use-any](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#use-any): Since Go 1.18, interface{} has an alias: any. This rule proposes to replace instances of interface{} with any.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -96,7 +96,6 @@ linters-settings:
       - name: unused-parameter
         disabled: true
       - name: use-any
-        disabled: true
       - name: var-declaration
       - name: var-naming
         disabled: true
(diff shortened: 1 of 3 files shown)


shirou:
I don't know much about the Revive, but is it possibly only supported on Linux?. It seems [net/net_windows.go](https://github.com/shirou/gopsutil/blob/master/net/net_windows.go#L329) and internal/common/common_windows.go have also `interface{}`.

ccoVeille:
Somehow, I took a look, and here is the solution.

https://github.com/ccoVeille/todo/issues/260#issuecomment-2624596652

So adding a golangci-lint job per arch via matrix seems the solution.

It's strange I had never faced this. So good catch @shirou TIL something and I'm happy about it

shirou (approved):
great. Thank you so much!

PR #1785, opened 2025-01-30
chore: enable golangci-lint on Windows and MacOS

mmorel-35:
#### Description

Following https://github.com/shirou/gopsutil/pull/1784#issuecomment-2624558245

I have deliberately added '//nolint:... //FIXME' to make them easily to spot and fix

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.go text eol=lf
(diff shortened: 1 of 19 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.go text eol=lf
```
ccoVeille, line 1: This one might be out of scope, at least the scope of the PR no?

If not please add a comment in the file about its purpose

ccoVeille (approved):
So much changes and fixes from a single comment that let us discover golangci-lint was not analysising code out of its compilation architecture

.gitattributes
ccoVeille: Indeed, it makes sense.

please comsider my point about adding an inline comment to explain it

PR #1786, opened 2025-02-02
remove wmi to get number of physical cores

OrbintSoft:
This is first attempt to remove the slow and unreliable wmi from gopsutil.

cpuid get information directly from the cpu where the code is running, so it's very easy and fast to get the number of physical cores.

The only issue could be in a system with more than one processor, I don't have one to test it, I assume that if you have more than one processor they are all of same model.
I am not aware of a frankestein system with more cpu of different models, Intel recommends to use cpu of same model and and in any case says to use only 2 cpu with same number of cores:
https://www.intel.com/content/www/us/en/support/articles/000055408/server-products/server-boards.html

So the code should work in all situations, and probably even more reliable than wmi.

Anyway if someone has a multi cpu system to test it would be nice.

### DIFF
--- a/cpu/cpu_test.go
+++ b/cpu/cpu_test.go
@@ -79,24 +79,27 @@ func TestTimes(t *testing.T) {
 }
 
 func TestCounts(t *testing.T) {
-	v, err := Counts(true)
+	logicalCount, err := Counts(true)
 	skipIfNotImplementedErr(t, err)
 	if err != nil {
 		t.Errorf("error %v", err)
 	}
-	if v == 0 {
-		t.Errorf("could not get logical CPU counts: %v", v)
+	if logicalCount == 0 {
+		t.Errorf("could not get logical CPU counts: %v", logicalCount)
 	}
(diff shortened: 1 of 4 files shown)


Lomanic:
Thanks for your time spent on this @OrbintSoft.

Unfortunately I would reject this (or at least put it on hold), it still doesn't implement the same thing as python's psutil ([see the link in comment you deleted here](https://github.com/shirou/gopsutil/pull/1786/files#diff-1ea4829c2e1b7e46628e250c5c03d44d080997e5dc03f763f53a28094f6c2b81L215)).

I may be able to finally implement physical core counting calling GetLogicalProcessorInformationEx in the near future.

OrbintSoft:
> Having a different number of CPU cores is common in ARM SoCs. For example, the RK3399 has 2x Cortex-A72 and 4x Cortex-A53, so this approach might not be reliable in such situations.

this code is windows only where arm support is limited. anyway I agree, I will try to implement the python approach that is more generic.

I close this PR for now.

PR #1797, opened 2025-02-20
chore(golangci-lint): GOOS and GOARCH matrix

mmorel-35:
#### Description

Provide golangci-lint with GOOS and GOARCH variations. The list is aligned with `make vet` list.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


shirou (approved):
This is a wonderful improvement! Also, thank you very much for all the fixes!

PR #1799, opened 2025-02-22
chore: reduce number of executed jobs

mmorel-35:
#### Description

Reduce the number of executed jobs by two ways

* jobs are triggers on push on master only or on pull requests
* with the concurrency configuration the jobs of a same PR get canceled if there is a more recent commit happening

It also sort fields and indent arrays in workflows

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -1,5 +1,15 @@
-on: [push, pull_request]
 name: Build Test
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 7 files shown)


shirou (approved):
Thank you for your contribution! Since you were still pushing changes, I wasn’t sure when to start the review. It would be helpful if you could set the PR to draft status until it’s ready for review or use the review request feature.  

That aside, adding concurrency is definitely a good idea. I also appreciate the other small fixes you made.

PR #1804, opened 2025-03-03
chore: update go version in module

mmorel-35:
#### Description

[Go 1.24.0](https://go.dev/blog/go1.24) has been released on since 11 February 2025

This updates Go version from 1.18 to 1.23

Notice that loong64 arch is only available since Go1.19 so it's surprising to see that gopsutil supports it (at least partially) with a Go1.18 module.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -122,3 +122,5 @@ linters-settings:
   usetesting:
       os-create-temp: false
       os-mkdir-temp: false
+run:
+  timeout: 5m
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/shirou/gopsutil/v4
 
-go 1.18
+go 1.23
 
 require (
 	github.com/ebitengine/purego v0.8.2


shirou:
Thank you for the suggestion!  

However, gopsutil is a library that supports a wide range of platforms. Not all platforms provide the latest version of Go in their package managers, and some users might not be able to upgrade their Go version.  

gopsutil depends on `golang.org/x/sys`, and x/sys's golang dependency is regularly updated. So, in practice, the version from `golang.org/x/sys` might be used, but I still believe that the library itself should aim to support the lowest possible Go version.  

Of course, if there’s a need to use new language features in Go, I’m not against it. However, I think we should discuss whether it’s truly necessary at that time.

shirou:
I just noticed https://github.com/golang/go/issues/69095. This proposal suggests automatically updating the `go.mod` version to `1.(N-1).0`, including `x/sys`, and already accepted.

In the discussion, there are comments suggesting that the required version should be set to the minimum, but it seems that people like rsc believe it's important to encourage the use of the latest version.

Personally, I also agree with the idea of keeping the version at the minimum. However, the overall direction of Go team seems to be to promote the use of the latest version. Therefore, I've changed my mind and now think it might be better to change `1.23.0` in `go.mod`.

shirou:
Thank you for the update. However, when I looked at the diff, I wondered, "Why is the `toolchain` directive included?" So I did some research and found https://github.com/golang/go/issues/65847. I haven't read through the entire issue comments, but it seems that the `toolchain` directive might be automatically added and could potentially cause issues.  

It doesn't seems to be resolved yet, so how about removing the `toolchain` directive in this PR for now?

shirou (approved):

PR #1810, opened 2025-03-06
fix: deprecated set-output

mmorel-35:
#### Description

The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


shirou (approved):
Indeed. Thank you so much!

PR #1815, opened 2025-03-09
chore: use testify instead of testing

mmorel-35:
#### Description

Uses testify instead of testing package so it is more concise and more readable.

### DIFF
--- a/cpu/cpu_darwin_test.go
+++ b/cpu/cpu_darwin_test.go
@@ -7,6 +7,9 @@ import (
 	"os"
 	"runtime"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestInfo_AppleSilicon(t *testing.T) {
@@ -15,19 +18,13 @@ func TestInfo_AppleSilicon(t *testing.T) {
 	}
 
 	v, err := Info()
-	if err != nil {
-		t.Errorf("cpu info should be implemented on darwin systems")
-	}
+	require.NoErrorf(t, err, "cpu info should be implemented on darwin systems")
 
 	for _, vv := range v {
-		if vv.ModelName == "" {
-			t.Errorf("could not get CPU info: %v", vv)
-		}
+		assert.NotEmptyf(t, vv.ModelName, "could not get CPU info: %v", vv)
 		if vv.Mhz <= 0 && os.Getenv("CI") != "true" {
 			t.Errorf("could not get frequency of: %s", vv.ModelName)
 		}
-		if vv.Mhz > 6000 {
-			t.Errorf("cpu frequency is absurdly high value: %f MHz", vv.Mhz)
-		}
+		assert.LessOrEqualf(t, vv.Mhz, float64(6000), "cpu frequency is absurdly high value: %f MHz", vv.Mhz)
 	}
 }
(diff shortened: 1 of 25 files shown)


ccoVeille:
Why testify and not [gotest.tool](https://github.com/gotestyourself/gotest.tools) ?

https://pkg.go.dev/gotest.tools/v3/assert

testify is very old school. We both knows it by trying to maintain [testifylint](https://github.com/Antonboom/testifylint).

gopsutils now uses Go 1.23 via recent go.mod changes, so I would like to suggest using gotest.tool

ccoVeille:
Hi @shirou 

> I know that `testify` has been around for a long time. However, it is still being updated, and most importantly, it works. it is functioning right now.

testify is old, known and widely used.

Unfortunately, it has many issues. A v2 was planned, but it is now a dead end.

Please read

- https://github.com/stretchr/testify/discussions/1560#discussioncomment-8657735

[testifylint](https://github.com/Antonboom/testifylint) is a way to avoid
 
> By the way, why do you think `gotest.tool` is the better choice? What advantages does it have?

We had a recent discussion with @mmorel-35 about the possible alternative solutions.

We listed the following ones:

- https://github.com/alecthomas/assert

- https://github.com/gookit/goutil

- https://github.com/gotestyourself/gotest.tools

- https://github.com/shoenig/test

- https://github.com/onsi/ginkgo


We discussed to the fact gotest.tool is:
- pretty clean
- documented
- maintained
- stable
- close to testify syntax
- provided with a CLI to migrate tests from testify to gotest.tool
- the dependencies is very limited
https://github.com/gotestyourself/gotest.tools/blob/main/go.mod, while testify comes with a [lot of dependency](https://github.com/stretchr/testify/blob/master/go.mod), including [go-spew](https://github.com/davecgh/go-spew) that is no longer maintained.

shirou:
Thank you for the detailed explanation. I see, that makes sense. I also felt that `testify` lacks consistency in some of its functions.  

I appreciate the introduction to other testing libraries as well. Having fewer dependencies is a great advantage, and the existence of a migration tool is very appealing.  

For other tools I’m working on, I’ll consider using these libraries. As for `gopsutil`, I’ll first standardize on `testify` on this PR and then consider migrating to another library from there. Thank you!

shirou (approved):
LGTM! Thank you for the large number of changes!

thaJeztah:
Thanks! In general, I don't think testify is necessarily bad, and I think maintenance was picked up again. Testify is used a bit more widely,  which can sometimes be a benefit for contributors.

gotest.tools started as a bit of a side project to be used in the docker project; some of the existing assertion frameworks were providing too many features and some "opninionated" assertions, that looked nice, but were abstracting too much away, which in a couple of occasions resulted in tests not actually asserting what you'd think they'd do.

gotest.tools was written at the time to be much more minimal; provide the basics, with options to extent where needed. Not encouraging to hide important checks away behind the assertion library.

I should add that I'm NOT a maintainer for gotest.tools, but @dnephin is a former colleague on my team, and as a user of it, have contributed various patches to help maintain it. (Generally, code frequency in the repository should be low; mostly as a result of the project, by design, being minimal)

ccoVeille:
Thanks @thaJeztah for your detailed reply.

I looked at the package API in detail, and I understand what you meant.

The methods are pretty basic for gotest.tool. And it's OK to be simple. I understand the design choice. It provides a good solution for a lot of simple use cases.

From my understanding, depending on my needs and on the code base I might choose it.

But, if the tests need to be more complex, and the error messages more meaningful, I think I will use @shoenig test package.

I understand the choice of @shirou to merge like this. testify was already used in the repository, well known, contributors can easily add tests by using a well known library.

PR #1816, opened 2025-03-09
chore: enable begin rule from thelper

mmorel-35:
#### Description

Enables begin rule from thelper, it also turns SkipIfNotImplementedErr into a common function


shirou (approved):
Thank you. Actually, `skipIfNotImplementedErr` has been a long-time concern of mine.

PR #1817, opened 2025-03-09
chore: enable var-naming from revive

mmorel-35:
#### Description

Enable [var-naming](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming) from revive 

Only fixes private variable or functions naming and add `nolint:revive //FIXME` to the rest of cases.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -99,7 +99,10 @@ linters-settings:
       - name: use-any
       - name: var-declaration
       - name: var-naming
-        disabled: true
+        arguments:
+          - ["ID"] # AllowList
+          - ["VM"] # DenyList
+          - - upperCaseConst: true # Extra parameter (upperCaseConst|skipPackageNameChecks)
   testifylint:
     enable-all: true
   thelper:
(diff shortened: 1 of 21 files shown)


shirou (approved):
Sorry for the long delay, but I have reviewed all of this PR and think it LGTM. However, I would like to exclude these lint rules for auto-generated files.  

That said, at the moment, we are making manual changes after generating the files, as seen here:  
[https://github.com/shirou/gopsutil/pull/1817/files#diff-faf6b4886c9d98b986c892b2e29cda6c98bde8752b2df7552af7079436d08ecbR19-R22](https://github.com/shirou/gopsutil/pull/1817/files#diff-faf6b4886c9d98b986c892b2e29cda6c98bde8752b2df7552af7079436d08ecbR19-R22),  
so for now, I don't think it's an issue, because it's not fully auto-generated. Then I will merge your great PR.

However, in the future, if I decide to disable this linting for certain files, I hope you’ll understand.

PR #1825, opened 2025-03-23
chore: enable govet linter

mmorel-35:
#### Description

- Adding `govet` to the list of enabled linters in the Golang CI configuration

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,7 @@ linters:
     - gocritic
     - gomodguard
     - gosec
+    - govet
     - importas
     - misspell
     - nakedret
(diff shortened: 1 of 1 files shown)


ccoVeille:
I don't think there is a need for [fieldalignment](https://golangci-lint.run/usage/linters/#govet) rule in govet linter.

I'm assuming it's the root cause of so many changes.

gopsutil is used on several OSes where alignment varies.

So if you think there is a need for it, you could take a look at https://pkg.go.dev/structs#HostLayout

(One of my recent discovery, thanks again @cristaloleg for this.)

I would like to recommend you to either:
- disable this rule, and revert the changes
- revert changes, use `structs` package, if needed disable the fieldalignment rule (I'm unsure govet detects the use of structs packages)

shirou (approved):
Thank you so much!

PR #1826, opened 2025-03-23
fix: address nolint contextcheck

mmorel-35:
#### Description

This pull request includes changes to the `mem/mem_solaris.go` file to ensure that context is properly passed to various functions. The most important changes include modifying function signatures to accept a `context.Context` parameter and updating function calls accordingly.


shirou (approved):
LGTM 👍

PR #1829, opened 2025-03-29
[chore]: bump golangci-lint to v2.1.1

mmorel-35:
#### Description

This updates golangci-lint version to v2 and golangci-lint-action to V7.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,20 +1,25 @@
+version: "2"
+formatters:
+  enable:
+    - gci
+    - gofumpt
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
+        - prefix(github.com/shirou)
 issues:
   max-same-issues: 0
-  exclude-rules: []
 linters:
   enable:
     - asciicheck
     - contextcheck
     - durationcheck
     - errorlint
-    - gci
     - gocritic
-    - gofmt
-    - gofumpt
-    - goimports
     - gomodguard
     - gosec
-    - gosimple
     - importas
     - misspell
     - nakedret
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -1,20 +1,34 @@
...
+    - gofumpt
+    - goimports
+  exclusions:
+    generated: lax
+  settings:
+    gci:
+      sections:
+        - standard
+        - default
+        - prefix(github.com/shirou)
```
ccoVeille, line 14: There is a problem for me here

gci supersedes goimports
gofumpt supersedes gofumpt

So here, your conf should be to something as simple as this

```suggestion
formatters:
  enable:
    - gci
    - gofumpt
  exclusions:
    generated: lax
  settings:
    gci:
      sections:
        - standard
        - default
        - prefix(github.com/shirou)
```

The only reasons I would like to see gofmt with gofumpt is if you use extra features such as [simplify](https://pkg.go.dev/cmd/gofmt#hdr-The_simplify_command) and [rewrites-rules](https://pkg.go.dev/cmd/gofmt)

```suggestion
formatters:
  enable:
    - gci
    - gofmt
    - gofumpt
  exclusions:
    generated: lax
  settings:
    gofmt:
      simplify: true
      rewrite-rules:
        - pattern: 'interface{}'
          replacement: 'any'
    gci:
      sections:
        - standard
        - default
        - prefix(github.com/shirou)
```

shirou (approved):
I haven’t been following all the changes in the configuration file syntax of the golangci-lint, but based on your previous contributions, I’m confident this PR will work just as well. Thank you very much!

Issue #1236, opened 2022-01-24
cpu information not available on windows7 32bit


sunyunzhang:
error:  call to NtquerySystemInformation returned 3221225476. err: The operation completed successfully

Lomanic:
Error is raised there [ https://github.com/shirou/gopsutil/blob/e0ec1b9cda4470db704a862282a396986d7e930c/cpu/cpu_windows.go#L175](https://github.com/shirou/gopsutil/blob/e0ec1b9cda4470db704a862282a396986d7e930c/cpu/cpu_windows.go#L175), with  `cpu.Times(true)`

3221225476 is 0xC0000004 which is [`STATUS_INFO_LENGTH_MISMATCH`](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55), _The specified information record length does not match the length that is required for the specified information class._.

I will try to reproduce on Tiny7 which is a stripped down 32bit win7, but I'm surprised.

Howie59:
i reproduce it on win32, does anyone know how to handle?

Howie59:
@Lomanic

Lomanic:
I could indeed reproduce on Tiny7.

I tried to port https://github.com/spiffe/spire/blob/c2057a4906c5b4bc9796e236e905002183021e50/pkg/agent/plugin/workloadattestor/docker/process/winapi.go#L117 as it (re-)allocates buffer size dynamically [instead of setting a really huge one and hope for the best](https://github.com/shirou/gopsutil/blob/c4a0f4689f12914053ab0a62ab6f37dfb36a6807/cpu/cpu_windows.go#L152-L155), [but this part from that code](https://github.com/spiffe/spire/blob/c2057a4906c5b4bc9796e236e905002183021e50/pkg/agent/plugin/workloadattestor/docker/process/winapi.go#L145-L154) I don't get for the moment.

PR #1831, opened 2025-04-01
Fix win32_SystemProcessorPerformanceInformation struct

niemp100:
Fixes An issue on Windows 32Bit Systems where the CPU information is not available and throwing an windows error response code from the API discussed here: #1236


shirou (approved):
I haven't check on 32-bit Windows machine, but LGTM! Thank you for your contribution!

Issue #1834, opened 2025-04-07
Flaky panic on macOS: unsafe.Slice: ptr is nil and len is not zero in perCPUTimes


dvovk:
Hi 👋

We’re encountering a flaky runtime panic on macOS systems when calling cpu.Times() from github.com/shirou/gopsutil/v4@v4.25.2. The panic looks like this:
```
panic: runtime error: unsafe.Slice: ptr is nil and len is not zero

goroutine 1 [running]:
github.com/shirou/gopsutil/v4/cpu.perCPUTimes(0x14000a37660)
	github.com/shirou/gopsutil/v4@v4.25.2/cpu/cpu_darwin.go:158 +0x4ec
github.com/shirou/gopsutil/v4/cpu.TimesWithContext({0x104548ae0?, 0x1400002d110?}, 0x1)
	github.com/shirou/gopsutil/v4@v4.25.2/cpu/cpu_darwin.go:70 +0x70
github.com/shirou/gopsutil/v4/cpu.Times(...)
	github.com/shirou/gopsutil/v4@v4.25.2/cpu/cpu_darwin.go:59
github.com/shirou/gopsutil/v4/cpu.init.0()
```

This seems related to changes introduced in [v4.24.9](https://github.com/shirou/gopsutil/releases/tag/v4.24.9), which updated the macOS CPU code to use unsafe.Slice.

Reproducibility

This happens intermittently — we’ve seen it in CI or test environments on macOS. Runner details https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

dvovk:
This one may solve the issue https://github.com/shirou/gopsutil/pull/1835

shirou:
Since #1835 has been merged, I close this issue. If the panic still happen, feel free to re-open this issue. Thank you for your contribution!

PR #1835, opened 2025-04-07
fix nil ptr

dvovk:
This pull request includes a small but important change to the `perCPUTimes` function in the `cpu/cpu_darwin.go` file. The change adds a check to ensure that the `cpuload` variable is not nil before proceeding, which helps prevent potential runtime errors.

* [`cpu/cpu_darwin.go`](diffhunk://#diff-331b840c2c5c144a66ee5b2729fc9bbfab8987d13bae2d3f8ecf1706199e985aR155-R158): Added a nil check for the `cpuload` variable in the `perCPUTimes` function to handle cases where `host_processor_info` might return a nil value.

Possible fix for: https://github.com/shirou/gopsutil/issues/1834

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -5,6 +5,7 @@ package cpu
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


shirou (approved):
Good catch. Thank you so much!

PR #1837, opened 2025-04-11
Use the context provide to UsageWithContext public function

GustavoCaso:
Currently, at work, we use gopsutil. We wanted to use the `disk.UsageWithContext` to allow setting some timeout when calling the Usage function. 

We realized that the function UsageWithContext does not use the ctx provided. 

It would be a good addition to this great package. 

Before adding a test or even applying this pattern to other functions `*WithContext,` I wanted to know if this is something that we would be ok with having in the package. If so, I can add tests and extend the pattern to the rest of the  `*WithContext` functions. 

Looking forward for your feedback.

### DIFF
--- a/disk/disk.go
+++ b/disk/disk.go
@@ -71,6 +71,31 @@ func Usage(path string) (*UsageStat, error) {
 	return UsageWithContext(context.Background(), path)
 }
 
+func UsageWithContext(ctx context.Context, path string) (*UsageStat, error) {
+	type result struct {
+		usage *UsageStat
+		err   error
+	}
+
+	resultCh := make(chan result, 1)
+
+	go func() {
+		usage, err := getUsage(ctx, path)
+		resultCh <- result{usage: usage, err: err}
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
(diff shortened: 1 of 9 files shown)


shirou:
Thank you for your suggestion.

gopsutil aims to support a wide range of platforms. Some of these platforms use `context` internally in their implementation, while others do not. For example, `disk/disk_aix_nocgo.go` makes use of `ctx`. Because of these differences, we can't write a universal implementation that cancels the context in all cases. 

If we were to implement such logic, I believe the best approach would be to define a separate function that handles context cancellation, and then call that function within each platform-specific `UsageWithContext` function only when necessary. This way, platforms that don't need it can simply omit the call.

Lomanic:
To be honest, the `WithContext()` functions were initially introduced because of wmi disastrous performance on windows https://github.com/shirou/gopsutil/issues/469#issuecomment-352672929, I personally wouldn't mind them to be simply removed instead (hat would be a breaking change so for a major version bump), they've been a bit of a maintenance annoyance more than anything, duplicating all the API only for a few functions calling wmi on windows, while the number of these slow wmi calls has been greatly reduced over the years.

PR #1838, opened 2025-04-12
[disk][linux] add bcachefs magic

NewbieOrange:
Adds bcachefs magic to disk/linux, allow querying for bcachefs partitions


shirou (approved):
Thank you! And I found some of other magics such as `CIFS_SUPER_MAGIC` is still missing. So I will add these in another PR.

PR #1839, opened 2025-04-13
Remove wmi for Get physical CPU core count

OrbintSoft:
Reimplemented get physical core count with windows api, so we can get rid of wmi.

### DIFF
--- a/cpu/cpu_test.go
+++ b/cpu/cpu_test.go
@@ -61,25 +61,15 @@ func TestTimes(t *testing.T) {
 func TestCounts(t *testing.T) {
 	logicalCount, err := Counts(true)
 	common.SkipIfNotImplementedErr(t, err)
-	if err != nil {
-		t.Errorf("error %v", err)
-	}
-	if logicalCount == 0 {
-		t.Errorf("could not get logical CPU counts: %v", logicalCount)
-	}
+	require.NoError(t, err)
+	assert.NotZerof(t, logicalCount, "could not get logical CPU counts: %v", logicalCount)
 	t.Logf("logical cores: %d", logicalCount)
 	physicalCount, err := Counts(false)
 	common.SkipIfNotImplementedErr(t, err)
-	if err != nil {
-		t.Errorf("error %v", err)
-	}
-	if physicalCount == 0 {
-		t.Errorf("could not get physical CPU counts: %v", physicalCount)
-	}
+	require.NoError(t, err)
+	assert.NotZerof(t, physicalCount, "could not get physical CPU counts: %v", physicalCount)
 	t.Logf("physical cores: %d", physicalCount)
-	if physicalCount > logicalCount {
-		t.Errorf("physical cpu cannot be more than logical cpu: %v > %v", physicalCount, logicalCount)
-	}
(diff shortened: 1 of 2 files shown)


cpu/cpu_windows.go
```diff
@@ -200,30 +204,83 @@ type systemInfo struct {
...
+	if err != nil && !errors.Is(err, windows.NTE_OP_OK) {
+		return 0, fmt.Errorf("failed to get logical processor information: %w", err)
+	}
+
+	// Iterate through the buffer to count physical cores
+	offset := uintptr(0)
+	ncpus := 0
+	for offset < uintptr(length) {
+		select {
+		case <-ctx.Done():
```
Lomanic, line 254: Did you extract [this part in the `common.WMIQueryWithContext()` function](https://github.com/shirou/gopsutil/blob/ef5056a446730347150887a7e485da69bd97ae3e/internal/common/common_windows.go#L213-L214)? With the function being way faster without wmi, is this still really needed?

Lomanic (changes requested):
Awesome work, thanks a lot @OrbintSoft. Tested on my windows 10 laptop, compared to [`psutil.cpu_counts(logical=False)`](https://psutil.readthedocs.io/en/latest/#psutil.cpu_count) and everything is fine.

Besides my other remarks, would you also mind squashing your commits?

cpu/cpu_test.go
```diff
@@ -59,16 +59,27 @@ func TestTimes(t *testing.T) {
...
 
 func TestCounts(t *testing.T) {
-	v, err := Counts(true)
+	logicalCount, err := Counts(true)
 	common.SkipIfNotImplementedErr(t, err)
-	require.NoError(t, err)
-	assert.NotZerof(t, v, "could not get logical CPU counts: %v", v)
-	t.Logf("logical cores: %d", v)
-	v, err = Counts(false)
+	if err != nil {
```
shirou, line 64: Would you mind reverting it back to using the `testify/assert` or `testify/require` package? We've standardized on using `testify` in accordance with #1815, so it would be great to keep things consistent.

cpu/cpu_windows.go
shirou: Implementing cancellation individually within each platform-specific function, rather than in a shared/common function, would be difficult and inconsistent. So while I appreciate your effort, please remove this cancellation logic.

This also relates to [this comment](https://github.com/shirou/gopsutil/pull/1837#issuecomment-2816997435), but it would be ideal if we could implement some kind of common function for handling cancellation.

shirou: @Lomanic Could you check the update ?

Lomanic:
LGTM besides my last nit in https://github.com/shirou/gopsutil/pull/1839/files#r2047042897, but it's kind of personal preference, so can be merged at will. Thanks again @OrbintSoft 🙏

OrbintSoft:
Can we merge or do you need other changes?

shirou (approved):
Thank you for updating the PR, and sorry for the delay—I'll go ahead and merge it now. We truly appreciate your great contribution!

Issue #1836, opened 2025-04-08
Improve error messages by adding helpful context


douglascamata:
**Is your feature request related to a problem? Please describe.**

I'm currently trying to debug an issue when running `github.com/shirou/gopsutil/v4/host.Info()` in a Windows 2019 k8s node and for some reason it fails (I could already figure out the reason, but it's not the focus here). The error message is short and not helpful:

```
The system cannot find the file specified.
```

There's no information about which one of the `InfoStat` fields caused the issue and which file cannot be found.

**Describe the solution you'd like**

I would like the error to provide all the additional context related to the error. For example, if the problem happens when trying to fill in the `InfoStat.KernelVersion` field:

```
KernelVersion: the system cannot find the file "<whatever file it tried to find>"
```

**Describe alternatives you've considered**

None so far.

**Additional context**

None so far.

s0ders:
Working on a refactor (on the `host` package) to use `fmt.Errorf` on some key error returns to add some textual context in order to make the error message more helpful.

s0ders:
The PR has been merged, I believe this should fix your issue @douglascamata

douglascamata:
@s0ders thank you!

PR #1840, opened 2025-04-17
refactor: using fmt.Errorf in InfoWithContext

s0ders:
The following brings:
- Refactoring of `InfoWithContext` to wrap errors with `fmt.Errorf` in order to add textual context, making error messages more helpful for the end user.

ref: #1836

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


ccoVeille (approved):

shirou (approved):
You're absolutely right. Improving error messages like this PR is definitely important. Thank you!

PR #1842, opened 2025-04-19
chore: enable staticcheck

mmorel-35:
#### Description

- Adding `staticcheck` to the list of enabled linters in the Golang CI configuration 
- It excludes all the rules with which the code doesn't comply yet and will be fixed in dedicated PR and fixed the quick fixes.

### DIFF
--- a/cpu/cpu_solaris.go
+++ b/cpu/cpu_solaris.go
@@ -159,10 +159,7 @@ func parseISAInfo(cmdOutput string) ([]string, error) {
 		return nil, errors.New("attempted to parse invalid isainfo output")
 	}
 
-	flags := make([]string, len(words)-4)
-	for i, val := range words[4:] { //nolint:gosimple //FIXME
-		flags[i] = val
-	}
+	flags := words[4:]
 	sort.Strings(flags)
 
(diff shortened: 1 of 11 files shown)


cpu/cpu_solaris.go
```diff
@@ -160,9 +160,7 @@ func parseISAInfo(cmdOutput string) ([]string, error) {
 	}
 
 	flags := make([]string, len(words)-4)
-	for i, val := range words[4:] { //nolint:gosimple //FIXME
-		flags[i] = val
-	}
+	copy(flags, words[4:])
 	sort.Strings(flags)
```
ccoVeille, line 164: I don't understand the code was very complicated, but now it's clearer.

What about using this?

```go
         flags := words[4:]
         sort.Strings(flags)
```

shirou: @ccoVeille Thank you for your suggestion. I agree with you. 
Could you please update the code? > @mmorel-35

ccoVeille (approved):

shirou (approved):
Thank you so much!

PR #1848, opened 2025-05-07
Fix issues with GetLastError() usage

secDre4mer:
Fix some issues with incorrect `GetLastError` usage:

- Use the error returned from `Proc.Call` instead, which is not racy in case of thread switches
- Don't check `GetLastError` after `GetDriveType`, since it does not set it

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -71,12 +71,14 @@ func TimesWithContext(_ context.Context, percpu bool) ([]TimesStat, error) {
 	var lpIdleTime common.FILETIME
 	var lpKernelTime common.FILETIME
 	var lpUserTime common.FILETIME
-	r, _, _ := common.ProcGetSystemTimes.Call(
+	// GetSystemTimes returns 0 for error, in which case we check err,
+	// see https://pkg.go.dev/golang.org/x/sys/windows#LazyProc.Call
+	r, _, err := common.ProcGetSystemTimes.Call(
 		uintptr(unsafe.Pointer(&lpIdleTime)),
 		uintptr(unsafe.Pointer(&lpKernelTime)),
 		uintptr(unsafe.Pointer(&lpUserTime)))
 	if r == 0 {
-		return ret, windows.GetLastError()
+		return nil, err
 	}
 
 	LOT := float64(0.0000001)
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -19,7 +19,6 @@ import (
 var (
 	procGetDiskFreeSpaceExW     = common.Modkernel32.NewProc("GetDiskFreeSpaceExW")
 	procGetLogicalDriveStringsW = common.Modkernel32.NewProc("GetLogicalDriveStringsW")
-	procGetDriveType            = common.Modkernel32.NewProc("GetDriveTypeW")
 	procGetVolumeInformation    = common.Modkernel32.NewProc("GetVolumeInformationW")
 )
 
@@ -109,15 +108,14 @@ func PartitionsWithContext(ctx context.Context, _ bool) ([]PartitionStat, error)
 			if v >= 65 && v <= 90 {
 				path := string(v) + ":"
 				typepath, _ := windows.UTF16PtrFromString(path)
-				typeret, _, _ := procGetDriveType.Call(uintptr(unsafe.Pointer(typepath)))
-				if typeret == 0 {
-					err := windows.GetLastError()
-					warnings.Add(err)
+				typeret := windows.GetDriveType(typepath)
+				switch typeret {
+				case windows.DRIVE_UNKNOWN:
 					continue
-				}
-				// 2: DRIVE_REMOVABLE 3: DRIVE_FIXED 4: DRIVE_REMOTE 5: DRIVE_CDROM
-
-				if typeret == 2 || typeret == 3 || typeret == 4 || typeret == 5 {
+				case windows.DRIVE_REMOVABLE,
+					windows.DRIVE_FIXED,
+					windows.DRIVE_REMOTE,
+					windows.DRIVE_CDROM:
 					lpVolumeNameBuffer := make([]byte, 256)
 					lpVolumeSerialNumber := int64(0)
 					lpMaximumComponentLength := int64(0)
@@ -134,7 +132,8 @@ func PartitionsWithContext(ctx context.Context, _ bool) ([]PartitionStat, error)
 						uintptr(unsafe.Pointer(&lpFileSystemNameBuffer[0])),
 						uintptr(len(lpFileSystemNameBuffer)))
 					if driveret == 0 {
-						if typeret == 2 || typeret == 4 || typeret == 5 {
+						switch typeret {
+						case windows.DRIVE_REMOVABLE, windows.DRIVE_REMOTE, windows.DRIVE_CDROM:
 							continue // device is not ready will happen if there is no disk in the drive
 						}
 						warnings.Add(err)
@@ -197,9 +196,6 @@ func IOCountersWithContext(_ context.Context, names ...string) (map[string]IOCou
 			path := string(rune(v)) + ":"
 			typepath, _ := windows.UTF16PtrFromString(path)
 			typeret := windows.GetDriveType(typepath)
-			if typeret == 0 {
-				return drivemap, windows.GetLastError()
-			}
 			if typeret != windows.DRIVE_FIXED {
 				continue
 			}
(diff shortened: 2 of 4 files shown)


mem/ex_windows.go
```diff
@@ -28,16 +26,16 @@ func NewExWindows() *ExWindows {
 func (e *ExWindows) VirtualMemory() (*ExVirtualMemory, error) {
 	var memInfo memoryStatusEx
 	memInfo.cbSize = uint32(unsafe.Sizeof(memInfo))
-	mem, _, _ := procGlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memInfo)))
+	mem, _, err := procGlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memInfo)))
 	if mem == 0 {
```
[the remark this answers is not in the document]
secDre4mer: According to https://pkg.go.dev/syscall?GOOS=windows#Proc.Call:

> The returned error is always non-nil, constructed from the result of GetLastError. Callers must inspect the primary return value to decide whether an error occurred (according to the semantics of the specific function being called) before consulting the error. The error always has type [Errno](https://pkg.go.dev/syscall?GOOS=windows#Errno).

So the canonical way is:

 - check the return value (in this case `mem`) on whether an error occurred
 - If one occurred and the called function stores error details in `GetLastError`, use `err`

mem/ex_windows.go
```diff
@@ -28,16 +26,16 @@ func NewExWindows() *ExWindows {
...
-		return nil, windows.GetLastError()
+		return nil, err
 	}
 
 	var perfInfo performanceInformation
 	perfInfo.cb = uint32(unsafe.Sizeof(perfInfo))
-	perf, _, _ := procGetPerformanceInfo.Call(uintptr(unsafe.Pointer(&perfInfo)), uintptr(perfInfo.cb))
+	perf, _, err := procGetPerformanceInfo.Call(uintptr(unsafe.Pointer(&perfInfo)), uintptr(perfInfo.cb))
 	if perf == 0 {
```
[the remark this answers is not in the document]
secDre4mer: See discussion above.

mem/ex_windows.go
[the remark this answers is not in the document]
ccoVeille: Thanks @secDre4mer 

But it's neither obvious, not matching the classic way to handle errors in Go

So I would appreciate if you could add a comment about it here and at the other place reported by @mmorel-35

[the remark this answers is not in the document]
secDre4mer: Done.

disk/disk_windows.go
```diff
@@ -109,15 +108,14 @@ func PartitionsWithContext(ctx context.Context, _ bool) ([]PartitionStat, error)
...
+				if typeret == windows.DRIVE_UNKNOWN {
 					continue
 				}
-				// 2: DRIVE_REMOVABLE 3: DRIVE_FIXED 4: DRIVE_REMOTE 5: DRIVE_CDROM
-
-				if typeret == 2 || typeret == 3 || typeret == 4 || typeret == 5 {
+				if typeret == windows.DRIVE_REMOVABLE ||
+					typeret == windows.DRIVE_FIXED ||
+					typeret == windows.DRIVE_REMOTE ||
+					typeret == windows.DRIVE_CDROM {
```
ccoVeille, line 118: Using a `switch` would improve readability

disk/disk_windows.go
```diff
@@ -134,7 +132,9 @@ func PartitionsWithContext(ctx context.Context, _ bool) ([]PartitionStat, error)
 						uintptr(unsafe.Pointer(&lpFileSystemNameBuffer[0])),
 						uintptr(len(lpFileSystemNameBuffer)))
 					if driveret == 0 {
-						if typeret == 2 || typeret == 4 || typeret == 5 {
+						if typeret == windows.DRIVE_REMOVABLE ||
+							typeret == windows.DRIVE_REMOTE ||
+							typeret == windows.DRIVE_CDROM {
```
ccoVeille, line 137: Using a `switch` would improve readability

cpu/cpu_windows.go
```diff
@@ -71,12 +71,12 @@ func TimesWithContext(_ context.Context, percpu bool) ([]TimesStat, error) {
...
 	var lpUserTime common.FILETIME
-	r, _, _ := common.ProcGetSystemTimes.Call(
+	r, _, err := common.ProcGetSystemTimes.Call(
 		uintptr(unsafe.Pointer(&lpIdleTime)),
 		uintptr(unsafe.Pointer(&lpKernelTime)),
 		uintptr(unsafe.Pointer(&lpUserTime)))
 	if r == 0 {
-		return ret, windows.GetLastError()
+		return nil, err
 	}
```
ccoVeille, line 80: A comment equivalent to what you wrote in other file would be appreciated.

Because here, it feels like some errors could be ignored

mem/mem_windows.go
```diff
@@ -40,9 +40,9 @@ func VirtualMemory() (*VirtualMemoryStat, error) {
 func VirtualMemoryWithContext(_ context.Context) (*VirtualMemoryStat, error) {
 	var memInfo memoryStatusEx
 	memInfo.cbSize = uint32(unsafe.Sizeof(memInfo))
-	mem, _, _ := procGlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memInfo)))
+	mem, _, err := procGlobalMemoryStatusEx.Call(uintptr(unsafe.Pointer(&memInfo)))
 	if mem == 0 {
-		return nil, windows.GetLastError()
+		return nil, err
 	}
```
ccoVeille, line 46: A comment equivalent to what you wrote in other file would be appreciated.

Because here, it feels like some errors could be ignored

mem/mem_windows.go
```diff
@@ -93,9 +93,9 @@ func SwapMemoryWithContext(_ context.Context) (*SwapMemoryStat, error) {
 	// Get total memory from performance information
 	var perfInfo performanceInformation
 	perfInfo.cb = uint32(unsafe.Sizeof(perfInfo))
-	mem, _, _ := procGetPerformanceInfo.Call(uintptr(unsafe.Pointer(&perfInfo)), uintptr(perfInfo.cb))
+	mem, _, err := procGetPerformanceInfo.Call(uintptr(unsafe.Pointer(&perfInfo)), uintptr(perfInfo.cb))
 	if mem == 0 {
-		return nil, windows.GetLastError()
+		return nil, err
 	}
```
ccoVeille, line 99: A comment equivalent to what you wrote in other file would be appreciated.

Because here, it feels like some errors could be ignored

mem/mem_windows.go
```diff
@@ -161,9 +161,9 @@ func SwapDevicesWithContext(_ context.Context) ([]*SwapDevice, error) {
 	// the following system call invokes the supplied callback function once for each page file before returning
 	// see https://docs.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumpagefilesw
 	var swapDevices []*SwapDevice
-	result, _, _ := procEnumPageFilesW.Call(windows.NewCallback(pEnumPageFileCallbackW), uintptr(unsafe.Pointer(&swapDevices)))
+	result, _, err := procEnumPageFilesW.Call(windows.NewCallback(pEnumPageFileCallbackW), uintptr(unsafe.Pointer(&swapDevices)))
 	if result == 0 {
-		return nil, windows.GetLastError()
+		return nil, err
 	}
```
ccoVeille, line 167: A comment equivalent to what you wrote in other file would be appreciated.

Because here, it feels like some errors could be ignored

mem/mem_windows.go
secDre4mer: Done

mem/mem_windows.go
secDre4mer: Done

cpu/cpu_windows.go
secDre4mer: Done

disk/disk_windows.go
secDre4mer: Done

mem/mem_windows.go
secDre4mer: Done

ccoVeille (approved):

shirou (approved):
Thank you very much for your significant contribution! You're right — it seems this `GetLastError` is unnecessary. (Note: [Maybe around here](https://github.com/golang/go/blob/5ff38e476177ce9e67375bd010bea2e030f2fe19/src/runtime/sys_windows_amd64.s#L66)?)

PR #1855, opened 2025-05-17
use strict exclusion on generated files

mmorel-35:
#### Description

This updates golangci-lint to use strict generated exclusion.
See [the documentation](https://golangci-lint.run/usage/configuration/#linters-configuration)

It allows to remove lint exclusions concerning var namings.


.golangci.yml
```diff
@@ -129,7 +129,7 @@ linters:
       os-create-temp: false
       os-mkdir-temp: false
   exclusions:
-    generated: lax
+    generated: strict
```
ccoVeille, line 132: `strict` is the default value

```suggestion
```

disk/disk_freebsd_386.go
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BSD-3-Clause
-// Created by cgo -godefs - DO NOT EDIT
+// Code generated by cgo -godefs - DO NOT EDIT.
 // cgo -godefs types_freebsd.go
```
ccoVeille, line 3: If the files are generated, don't you have to edit a file that contains the header you expect?

I mean something generated these files, and I don't see something about updating what generate them

ccoVeille: Yes, that's what I'm worried about.

If someone regenerates the files, all files will have changes because you modified the header manually.

ccoVeille: In this PR, did you update the files whose headers are updated by running mktypes.sh or did you update them manually?

I mean if you switch to a clean branch, will you get the header updated by launching mytypes.sh

I mean if the files are updated, it shows that the way go too cgo works was updated and we needed to update these headers.

So it could have been done in another PR.

Otherwise, I would keep these files unchanged, and keep the lax settings.

Maybe @shirou would have a better understanding than me on how these files are maintained.

ccoVeille: Said otherwise, I feel like this PR should be closed.

Because using lax here is not that a problem.

But, via another issue we should make sure there is a process in the CI to check if generated files don't need to be updated.

I feel like this is missing and done manually right now.

PR #1856, opened 2025-05-17
enable all go-critic

mmorel-35:
#### Description

Enable and fix more go-critic rules

### DIFF
--- a/cpu/cpu_plan9_test.go
+++ b/cpu/cpu_plan9_test.go
@@ -35,7 +35,7 @@ var timesTests = []struct {
 func TestTimesPlan9(t *testing.T) {
 	for _, tt := range timesTests {
 		t.Run(tt.mockedRootFS, func(t *testing.T) {
-			t.Setenv("HOST_ROOT", filepath.Join("testdata/plan9", tt.mockedRootFS))
+			t.Setenv("HOST_ROOT", filepath.Join("testdata", "plan9", tt.mockedRootFS))
 			stats, err := Times(false)
 			common.SkipIfNotImplementedErr(t, err)
 			require.NoError(t, err)
--- a/disk/disk_freebsd.go
+++ b/disk/disk_freebsd.go
@@ -33,7 +33,8 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
 		return ret, err
 	}
 
-	for _, stat := range fs {
+	for i := range fs {
+		stat := &fs[i]
 		opts := []string{"rw"}
 		if stat.Flags&unix.MNT_RDONLY != 0 {
 			opts = []string{"ro"}
@@ -182,7 +183,8 @@ func SerialNumberWithContext(ctx context.Context, name string) (string, error) {
 			break
 		}
 	}
-	if err = s.Err(); err != nil {
+	err = s.Err()
+	if err != nil {
 		return "", err
 	}
 	return serial, nil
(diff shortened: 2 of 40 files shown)


disk/disk_freebsd.go
```diff
@@ -182,7 +182,8 @@ func SerialNumberWithContext(ctx context.Context, name string) (string, error) {
 			break
 		}
 	}
-	if err = s.Err(); err != nil {
+	err = s.Err()
+	if err != nil {
 		return "", err
 	}
```
ccoVeille, line 188: I'm curious. What go-critic rule implied this?

cpu/cpu_plan9_test.go
```diff
@@ -35,7 +35,7 @@ var timesTests = []struct {
 func TestTimesPlan9(t *testing.T) {
 	for _, tt := range timesTests {
 		t.Run(tt.mockedRootFS, func(t *testing.T) {
-			t.Setenv("HOST_ROOT", filepath.Join("testdata/plan9", tt.mockedRootFS))
+			t.Setenv("HOST_ROOT", filepath.Join("testdata", "plan9", tt.mockedRootFS))
```
ccoVeille, line 38: Is there a linter in go-critic that reported it?

I was looking for it a few months ago

disk/disk_linux.go
```diff
@@ -486,7 +486,7 @@ func IOCountersWithContext(ctx context.Context, names ...string) (map[string]IOC
 	return ret, nil
 }
 
-func udevData(ctx context.Context, major uint32, minor uint32, name string) (string, error) {
+func udevData(ctx context.Context, major, minor uint32, name string) (string, error) {
```
ccoVeille, line 489: I know such changes are implied by gofumpt "extra" rule, is this one linked to a go-critic rule?

load/load_aix_nocgo.go
```diff
@@ -20,7 +21,7 @@ func AvgWithContext(ctx context.Context) (*AvgStat, error) {
 		return nil, err
 	}
 
-	idx := strings.Index(string(line), "load average:")
+	idx := bytes.Index(line, []byte("load average:"))
```
ccoVeille, line 24: Was it reported by gocritic? It looks alike something reported by `mirror` linter

ccoVeille (approved):

ccoVeille:
Thanks for pointing out each rule that were involved.

process/process_windows.go
```diff
@@ -882,7 +882,8 @@ func getFromSnapProcess(pid int32) (int32, int32, string, error) { //nolint:unpa
 	defer windows.CloseHandle(snap)
 	var pe32 windows.ProcessEntry32
 	pe32.Size = uint32(unsafe.Sizeof(pe32))
-	if err = windows.Process32First(snap, &pe32); err != nil {
+	err = windows.Process32First(snap, &pe32)
+	if err != nil {
```
shirou, line 886: Is this change perhaps intended to a code like this?
```suggestion
	if err := windows.Process32First(snap, &pe32); err != nil {
```

shirou: Hmm, I think the original code might be incorrect. It was probably supposed to use `if err :=`.

I'm also curious why [this code](https://github.com/shirou/gopsutil/pull/1856/files/cf5debe6a1b2d1a705cc557ef0c57857cdf03dfd#diff-ae26c3608d9221e521907319158f351ef1a80515c3dc5b2e780c5eda6758f5a8R894) wasn't detected.

PR #1861, opened 2025-05-24
[disk][linux]: fix parsing mount option when use 1/mounts on Partition

### LINKED ISSUE #1858: Mount options does not split by comma
**Describe the bug**
In `PartitionsWithContext()`, the value of`PartitionStat.Opts` doesn't use `,` when splition. So the options are in one string. As contrast, mount options are split when using `/proc/1/mountinfo`.
https://github.com/shirou/gopsutil/blob/34ac457e154c695cf35ed79ffa15eb0966c650bb/disk/disk_linux.go#L292-L325

**To Reproduce**
```go
package main

import (
	"fmt"

	"github.com/shirou/gopsutil/v4/disk"
)

func main() {
	partitions, err := disk.Partitions(true)
	if err != nil {
		log.Fatalln(err)
	}
	for _, each := range partitions {
		fmt.Printf("+%v\n", each)
	}
}
```

**Expected behavior**
Mount options of each mountpoint should be split by `,`.

 - [x] Linux: 

**Additional context**

shirou:
This fixes #1858.

On `disk.Parition()` When we can not read `/proc/1/mountinfo` because version <2.6.26, fallback to read `/proc/1/mounts`. However, mount options are not splitted. 

Here is a example line of `mounts`, 

```
tmpfs /run/user/1000 tmpfs rw,nosuid,nodev,relatime,size=1285356k,nr_inodes=321339,mode=700,uid=1000,gid=1000 0 0
```

and this is parsed like this, `opts` is just one long line.

```
{"device":"tmpfs","mountpoint":"/run/user/1000","fstype":"tmpfs","opts":["rw,nosuid,nodev,relatime,size=1285356k,nr_inodes=321339,mode=700,uid=1000,gid=1000"]}
```

This PR fixes this issue.

## Add tests

This PR also split parse to `parseFieldsOnMounts` and `parseFieldsOnMountinfo`. These functions are used according to which file is used. And also introduce tests for those functions.

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -296,7 +296,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 				Device:     fields[0],
 				Mountpoint: unescapeFstab(fields[1]),
 				Fstype:     fields[2],
-				Opts:       strings.Fields(fields[3]),
+				Opts:       strings.Split(fields[3], ","),
 			}
 
 			if !all {
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -285,81 +285,99 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 		return nil, err
 	}
 
+	var ret []PartitionStat
+	if useMounts { // use mounts file
+		ret = parseFieldsOnMounts(lines, all, fs)
+	} else { // use mountinfo
+		ret, err = parseFieldsOnMountinfo(ctx, lines, all, fs, filename)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing mountinfo file %s: %w", filename, err)
+		}
+	}
+
+	return ret, nil
+}
+
+func parseFieldsOnMounts(lines []string, all bool, fs []string) []PartitionStat {
 	ret := make([]PartitionStat, 0, len(lines))
-
 	for _, line := range lines {
-		var d PartitionStat
-		if useMounts {
-			fields := strings.Fields(line)
+		fields := strings.Fields(line)
 
-			d = PartitionStat{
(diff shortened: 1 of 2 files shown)


disk/disk_linux.go
```diff
@@ -296,7 +296,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 				Device:     fields[0],
 				Mountpoint: unescapeFstab(fields[1]),
 				Fstype:     fields[2],
-				Opts:       strings.Fields(fields[3]),
+				Opts:       strings.Split(fields[3], ","),
```
ccoVeille, line 299: No test to update/add for the issue you describe?

disk/disk_linux.go
```diff
@@ -285,81 +285,99 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
...
+	if useMounts { // use mounts file
+		ret = parseFieldsOnMounts(lines, all, fs)
+	} else { // use mountinfo
+		ret, err = parseFieldsOnMountinfo(ctx, lines, all, fs, filename)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing mountinfo file %s: %w", filename, err)
+		}
+	}
 
+	return ret, nil
```
ccoVeille: This can be simplified, it would be more readable


```suggestion
	if useMounts { // use mounts file
		ret = parseFieldsOnMounts(lines, all, fs)
		return ret, nil
	}

	// use mountinfo
	ret, err = parseFieldsOnMountinfo(ctx, lines, all, fs, filename)
	if err != nil {
		return nil, fmt.Errorf("error parsing mountinfo file %s: %w", filename, err)
	}
 
	return ret, nil
```

ccoVeille (approved):

PR #1863, opened 2025-05-25
enable shadow from govet

mmorel-35:
#### Description

This enable and fixes shadow rule from [govet](https://golangci-lint.run/usage/linters/#govet).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -62,7 +62,6 @@ linters:
       disable:
         - copylocks
         - fieldalignment
-        - shadow
         - testinggoroutine
       enable-all: true
     perfsprint:
@@ -143,6 +142,10 @@ linters:
       - common-false-positives
       - legacy
       - std-error-handling
+    rules:
+      - text: 'shadow: declaration of "err" shadows declaration at line'
+        linters:
+          - govet
     warn-unused: true
 output:
   show-stats: false
--- a/cpu/cpu_netbsd.go
+++ b/cpu/cpu_netbsd.go
@@ -36,7 +36,8 @@ func Times(percpu bool) ([]TimesStat, error) {
 	return TimesWithContext(context.Background(), percpu)
 }
 
-func TimesWithContext(_ context.Context, percpu bool) (ret []TimesStat, err error) {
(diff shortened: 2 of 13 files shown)


ccoVeille (approved):

shirou:
Sorry for the delayed review.

I’m in favor of adding linting for shadowing. However, after reading this PR twice, I still feel that applying it to `err` might be overkill.

For example, take the following code:

```golang
if err := common.Sleep(ctx, interval); err != nil {
```

Changing it to:
```golang
err := common.Sleep(ctx, interval)
if err != nil {
```

feels a bit much.

The former is a common idiom in Go and makes it clear that we're only dealing with the `err` returned by `common.Sleep`, which seems preferable.

Since `err` is such a commonly used variable name in Go, I feel like some amount of shadowing is unavoidable. Is there a way to configure the linter to ignore shadowing specifically for `err`?

ccoVeille:
It works. The issue comes often from the very first err that is define in the scope, that implies everything else will be reported as shadowed.

The solution is to use the `if err := whatever(); err != nil {` everywhere

shirou (commented):
Sorry for the very late reply. I've been thinking about this for a while. It's true that detecting variable shadowing has many advantages, however I still believe that universally disallowing shadowing of `err` hurts code readability.
We've had various discussions and you've made several changes to the PR, but I don't think my stance on this has changed. I truly appreciate your dedicated contributions.

.golangci.yml
```diff
@@ -142,6 +141,10 @@ linters:
       - common-false-positives
       - legacy
       - std-error-handling
+    rules:
+      - text: 'shadow: declaration of "err" shadows declaration at line'
+        linters:
+          - govet
```
mmorel-35, line 148: I followed @ccoVeille advice and rolled-back `err` related changes. Shall I just close this PR or does it fits better your expectations @shirou ?

shirou (approved):
Thank you for the thoughtful discussion and updates. I've reviewed the changes, and I believe they are now clear and understandable. I will go ahead and merge the PR.

PR #1864, opened 2025-05-28
Restored network counters for a process

aknopov:
Restored functionality for counting network IO per process using Pcap library.

- This change requires asynchronous accumulation of network counts that provided by Pcap
- The counting starts by calling `net.CollectNetIO(...)` with all count initiated to 0
- Pcap packets are associated with a process using `ConnectionStat` data
- `ConnectionStat` is requested periodically with interval specified in `net.CollectNetIO(...)` parameter
- Some connections last less than this interval and the implementation makes an effort to figure associated Pid of these connections.
- Since Pcap doesn't provide packets for Unix sockets , "unix" connection kind is assumed as "all"

On the bright side
- Code probably works for OS'es other than Windows or Linux but this wasn't tested

**Other code fixes** - I took the liberty to fix some errors that didn't allow the project to compile in my environment.


internal/common/common_windows.go
```diff
@@ -290,15 +290,15 @@ func CallWithExpandingBuffer(fn func() NtStatus, buf *[]byte, resultLength *uint
 }
 
 func NtQuerySystemInformation(
-	SystemInformationClass uint32,
-	SystemInformation *byte,
-	SystemInformationLength uint32,
-	ReturnLength *uint32,
+	systemInformationClass uint32,
```
aknopov, line 293: Linter didn't like upper case

net/net_pcap.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+}
+
+func ipMatch(ip net.IP, nicAddr pcap.InterfaceAddress) bool {
+	return slices.Equal(ip.Mask(nicAddr.Netmask), nicAddr.IP.Mask(nicAddr.Netmask))
+}
+
+// Figures which address is local and which remote
+func sortAddresses(addr1 *Addr, addr2 *Addr, dev *pcap.Interface) (*Addr, *Addr, error) {
+	ip1 := net.ParseIP(addr1.IP)
```
aknopov, line 123: It would be more performant to use `net.IP` in `Addr`

sensors/sensors_windows.go
```diff
@@ -19,7 +20,21 @@ type msAcpi_ThermalZoneTemperature struct { //nolint:revive //FIXME
 	InstanceName       string
 }
 
+type win32_PerfFormattedData_Counters_ThermalZoneInformation struct { //nolint:revive //FIXME
+	Name                     string
+	HighPrecisionTemperature uint32
+}
+
 func TemperaturesWithContext(ctx context.Context) ([]TemperatureStat, error) {
+	ret, err := acpiQuery(ctx)
```
aknopov, line 29: Previous implementation require elevated privileges. Added fallback to Win32 API though it has less parameters

shirou (changes requested):
Thank you very much for your PR. It was very informative, and I appreciate the opportunity to see how such a feature could be implemented.

As a general rule, we do not usually add functionality that is not already included in psutil. However, I did consider including this feature for the following reasons:

- There is a system-wide NetIOCounter.
- The functionality exists across all major operating systems.

That said, it appears that github.com/google/gopacket, which is used in this PR, requires CGO. Since gopsutil aims to remain a pure Go library, we want to avoid dependencies that require CGO.

Would there be an alternative implementation that does not rely on CGO?

PR #1866, opened 2025-06-06
fix: optimize performance of process.OpenFiles()

### LINKED ISSUE #1781: process.OpenFiles() does not scale well and consumes ton of CPU
**Describe the bug**

I have an issue with fetching open files for a process. I am using https://github.com/shirou/gopsutil. Calling `process.OpenFiles()` which consumes a lot of CPU, profiling link: https://pprof.me/171f81b809d99641cf6a32fee8d78eba/?profileType=profile%253Acpu%253Ananoseconds%253Acpu%253Ananoseconds%253Adelta&color_by=filename. 

**To Reproduce**

Run any program and call process.OpenFiles(). Make sure to call it on envs with around 1k processes and for each process around 200 fds.

**Expected behavior**

Expected not to stress the CPU that much. I had the same code with python where I call https://github.com/giampaolo/psutil/blob/d4b37d9628e634c00a2990e488b1cc38ea6a41b5/psutil/_pslinux.py#L2225. And it's not hogging that much CPU, I don't understand though because both code they perform the same thing: read the directory under /proc/<pid>/fd and then resolve the links. Apparently os.ReadLink is very costly. 

 - [X] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]

NAME="Red Hat Enterprise Linux"
VERSION="9.4 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.4"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.4 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.4
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.4"

NitroCao:
fixes #1781

### DIFF
--- a/internal/common/readlink_linux.go
+++ b/internal/common/readlink_linux.go
@@ -38,14 +38,16 @@ func fixCount(n int, err error) (int, error) {
 // Readlink behaves like os.Readlink but caches the buffer passed to syscall.Readlink.
 func Readlink(name string) (string, error) {
 	b := bufferPool.Get().(*[]byte)
-	defer bufferPool.Put(b)
 
 	n, err := ignoringEINTR2(func() (int, error) {
 		return fixCount(syscall.Readlink(name, *b))
 	})
 	if err != nil {
+		bufferPool.Put(b)
 		return "", &os.PathError{Op: "readlink", Path: name, Err: err}
 	}
 
-	return string((*b)[:n]), nil
+	result := string((*b)[:n])
+	bufferPool.Put(b)
+	return result, nil
 }


NitroCao:
Is it good and necessary to provide an iterator method for `(*Process).OpenFiles()` called `OpenFilesSeq()` to return a `iter.Seq[OpenFilesStat]`?

shirou:
The Iterator is not necessary. However, since Go 1.23 —where Iterator was introduced— is the oldest Go version we support, it’s fine to use it, I think.

NitroCao:
@shirou Please merge the PR if you think everything is ok.

internal/common/readlink_linux.go
```diff
@@ -0,0 +1,51 @@
...
+
+	n, err := ignoringEINTR2(func() (int, error) {
+		return fixCount(syscall.Readlink(name, *b))
+	})
+	if err != nil {
+		return "", &os.PathError{Op: "readlink", Path: name, Err: err}
+	}
+
+	return string((*b)[:n]), nil
+}
```
shirou: How about to remove `defer`? `defer` is safer but it costs a little.
```suggestion
func Readlink(name string) (string, error) {
	b := bufferPool.Get().(*[]byte)
        // avoid derfer to optimization. This is very special tuning.
	n, err := ignoringEINTR2(func() (int, error) {
		bufferPool.Put(b)
		return fixCount(syscall.Readlink(name, *b))
	})
	if err != nil {
		bufferPool.Put(b)
		return "", &os.PathError{Op: "readlink", Path: name, Err: err}
	}

	bufferPool.Put(b)
	return string((*b)[:n]), nil
}
```

I take a benchmark using [the sample code](https://github.com/shirou/gopsutil/issues/1781#issuecomment-2888888902).

Using defer have a great improvement.
```
goos: linux
goarch: amd64
pkg: github.com/shirou/gopsutil/v4/process
cpu: AMD Ryzen 7 5800HS with Radeon Graphics
                         │   old.txt   │            new.txt            │
                         │   sec/op    │   sec/op     vs base          │
FillFromfdWithContext-16   57.51µ ± 3%   57.70µ ± 6%  ~ (p=0.529 n=10)

                         │   old.txt    │               new.txt                │
                         │     B/op     │     B/op      vs base                │
FillFromfdWithContext-16   1.999Ki ± 0%   1.123Ki ± 0%  -43.82% (p=0.000 n=10)

                         │  old.txt   │              new.txt               │
                         │ allocs/op  │ allocs/op   vs base                │
FillFromfdWithContext-16   41.00 ± 0%   34.00 ± 0%  -17.07% (p=0.000 n=10)
```

But if not using defer, it looks better.

```
goos: linux
goarch: amd64
pkg: github.com/shirou/gopsutil/v4/process
cpu: AMD Ryzen 7 5800HS with Radeon Graphics
                         │   old.txt    │               new.txt               │
                         │    sec/op    │   sec/op     vs base                │
FillFromfdWithContext-16   57.509µ ± 3%   6.917µ ± 5%  -87.97% (p=0.000 n=10)

                         │   old.txt   │              new.txt               │
                         │    B/op     │    B/op     vs base                │
FillFromfdWithContext-16   2047.0 ± 0%   149.0 ± 0%  -92.72% (p=0.000 n=10)

                         │   old.txt   │              new.txt               │
                         │  allocs/op  │ allocs/op   vs base                │
FillFromfdWithContext-16   41.000 ± 0%   6.000 ± 0%  -85.37% (p=0.000 n=10)
```

I usually prefer the safety of `defer`, but in this case, speed is the main concern. And this is a short function, so I don't think it is a problem. What do you think?

PR #1868, opened 2025-06-16
feat(exwindows): expose phys and pagefile memory

kruskall:
add a way to retrieve pagefile info from memorystatusex
add phys memory for usability


shirou (changes requested):

Just to clarify, the values "TotalPageFile/AvailPageFile" here also include physical memory size, so they are not exactly equivalent to "swap" on Linux. Since the Ex structure doesn’t need to match the field names used on other platforms, would you mind changing the field names to `PageFileTotal` and `PageFileAvail`, which are closer to the original naming?

PR #1882, opened 2025-07-14
[mem][linux]: Used calculated from Total and Available

### LINKED ISSUE #1873: Consider using MemAvailable for Used memory calculation

Current memory "used" calculation is based on the same old formula used by the `free` command on Linux: [Used = Total - Free - Buffers - Cached](https://github.com/shirou/gopsutil/blob/master/mem/mem_linux.go#L307)

The issue is that this formula is not accurate because the `Cached` includes memory that is not freeable. That's why the Linux Kernel 3.14 introduced the `MemAvailable` state: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

For that reason, tools like `free` or `top` changed their used formula to `Used = Total - Available` since 
since version `v4.0.1`, [reference commit](https://gitlab.com/procps-ng/procps/-/commit/2184e90d2).

Comments on the [Used memory on Linux includes Slab](https://github.com/shirou/gopsutil/issues/899) rely on the accuracy of old `free` formula.

See procs-ps calculation: https://gitlab.com/procps-ng/procps/-/blob/v4.0.1/library/meminfo.c#L702

Rely on the kernel metrics to calculate the used memory: https://elixir.bootlin.com/linux/v6.15.4/source/mm/show_mem.c#L32

shirou:
fix #1873 

`mem.Used` calculated from `Total - Free - Buffers - Cache` even if `MemAvailable` can be used.

### DIFF
--- a/mem/mem_linux.go
+++ b/mem/mem_linux.go
@@ -296,15 +296,17 @@ func fillFromMeminfoWithContext(ctx context.Context) (*VirtualMemoryStat, *ExVir
 
 	ret.Cached += ret.Sreclaimable
 
-	if !memavail {
+	if memavail {
+		ret.Used = ret.Total - ret.Available
+	} else {
 		if activeFile && inactiveFile && sReclaimable {
 			ret.Available = calculateAvailVmem(ctx, ret, retEx)
 		} else {
 			ret.Available = ret.Cached + ret.Free
 		}
+		ret.Used = ret.Total - ret.Free - ret.Buffers - ret.Cached
 	}
 
-	ret.Used = ret.Total - ret.Free - ret.Buffers - ret.Cached
 	ret.UsedPercent = float64(ret.Used) / float64(ret.Total) * 100.0
 
 	return ret, retEx, nil
--- a/mem/mem_linux_test.go
+++ b/mem/mem_linux_test.go
@@ -33,8 +33,8 @@ var virtualMemoryTests = []struct {
 		"intelcorei5", &VirtualMemoryStat{
 			Total:          16502300672,
 			Available:      11495358464,
-			Used:           3437277184,
-			UsedPercent:    20.82907863769651,
+			Used:           5006942208,
+			UsedPercent:    30.340873721295385,
 			Free:           8783491072,
 			Active:         4347392000,
 			Inactive:       2938834944,
@@ -74,8 +74,8 @@ var virtualMemoryTests = []struct {
 		"issue1002", &VirtualMemoryStat{
 			Total:          260579328,
 			Available:      215199744,
-			Used:           34328576,
-			UsedPercent:    13.173944481121694,
+			Used:           45379584,
+			UsedPercent:    17.414882580401773,
 			Free:           124506112,
 			Active:         108785664,
 			Inactive:       8581120,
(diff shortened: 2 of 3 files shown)


mem/mem_linux.go
```diff
@@ -296,15 +296,17 @@ func fillFromMeminfoWithContext(ctx context.Context) (*VirtualMemoryStat, *ExVir
 
 	ret.Cached += ret.Sreclaimable
 
-	if !memavail {
+	if memavail {
```
braydonk: I wonder if this should be opt-out/opt-in? Might be a breaking change for users, or if not breaking at least might be an unnoticed change that makes their stats look different all of a sudden.

The other stats are still available in `struct VirtualMemoryStat` so it's possible for consumers to redo the existing calculation on their own, but some manner of `ctx` option to revert to the old calculation might be a good idea?

mem/mem_linux.go
```diff
@@ -296,15 +303,35 @@ func fillFromMeminfoWithContext(ctx context.Context) (*VirtualMemoryStat, *ExVir
...
+		ret.Used = ret.Total - ret.Available
+	} else {
 		if activeFile && inactiveFile && sReclaimable {
 			ret.Available = calculateAvailVmem(ctx, ret, retEx)
 		} else {
 			ret.Available = ret.Cached + ret.Free
 		}
+		ret.Used = ret.Total - ret.Free - ret.Buffers - ret.Cached
 	}
```
rogercoll:
```suggestion
	if !memavail {
		if activeFile && inactiveFile && sReclaimable {
			ret.Available = calculateAvailVmem(ctx, ret, retEx)
		} else {
			ret.Available = ret.Cached + ret.Free
		}
	}
	// Opt-Out of calculating Mem.Used if the context has the context key set to true.
	// This is used for backward compatibility with applications that expect the old calculation method.
	// However, we plan to standardize on using MemAvailable in the future.
	// Therefore, please avoid using this opt-out unless it is absolutely necessary.
	// see https://github.com/shirou/gopsutil/issues/1873
	if val, ok := ctx.Value(WillBeDeleteOptOutMemAvailableCalc).(bool); ok && val {
		ret.Used = ret.Total - ret.Free - ret.Buffers - ret.Cached
	} else {
		ret.Used = ret.Total - ret.Available
	}
```

I think that only the `rect.Used` formula needs to change with the opt-out context.

shirou: Hmm, The original commit was intended to avoid the negative conditional expression (!memavail). Additionally, the opt-out commit was meant to create a completely separate conditional branch that could be cleanly removed later.

However, I acknowledge that the conditional logic has become a bit complex, as you pointed out. It seems better to follow your suggestion. Thank you.

mem/mem_linux.go
```diff
@@ -18,6 +18,13 @@ import (
 	"github.com/shirou/gopsutil/v4/internal/common"
 )
 
+// WillBeDeleteOptOutMemAvailableCalc is a context key to opt out of calculating Mem.Used.
+// This is not documented, and will be removed in Mar. 2026. This constant will be removed
+// in the future, but it is currently public. The reason is that making it public allows
+// developers to notice its removal when their build fails.
+// See https://github.com/shirou/gopsutil/issues/1873
+const WillBeDeleteOptOutMemAvailableCalc = "optOutMemAvailableCalc"
```
braydonk:
```suggestion
const WillBeDeletedOptOutMemAvailableCalc = "optOutMemAvailableCalc"
```

PR #1883, opened 2025-07-16
chore: enable unused-receiver from revive

mmorel-35:
#### Description

Enable [unused-receiver](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#unused-receiver) from revive.  It removes names for unused receivers.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -111,6 +111,7 @@ linters:
         - name: unnecessary-stmt
         - name: unreachable-code
         - name: unused-parameter
+        - name: unused-receiver
         - name: use-any
         - name: var-declaration
         - name: var-naming
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -49,7 +49,7 @@ func (i Invoke) Command(name string, arg ...string) ([]byte, error) {
 	return i.CommandWithContext(ctx, name, arg...)
 }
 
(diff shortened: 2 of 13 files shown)


disk/disk_freebsd.go
```diff
@@ -19,7 +19,7 @@ import (
 
 // PartitionsWithContext returns disk partition.
 // 'all' argument is ignored, see: https://github.com/giampaolo/psutil/issues/906
```
ccoVeille: I feel like a comment like this one should be provided for all functions you are updating.

Having -withcontext functions that ignore context sounds a bug.

I'm fine with the changes, I understand them. But it's end to something strange.

I feel like a comment like

- this function complies the interface/signature, unfortunately this os doesn't support context

ccoVeille: Another way to consider the change would be to check the context and return ctx.Error

ccoVeille: I opened 

- https://github.com/shirou/gopsutil/issues/1886

shirou:
Thanks for the PR. I agree that it's certainly possible to remove the variable name definitions. However, as mentioned in the comment, gopsutil needs to use the same function signatures across all platforms. Because of that, it's inevitable that some variables won't be used on certain platforms. I personally think it's better to explicitly use `_` to clearly indicate that the variable is unused on that platform.

PR #1885, opened 2025-07-19
chore: enable fatcontext linter

mmorel-35:
#### Description

Enable [fatcontext](https://golangci-lint.run/usage/linters/#fatcontext) .  It detects nested contexts in loops and function literals.


ccoVeille:
Did it detect something?

PR #1887, opened 2025-07-22
process: Fix parsing of uid/gid's to handle full uint32 range

halaney:
The following can be seen when doing a Process::Status():

    strconv.ParseInt: parsing "2976972800": value out of range

We're treating the uid/gid fields as if they're int32's when parsing, but they're uint32's. A large enough uid/gid can kick us over the edge.

Handle it as a uint32 and add a test around it.

This is similar to what
aace5e9 ("Groups in /proc/PID/status has type uint32.") did for Groups.


shirou (approved):
Good catch! Indeed, those should be `uint32`.

PR #1894, opened 2025-07-30
Remove common test helper to skip tests

rosstimothy:
The `internal/common/common_testing.go` file caused `testing` to leak into any binaries that depend directly or indirectly on this library. The file only contained a single helper function used to skip tests based on a particular error type. All tests that were making use of the helper have had the check inlined instead and the helper file has been removed.

Example dependency tree illustrating the `testing` leak:

```
goda tree "reach(github.com/gravitational/teleport/tool/tbot...:all, testing)"

  ├ github.com/gravitational/teleport/tool/tbot
    ├ github.com/gravitational/teleport/lib/tbot
      ├ github.com/gravitational/teleport/lib/tbot/config
        └ github.com/gravitational/teleport/lib/tbot/workloadidentity/workloadattest
          ├ github.com/gravitational/teleport/lib/tbot/workloadidentity/workloadattest/sigstore
          └ github.com/shirou/gopsutil/v4/process
            ├ github.com/shirou/gopsutil/v4/cpu
              └ github.com/shirou/gopsutil/v4/internal/common
            ├ github.com/shirou/gopsutil/v4/internal/common ~
            ├ github.com/shirou/gopsutil/v4/mem
              └ github.com/shirou/gopsutil/v4/internal/common ~
            └ github.com/shirou/gopsutil/v4/net
              └ github.com/shirou/gopsutil/v4/internal/common ~
```

### DIFF
--- a/mem/mem_plan9_test.go
+++ b/mem/mem_plan9_test.go
@@ -4,6 +4,7 @@
 package mem
 
 import (
+	"errors"
 	"reflect"
 	"testing"
 


shirou (changes requested):
Thank you for this PR. This is a good catch and a valuable improvement.

I can confirm that the `testing` package is indeed being pulled into the dependency tree for production builds, which shouldn't happen. Removing the test helper and inlining the error checks is a clean and straightforward solution.

While the inline checks are slightly more verbose (3 lines vs 1), this is an acceptable tradeoff - it's unfortunate to lose the conciseness, but keeping test dependencies out of production binaries is far more important. The inlined version is also more explicit and follows Go's idiomatic error handling patterns.

However, I see there are some files failing lint checks. Could you please fix those? Once the lint issues are resolved, I'd be happy to merge this.

PR #1895, opened 2025-08-04
Do not return an error on unknown CPU stepping

pgimalac:
As mentioned in https://github.com/DataDog/datadog-agent/issues/17935#issuecomment-3141818500, on some platforms the "stepping" entry in `/proc/cpuinfo` is "unknown", which makes `gopsutil` return an error.

I think this should just be ignored in that case, as if the `stepping` key was missing.

There is no test for this function so I didn't add any to cover this change, but if you want to I can add some.


shirou (approved):
LGTM! Thank you so much!

PR #1903, opened 2025-08-17
process: Cache dlOpen lib for faster lib calls

prashantv:
On darwin, calls that rely on purego functions such as process times or memory info will open/close the library, which dlOpen/dlCloses the same library repeatedly. This causes them to be significantly slower when they're used for periodic metrics emission.

Cache the library on the process to avoid the repeated dlOpen/dlClose.

The added benchmark shows significant improvements:
```
# old
BenchmarkProcessCPUTime-10        211495              5682 ns/op            2776 B/op         64 allocs/op
# new
BenchmarkProcessCPUTime-10       1159842              1021 ns/op             760 B/op         15 allocs/op
```

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -14,6 +14,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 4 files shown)


prashantv:
Can I get a review for this optimization, it's especially helpful for use-cases that get the CPU time periodically.

shirou (changes requested):
I see, indeed the initialization cost can be significant when it’s called frequently. The `Process` struct already caches some information to reduce overhead, so adding this as part of that caching aligns well with the existing implementation. Thank you.

However, I think the file name `process_state_none.go` is a bit unclear in terms of what "state" means. Could you consider renaming it to something like `process_osprocess_fallback.go` instead?

process/process_darwin.go
```diff
@@ -309,16 +324,11 @@ func (f *dlFuncs) getTimeScaleToNanoSeconds() float64 {
...
-func (f *dlFuncs) Close() {
-	f.lib.Close()
-}
-
 func (p *Process) ExeWithContext(_ context.Context) (string, error) {
-	funcs, err := loadProcFuncs()
+	funcs, err := p.os.funcs()
 	if err != nil {
-		return "", err
+		return "", nil
```
shirou, line 330: Sorry for the late comment. I think we should return `err` when `p.os.funcs()` returns an error. Was there a specific reason you decided to return `nil` instead?

process/process_darwin.go
```diff
@@ -343,11 +353,10 @@ type vnodePathInfo struct {
 // error.
 // Note: This might also work for other *BSD OSs.
 func (p *Process) CwdWithContext(_ context.Context) (string, error) {
-	funcs, err := loadProcFuncs()
+	funcs, err := p.os.funcs()
 	if err != nil {
-		return "", err
+		return "", nil
```
shirou, line 358: ditto

PR #1904, opened 2025-08-23
Remove wmi from cpu package

### LINKED ISSUE #394: Reading spend too much time

OrbintSoft:
With this change I completely removed the slow WMI queries, and replaced them with windows API/ windows registry and SMBIOS

The code should be 100% compatible with previous implementation, but if someone has a multiarcheture system , it would be nice to test it.

Closes #394

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -7,11 +7,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/bits"
+	"path/filepath"
 	"strconv"
+	"syscall"
 	"unsafe"
 
-	"github.com/yusufpapurcu/wmi"
 	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
 
 	"github.com/shirou/gopsutil/v4/internal/common"
 )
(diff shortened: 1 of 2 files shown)


cpu/cpu_windows.go
```diff
@@ -55,8 +49,20 @@ const (
 
 	// size of systemProcessorPerformanceInfoSize in memory
 	win32_SystemProcessorPerformanceInfoSize = uint32(unsafe.Sizeof(win32_SystemProcessorPerformanceInformation{})) //nolint:revive //FIXME
+	// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ne-wdm-power_information_level
+	processorInformation = 11
+)
+
+type Relationship uint32
```
Lomanic, line 56: this should not be exported

OrbintSoft:
yes, sure. I will split the PR. 
For the VendorId, I will try to check if there are other ways to get it. 
The original source should be cpuid, but cpuid gives the value only on the processor where the current thread is running.
Now on x86 architecture, it's not a problem, you can have all AMD, or all Intel or other minor licensed vendors (VIA technologies), but a mixed architecture is not possible,

On Arm, theoretically you can have an heterogeneous architecture (Linux supports it), but windows supports only Symmetric multiprocessing, so all processors should be of same model (same cores, same instruction set, same family), so we can be confident and get ProcessorFamily and ProcessorID from SMBIOS (that should be the source where also windows is retriving the information.
But for the vendor I am not sure, there is nothing that forbids 2 vendors to make two 100% compatible processors, and if they are of the same family (same cores, instruction set, etc), you can possibility have a windows ARM  machine with different cpu of different vendors.
I can try to investigate if I can get the value with https://learn.microsoft.com/en-us/windows/win32/api/setupapi/nf-setupapi-setupdigetdevicepropertyw, but I need some time to check.

@shirou can give a look at you at your registry to see what values you have on HKEY_LOCAL_MACHINE\HARDWARE\DESCRIPTION\System\CentralProcessor ?

thank you.

OrbintSoft:
mm, thank you I will look into it.

cpu/cpu_windows.go
OrbintSoft: this PR will be closed, look at https://github.com/shirou/gopsutil/pull/1915

OrbintSoft: I am splitting the PR as requested by @shirou

PR #1906, opened 2025-08-26
refactor: use slices.Contains to simplify code

wmypku:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


shirou (commented):
Thank you for your PR. I noticed that there are other functions, like `StringsContains`, that could be simplified in a similar way. Would you mind updating those as well?

wmypku:
> Thank you for your PR. I noticed that there are other functions, like `StringsContains`, that could be simplified in a similar way. Would you mind updating those as well?

Of course. Modified. Please review again.

shirou (approved):
LGTM! Thank you so much!

PR #1910, opened 2025-08-28
Fix: use compatible perfstat version to fix aix cgo builds

gremat:
The initial version of cpu_aix.go (now split into cpu_aix_cgo.go) already used perfstat.CpuUtilTotalStat() which was not available in the pinned perfstat version in go.mod (then and now) but only in a more recent commit on perfstat (actually, the very next one, compare [1]).
This commit bumps perfstat to the latest state, picking up f38a28bc for the required CpuUtilTotalStat() and also a couple of more recent changes like Power10 support.

[1] https://github.com/power-devops/perfstat/commits/main/


shirou (approved):
I haven’t checked whether this update works or not, but it seems to be just a library update. So, LGTM! Thank you for your contribution!

Issue #1911, opened 2025-09-03
Leak in PartitionsWithContext


StefanoBalzarottiNozomi:
**Describe the bug**
If you mount a directory as volume on a long path, the code gets stuck in an finite loop, causing also a memory leak by appending error to the list of warning.

**To Reproduce**
```go
for {            
    mounts, err := getVolumePaths(volNameBuf)
	// Here the code returns an error, and it will never succeed because size of volNameBuf is not enough
	if err != nil {
		warnings.Add(fmt.Errorf("failed to find paths for volume %s", windows.UTF16ToString(volNameBuf))) // array append
		continue // back to where we started
	}
```

**Expected behavior**
The code does't keak

**Environment (please complete the following information):**
 - [X] Windows: [`Windows 11 Enterprise`]

PR #1534, opened 2023-10-03
Fetch stats from folder mounted volumes

dloucasfx:
This fixes https://github.com/shirou/gopsutil/issues/1245 and https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27203

### Steps to reproduce
- Add a new disk to a windows VM
- from Disk Management tool, right-click the new volume, and select Change drive letter and paths.
- From there, add “Mount in the following empty NTFS folder” and select an empty dir on the C:\ drive and make sure there is NO assigned drive letter

### Before the fix: 
```
Resource SchemaURL: https://opentelemetry.io/schemas/1.9.0
Resource attributes:
     -> host.name: Str(wintestswatsplunk)
     -> os.type: Str(windows)
     -> cloud.provider: Str(azure)
     -> cloud.platform: Str(azure_vm)
     -> cloud.region: Str(eastus)
     -> host.id: Str(09f0d80d-d5a7-4082-953a-bfd10cb77cbf)
     -> cloud.account.id: Str(5db89e2c-2127-4418-9a31-dac67b04e87b)
     -> azure.vm.name: Str(win2016-swat-2021)
     -> azure.vm.size: Str(Standard_D2as_v4)
     -> azure.vm.scaleset.name: Str()
     -> azure.resourcegroup.name: Str(win2016-swat)
ScopeMetrics #0
ScopeMetrics SchemaURL:
InstrumentationScope otelcol/hostmetricsreceiver/filesystem 0.85.0-dev
Metric #0
Descriptor:
     -> Name: system.filesystem.usage
     -> Description: Filesystem bytes used.
     -> Unit: By
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: Cumulative
NumberDataPoints #0
Data point attributes:
     -> device: Str(C:)
     -> mode: Str(rw)
     -> mountpoint: Str(C:)
     -> type: Str(NTFS)
     -> state: Str(used)
StartTimestamp: 2023-09-26 14:11:00 +0000 UTC
Timestamp: 2023-10-03 21:22:58.5265249 +0000 UTC
Value: 19619004416
NumberDataPoints #1
Data point attributes:
     -> device: Str(C:)
     -> mode: Str(rw)
     -> mountpoint: Str(C:)
     -> type: Str(NTFS)
     -> state: Str(free)
StartTimestamp: 2023-09-26 14:11:00 +0000 UTC
Timestamp: 2023-10-03 21:22:58.5265249 +0000 UTC
[...]

PR #1914, opened 2025-09-05
CRITICAL: fix goroutine leak infinte loop

StefanoBalzarottiNozomi:
- Limited warning list to avoid memory leak
- Removed leaking goroutine and implemented context correctly, the code is stopped as soon as we can, the syscall is not interruptible, leaving a leaking goroutine is dangerous if code is called multiple times.
- Implemented windows API correctly, max paths are now set correctly and we require the buffer length when necessary.
- If we receive an unexpected error while iterating volumes, I return, this fixes the infinite loop.

- Added a test to check with volumes mounted as folder, the test is skipped because requires to mount a volume, so you need to run it manually, the code previously was not tested at all.

This fixes the issue that I reported here: https://github.com/shirou/gopsutil/issues/1911

Please, review this PR quickly, if you see problems at least revert this PR: https://github.com/shirou/gopsutil/pull/1534

The code was unsafe even before with that leaking goroutine, but the change introduced by that PR, is extremely dangerous it can cause BSODs on windows.

I hope you will accept my fix, thank you.

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -34,6 +34,7 @@ var (
 	procGetVolumePathNamesForVolumeNameW = common.Modkernel32.NewProc("GetVolumePathNamesForVolumeNameW")
 )
 
+// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getvolumeinformationa#parameters
 const (
 	rw       = "rw"
 	ro       = "ro"


shirou (changes requested):
This PR appears to include several unrelated changes, which makes it difficult to review. Could you separate them into individual PRs, each addressing a single issue ?

StefanoBalzarottiNozomi:
Sure I can, but maybe is better if first you revert this PR https://github.com/shirou/gopsutil/pull/1534.

PR #1915, opened 2025-09-07
Cores count without wmi

OrbintSoft:
@shirou As discussed here I made a PR just to get the cores count without wmi:
https://github.com/shirou/gopsutil/pull/1904

The code is a bit redundant, and I kept some extra safety checks, because I had to split the PR, once we will  completely merge everything, we will have just one source of information and everything will be more clean.

In next PR I will implement the call to get Mhz with procCallNtPowerInformation.

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/bits"
 	"strconv"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


OrbintSoft:
it seems that the CI is broken. @shirou  can you look into it?

cpu/cpu_windows.go
```diff
@@ -57,6 +58,20 @@ const (
 	win32_SystemProcessorPerformanceInfoSize = uint32(unsafe.Sizeof(win32_SystemProcessorPerformanceInformation{})) //nolint:revive //FIXME
 )
 
+type Relationship uint32
```
shirou: Could you change this variable to private?

cpu/cpu_windows.go
```diff
@@ -101,6 +116,15 @@ func Info() ([]InfoStat, error) {
 	return InfoWithContext(context.Background())
 }
 
+func forEachSetBit64(mask uint64, fn func(bit int)) {
```
shirou, line 125: Could you add some comment for this helper function?

shirou (changes requested):
Thank you for separating the PR. This makes it easier to understand, and it looks good to me. I also confirmed that the output is the same between the current implementation and this PR on my Windows 11 environment.

I left a few nit comments, but I’ll merge once those are addressed.

shirou (approved):
Thank you for the updates! LGTM and Let's merge !

PR #1919, opened 2025-09-14
get max Mhz per package without wmi

OrbintSoft:
continuing https://github.com/shirou/gopsutil/pull/1904

In next PR I will implement the SMBIOS call to get family and processorId.

I tested on the windows machines that I have at home, but if someone has a server with multiple packages / groups it would be very useful.

@Lomanic

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -20,6 +20,7 @@ import (
 var (
 	procGetNativeSystemInfo              = common.Modkernel32.NewProc("GetNativeSystemInfo")
 	procGetLogicalProcessorInformationEx = common.Modkernel32.NewProc("GetLogicalProcessorInformationEx")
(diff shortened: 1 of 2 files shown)


cpu/cpu_windows.go
```diff
@@ -171,20 +206,32 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
...
+				if globalLpl >= 0 && globalLpl < len(ppis) {
 					logicalCount++
+					m := int(ppis[globalLpl].maxMhz)
+					if m > maxMhz {
+						maxMhz = m
+					}
 				}
 			})
-
-			ret[i].Cores = int32(logicalCount)
```
OrbintSoft, line 186: this was a bug, cores count should be per package, not per group

PR #1920, opened 2025-09-15
limit memory usage on warnings list

StefanoBalzarottiNozomi:
-  Warning List has no bound limit, there are cases where the list accumulates an infinite amount of errors causing memory exhaustion on the machine.

- I think having more than 100 errors, is not helpful in troubleshooting, it can only cause system instability.

- I propose to limit the amount, this is the most simple solution, the if you think it's necessary a more complex logic can be added, ex: (making configurable, avoid duplicated errors, add error rotations, ..)

### DIFF
--- a/internal/common/warnings.go
+++ b/internal/common/warnings.go
@@ -3,12 +3,17 @@ package common
 
 import "fmt"
 
+const maxWarnings = 100 // An arbitrary limit to avoid excessive memory usage, it has no sense to store thousands of errors
+
 type Warnings struct {
 	List    []error
 	Verbose bool
 }
 
 func (w *Warnings) Add(err error) {
+	if len(w.List) >= maxWarnings {
+		return
+	}
 	w.List = append(w.List, err)
 }
 


StefanoBalzarottiNozomi:
Reference: https://github.com/shirou/gopsutil/issues/1911

internal/common/warnings.go
```diff
@@ -3,12 +3,17 @@ package common
...
 	List    []error
 	Verbose bool
 }
 
 func (w *Warnings) Add(err error) {
+	if len(w.List) >= maxWarnings {
+		return
+	}
 	w.List = append(w.List, err)
 }
```
ccoVeille: It's a good idea, but then I would prefer it's not a silent behavior.


```suggestion
const maxWarnings = 100 // An arbitrary limit to avoid excessive memory usage, it has no sense to store thousands of errors

 type Warnings struct {
 	List    []error
	TooManyErrors bool
 	Verbose bool
 }
 
 func (w *Warnings) Add(err error) {
	if len(w.List) >= maxWarnings {
		w.TooManyErrors = true
		return
	}
 	w.List = append(w.List, err)
 }
```

Or something like this

```suggestion
const maxWarnings = 100 // An arbitrary limit to avoid excessive memory usage, it has no sense to store thousands of errors

 type Warnings struct {
 	List    []error
	tooManyErrors bool
 	Verbose bool
 }
 
var ErrTooManyErrors = errors.New("too many errors reported, next ones were discarded") // a better wording could be suggested here

 func (w *Warnings) Add(err error) {
	if len(w.List) >= maxWarnings {
		if w.tooManyErrors {
			return
		}
		w.tooManyErrors = true
		err = ErrTooManyErrors
	}
 	w.List = append(w.List, err)
 }
```

I prefer the second one.

What do you guys think about this?

StefanoBalzarottiNozomi: it's ok to me, lets wait @shirou opinion

shirou: 👍 since `go build` shows the same behavior when there are many build errors.
However, I don’t think we need to define `ErrTooManyErrors` as public, because the behavior doesn’t really need to change depending on whether there are too many errors or not.

You may have already noticed, but I’d also suggest adding a notice in the `Error()` output, for example:

```
if w.tooManyerrors {
    return fmt.Sprintf("too many errors reported.")
}
```

internal/common/warnings.go
```diff
@@ -3,12 +3,20 @@ package common
 
 import "fmt"
 
+const maxWarnings = 100 // An arbitrary limit to avoid excessive memory usage, it has no sense to store thousands of errors
+const tooManyErrorsMessage = "too many errors reported."
```
ccoVeille: Maybe this


```suggestion
const tooManyErrorsMessage = "too many errors reported, next errors were discarded"
```

internal/common/warnings.go
```diff
@@ -25,7 +33,13 @@ func (w *Warnings) Error() string {
...
+		if w.tooManyErrors {
+			str += fmt.Sprintf("\t%s\n", tooManyErrorsMessage)
+		}
 		return str
 	}
+	if w.tooManyErrors {
+		return tooManyErrorsMessage
+	}
 	return fmt.Sprintf("Number of warnings: %v", len(w.List))
 }
```
ccoVeille, line 48: Maybe this would be clearer

```suggestion
	if w.tooManyErrors {
		return return fmt.Sprintf("Number of warnings: > %v - %s", len(w.List), tooManyErrorsMessage)
	}
	return fmt.Sprintf("Number of warnings: %v", len(w.List))
}
```

internal/common/warnings.go
```diff
@@ -3,12 +3,20 @@ package common
 
 import "fmt"
 
+const maxWarnings = 100 // An arbitrary limit to avoid excessive memory usage, it has no sense to store thousands of errors
```
ccoVeille: What about this

```suggestion
const maxWarnings = 100 // An arbitrary limit to avoid excessive memory usage, it has no sense to store hundreds of errors
```

ccoVeille (approved):

PR #1922, opened 2025-09-15
fix infinite loop

StefanoBalzarottiNozomi:
The code to get the partitions mounted as folders is wrong and can fail for many reasons, here I just block the loop at first error to avoid and infinite loop.

I will fix properly once all PR are merged.

Reference:

https://github.com/shirou/gopsutil/issues/1911
https://github.com/shirou/gopsutil/pull/1914

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -22,6 +22,7 @@ import (
 const (
 	volumeNameBufferLength = uint32(windows.MAX_PATH + 1)
 	volumePathBufferLength = volumeNameBufferLength
+	maxWarningsInDrive     = 5
 )
 
 var (
@@ -172,7 +173,10 @@ func PartitionsWithContext(ctx context.Context, _ bool) ([]PartitionStat, error)
 					break
 				}
 				warnings.Add(fmt.Errorf("failed to find next volume: %w", err))
-				break
+				if len(warnings.List) > maxWarningsInDrive {
+					break
+				}
+
 			}
 		}
 	}


disk/disk_windows.go
```diff
@@ -172,6 +172,7 @@ func PartitionsWithContext(ctx context.Context, _ bool) ([]PartitionStat, error)
 					break
 				}
 				warnings.Add(fmt.Errorf("failed to find next volume: %w", err))
+				break
```
StefanoBalzarottiNozomi: if `procFindNextVolumeW` returns an error different than ERROR_NO_MORE_FILES, it will continue to retry indefinitely.

For example if the error is ERROR_INSUFFICIENT_BUFFER, the call will never succeed, and function will stay blocked forever in an infinite loop.

shirou (commented):
I think using just break here is a bit wasteful. How about define `maxWarningsInDrive` (for example, 5), checking the length of `warnings.List`, and then break the loop if exceeded?

StefanoBalzarottiNozomi:
Ok I will do, but I think the correct solution is to check the error, there are some errors like network not reachable (not sure what is the exact error code) that could be retried, and some errors like ERROR_INSUFFICIENT_BUFFER, INVALID_PARAMETERS, or whatever that are useless to retry, they will always fail.

Anyway I think the whole code must be revised, the way the function is called, the error is checked and the context cancellation is wrongly implemented.

I put that break just as quick patch fix.

PR #1929, opened 2025-09-28
Use registry to remove wmi from cpu package

OrbintSoft:
Implemented registry to get model name and vendorId.
Removed wmi from cpu package

### DIFF
--- a/cpu/cpu_test.go
+++ b/cpu/cpu_test.go
@@ -104,7 +104,7 @@ func TestInfo(t *testing.T) {
 		assert.NotEmptyf(t, vv.ModelName, "could not get CPU Info: %v", vv)
 		if runtime.GOOS == "windows" {
 			assert.NotEmptyf(t, vv.VendorID, "could not get Vendor ID: %v", vv)
-			assert.NotEmptyf(t, vv.VendorID, "could not get CPU Family: %v", vv)
+			assert.NotEmptyf(t, vv.Family, "could not get CPU Family: %v", vv)
 			assert.Positivef(t, vv.Mhz, "could not get CPU Mhz: %v", vv)
 			assert.Positivef(t, vv.Cores, "could not get CPU Cores: %v", vv)
 		}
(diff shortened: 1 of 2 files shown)


cpu/cpu_test.go
```diff
@@ -102,6 +102,12 @@ func TestInfo(t *testing.T) {
 	assert.NotEmptyf(t, v, "could not get CPU Info")
 	for _, vv := range v {
 		assert.NotEmptyf(t, vv.ModelName, "could not get CPU Info: %v", vv)
+		if runtime.GOOS == "windows" {
+			assert.NotEmptyf(t, vv.VendorID, "could not get Vendor ID: %v", vv)
+			assert.NotEmptyf(t, vv.VendorID, "could not get CPU Family: %v", vv)
```
shirou:
```suggestion
			assert.NotEmptyf(t, vv.Family, "could not get CPU Family: %v", vv)
```

cpu/cpu_windows.go
```diff
@@ -246,12 +216,26 @@ func InfoWithContext(ctx context.Context) ([]InfoStat, error) {
 						maxMhz = m
 					}
 				}
+
+				registryKeyPath := filepath.Join(centralProcessorRegistryKey, strconv.Itoa(globalLpl))
+				key, err := registry.OpenKey(registry.LOCAL_MACHINE, registryKeyPath, registry.QUERY_VALUE|registry.READ)
+				if err == nil {
+					defer key.Close()
+					model = getRegistryStringValueIfUnset(key, "ProcessorNameString", model)
+					vendorId = getRegistryStringValueIfUnset(key, "VendorIdentifier", vendorId)
```
shirou: `defer` will not execute until this function returns, but I think we don’t need to wait for that.
```suggestion
					model = getRegistryStringValueIfUnset(key, "ProcessorNameString", model)
					vendorId = getRegistryStringValueIfUnset(key, "VendorIdentifier", vendorId)
					key.Close()
```

PR #1930, opened 2025-09-29
chore: enable noctx linter

mmorel-35:
#### Description

Enables and fixes issues from [noctx](https://golangci-lint.run/docs/linters/configuration/#noctx) linter

### DIFF
--- a/cpu/cpu_linux_test.go
+++ b/cpu/cpu_linux_test.go
@@ -2,6 +2,7 @@
 package cpu
 
 import (
+	"context"
 	"errors"
 	"os/exec"
 	"strconv"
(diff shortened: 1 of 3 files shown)


cpu/cpu_linux_test.go
```diff
@@ -41,7 +42,7 @@ func TestParseStatLine_424(t *testing.T) {
 }
 
 func TestCountsAgainstLscpu(t *testing.T) {
```
StefanoBalzarottiNozomi, line 44: you can use t.Context

mmorel-35: Yes but only when the module is in go 1.24 or greater

internal/common/common_freebsd.go
```diff
@@ -29,7 +30,7 @@ func SysctlUint(mib string) (uint64, error) {
 }
 
 func DoSysctrl(mib string) ([]string, error) {
-	cmd := exec.Command("sysctl", "-n", mib)
+	cmd := exec.CommandContext(context.Background(), "sysctl", "-n", mib)
```
shirou: I noticed that `DoSysctrl` doesn’t seem to be used anymore, so we can remove those functions. So I opened #1936.

PR #1931, opened 2025-10-07
Fixes #1284

### LINKED ISSUE #1284: invalid mountinfo line in file /proc/self/mountinfo with two hyphens
On my system, telegraf in unable to collect disk usage info. When executing telegraf with --test, I get this error message:

```
 [inputs.disk] Error in plugin: error getting disk usage info: found invalid mountinfo line in file /proc/self/mountinfo: 22 13 0:19 / /dev/shm rw,nosuid,nodev,noexec,relatime - tmpfs - rw 
```

[caused by disk/disk_linux.go](https://github.com/shirou/gopsutil/blob/cf6309316c89a96a0a4b848185b32830a6e8175e/disk/disk_linux.go#L311-L315): 

```
// split the mountinfo line by the separator hyphen
parts := strings.Split(line, " - ")'
if len(parts) != 2 {
    return nil, fmt.Errorf("found invalid mountinfo line in file %s: %s ", filename, line)
}
```

Obviously, line shouldn't be split at each '-', but only at the first '-'.

johnnybubonic:
I know this issue was closed, but it's literally a 2 second fix to bring it in line with proper documented parsing of the mountinfo format.

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -292,7 +292,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 	}
 
 	// use mountinfo
-	ret, err = parseFieldsOnMountinfo(ctx, lines, all, fs, filename)
+	ret, err = parseFieldsOnMountinfo(ctx, lines, all, filename)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing mountinfo file %s: %w", filename, err)
 	}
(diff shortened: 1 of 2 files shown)


disk/disk_linux.go
```diff
@@ -332,7 +332,7 @@ func parseFieldsOnMountinfo(ctx context.Context, lines []string, all bool, fs []
 		// (1) (2) (3)   (4)   (5)      (6)      (7)   (8) (9)   (10)         (11)
 
 		// split the mountinfo line by the separator hyphen
-		parts := strings.Split(line, " - ")
+		parts := strings.SplitN(line, " - ", 2)
```
StefanoBalzarottiNozomi, line 340: ok, but can you add a test?

johnnybubonic: To specifically account for the condition in #1284 ? Existing passes should work fine with coverage, but I can probably add a test for success on #1284 condition, sure. Give me a few.

johnnybubonic: This is done, but turns out the existing test (`disk/disk_linux_test.go:Test_parseFieldsOnMountinfo`) itself passes on mounts it should not - per [docs for `all`](Test_parseFieldsOnMountinfo):

> If all is false, returns physical devices only (e.g. hard disks, cd-rom drives, USB keys) and ignore all others (e.g. memory partitions such as /dev/shm)

This is caused by the erroneous assumption in `disk/disk_linux.go` on [line 316](https://github.com/shirou/gopsutil/blob/5aa9d1fe1259b81db1498c87866aa6da0789bf33/disk/disk_linux.go#L316):

```go
			if d.Device == "none" || !common.StringsHas(fs, d.Fstype) {
				continue
			}
```

However, per [`fstab(5)`](https://www.mankier.com/5/fstab#Description-The_first_field_(fs_spec).):

> For filesystems **with no storage**, **_any string_** can be used, and will show up in [df(1)](https://www.mankier.com/1/df) output, for example. Typical usage is proc for procfs; mem, none, or tmpfs for tmpfs. Other special filesystems, like udev and sysfs, are typically not listed in fstab.

_(emphasis added)_

Line 316 only assumes "none", however.

This leads to the `sysfs` test case (and thus THIS condition for #1284) *improperly* being included with `all == false` in `Test_parseFieldsOnMountinfo`.

This is, obviously, incorrect - but it'd be a result-breaking fix, technically, as e.g. sysfs would no longer be included IN `all == false`.

Ironically enough, I think *that* will be fixed by #1933 but I'll doublecheck.

johnnybubonic: alright, new changes should reflect accurate logic now, and properly handles several cases that were being improperly handled before as well.

StefanoBalzarottiNozomi: very nice, thanks for the fix

disk/disk_linux.go
```diff
@@ -325,44 +325,63 @@ func parseFieldsOnMounts(lines []string, all bool, fs []string) []PartitionStat
...
 
 		d := PartitionStat{
 			Device:     device,
 			Mountpoint: unescapeFstab(mountPoint),
-			Fstype:     fstype,
+			Fstype:     fsType,
 			Opts:       mountOpts,
 		}
 
-		if !all {
```
shirou, line 360: As golangci-lint says `unused-parameter: parameter 'fs' seems to be unused, consider removing or renaming it as _ (revive)`, it seems this `fs` filter has been removed. Could you recover?

disk/disk_linux.go
```diff
@@ -325,44 +325,63 @@ func parseFieldsOnMounts(lines []string, all bool, fs []string) []PartitionStat
...
 		if len(parts) != 2 {
 			return nil, fmt.Errorf("found invalid mountinfo line in file %s: %s ", filename, line)
 		}
 
 		fields := strings.Fields(parts[0])
 		blockDeviceID := fields[2]
+		rootDir := fields[3]
 		mountPoint := fields[4]
 		mountOpts := strings.Split(fields[5], ",")
+		fields = strings.Fields(parts[1])
```
shirou, line 356: Could you add length check?
```suggestion
		fields = strings.Fields(parts[1])
		if len(fields) < 6 {
			return nil, fmt.Errorf("invalid mountinfo line: %s", line)
		}

```

johnnybubonic:
I completely dropped this; my apologies! Things got busy at $dayjob. I'll try to have this addressed shortly.

disk/disk_linux.go
johnnybubonic: Yep, fixed. I removed it from the function signature itself since it's no longer necessary with the new logic and `parseFieldsOnMountinfo` is non-exported. I had Goland check for other usages; aside from the unit test, this function is only called from `PartitionsWithContext` so impact is minimal.

disk/disk_linux.go
johnnybubonic: Added length checks for both `fields` splits (the one highlighted here and the first split as well)

PR #1934, opened 2025-10-07
Fix potential edge case on procfs and alike

johnnybubonic:
On systems that use procfs or equivalent (Linux, Solaris), parsing the PID directory name (e.g. `/proc/<pid>`) is done strictly with `strconv.ParseInt`.

Per https://pkg.go.dev/strconv#ParseInt :

>The string may begin with a leading sign: "+" or "-".

Thus these cases would be interpreted erronously as PIDs:

* `/proc/+253/`
* `/proc/-253/`
  * This would even be interpreted as PID -253, since you use ParseInt and not ParseUint

This PR prevents this from occurring (and can avoid issues in the future were you not using an explicit base 10 -- e.g.:
`strconv.ParseInt(fname, 0, 32)` -- and it erroneously parsing e.g. `/proc/0xfd/`).

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"regexp"
 	"runtime"
 	"sort"
 	"sync"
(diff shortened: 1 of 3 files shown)


johnnybubonic:
> However, do you know if there’s any platform that adds `+` or `-` under /`procfs`? If so, could you please provide some documentation?

Not any of the supported plats on this library currently, no (unless some distro out there has some patched procfs; I doubt they would but Debian and Ubuntu are notorious for adding downstream "feature" patches that tend to break expected conventions), but this PR should accommodate and avoid issues that would cause if so.

PR #1937, opened 2025-10-11
use correct path lengths

StefanoBalzarottiNozomi:
Fix to use correct paths/buffer lengths, consider that NTFS supports also paths longer than 256 (windows.MAX_PATH)

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -200,7 +200,7 @@ func buildPartitionStat(path string) (PartitionStat, error) {
 			uintptr(unsafe.Pointer(volPath)),
 			uintptr(unsafe.Pointer(&volumeName[0])),
 			uintptr(len(volumeName)),
-			uintptr(0), // serial numbrt
+			uintptr(0), // serial number
 			uintptr(0), // max component length
 			uintptr(unsafe.Pointer(&fsFlags)),
 			uintptr(unsafe.Pointer(&fsName[0])),


disk/disk_windows.go
```diff
@@ -19,9 +20,9 @@ import (
 )
 
 const (
-	volumeNameBufferLength = uint32(windows.MAX_PATH + 1)
-	volumePathBufferLength = volumeNameBufferLength
-	maxWarningsInDrive     = 5
+	maxVolumeNameLength     = uint32(windows.MAX_PATH + 1) // this should be a GUID (50), but for safety I keep max_path
```
StefanoBalzarottiNozomi, line 23: it's a guid in string format that represent the volume, as far as I know all volumes are in guid format, but it's not well specified in documentation, so I keep MAX_PATH in case a volume can have a name in a longer format.

disk/disk_windows.go
```diff
@@ -101,15 +102,25 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
...
+
+	lpBuffer := make([]uint16, bufferLen)
+	// buffer can be longer than MAX_PATH
 	if diskret, _, err := procGetLogicalDriveStringsW.Call(
 		uintptr(len(lpBuffer)),
 		uintptr(unsafe.Pointer(&lpBuffer[0]))); diskret == 0 {
 		return partitionStats, err
 	}
+	drivesString := windows.UTF16ToString(lpBuffer)
+	drives := strings.Split(drivesString, "\x00")
```
StefanoBalzarottiNozomi, line 123: all drive names are separated by an UTF-16 null character

disk/disk_windows.go
```diff
@@ -101,15 +102,25 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
...
 
 	// Get drives with drive letters (including remote drives, ex: SMB shares)
-	lpBuffer := make([]byte, 254)
+	// We first call GetLogicalDriveStringsW with a buffer length of 0 to get the required buffer size.
+	// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlogicaldrivestringsw
+	bufferLen, _, err := procGetLogicalDriveStringsW.Call(
+		uintptr(0),
+		uintptr(0))
+
+	lpBuffer := make([]uint16, bufferLen)
```
StefanoBalzarottiNozomi, line 113: this fixes the call when you have many drives mounted and a buffer > windows.MAX_PATH

disk/disk_windows.go
```diff
@@ -176,16 +187,17 @@ func buildPartitionStat(path string) (PartitionStat, error) {
...
+		var fsFlags int64
 
+		// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getvolumeinformationw
 		ret, _, err := procGetVolumeInformation.Call(
 			uintptr(unsafe.Pointer(volPath)),
 			uintptr(unsafe.Pointer(&volumeName[0])),
 			uintptr(len(volumeName)),
-			uintptr(unsafe.Pointer(&serialNumber)),
-			uintptr(unsafe.Pointer(&maxComponentLength)),
+			uintptr(0), // serial numbrt
```
StefanoBalzarottiNozomi: these data are not used, so it's better to don't request them

disk/disk_windows.go
```diff
@@ -279,16 +291,28 @@ func LabelWithContext(_ context.Context, _ string) (string, error) {
...
+	firstResult, _, volumePathFirstErr := procGetVolumePathNamesForVolumeNameW.Call(
+		uintptr(unsafe.Pointer(&volNameBuf[0])),
+		uintptr(0),
+		uintptr(0),
+		uintptr(unsafe.Pointer(&returnLen)))
+	if firstResult == 0 && volumePathFirstErr != windows.ERROR_MORE_DATA {
+		return nil, fmt.Errorf("failed to get volume paths size for volume %s: %w", windows.UTF16ToString(volNameBuf), volumePathFirstErr)
+	}
+
+	volPathsBuf := make([]uint16, returnLen)
```
StefanoBalzarottiNozomi, line 310: this fixes the call when a volume is mounted as a folder in a path longer than 256 characters

disk/disk_windows.go
```diff
@@ -176,16 +191,17 @@ func buildPartitionStat(path string) (PartitionStat, error) {
...
+		var fsFlags int64
 
+		// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getvolumeinformationw
 		ret, _, err := procGetVolumeInformation.Call(
 			uintptr(unsafe.Pointer(volPath)),
 			uintptr(unsafe.Pointer(&volumeName[0])),
 			uintptr(len(volumeName)),
-			uintptr(unsafe.Pointer(&serialNumber)),
-			uintptr(unsafe.Pointer(&maxComponentLength)),
+			uintptr(0), // serial numbrt
```
shirou: typo?
```suggestion
			uintptr(0), // serial number
```

disk/disk_windows.go
```diff
@@ -101,15 +102,29 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
...
+	lpBuffer := make([]uint16, bufferLen)
+	// buffer can be longer than MAX_PATH
+	_, _, err = procGetLogicalDriveStringsW.Call(
 		uintptr(len(lpBuffer)),
-		uintptr(unsafe.Pointer(&lpBuffer[0]))); diskret == 0 {
-		return partitionStats, err
+		uintptr(unsafe.Pointer(&lpBuffer[0])))
+	if !errors.Is(err, windows.ERROR_SUCCESS) {
+		return partitionStats, err // The call failed with an unexpected error
 	}
```
shirou, line 120: As [the windows package documentation](https://pkg.go.dev/golang.org/x/sys/windows) "These calls return err == nil to indicate success." So I think we don’t need to check whether the error is `windows.ERROR_SUCCESS`; just check `err != nil`, as per the usual Go idiom.

shirou (changes requested):
Thank you for the PR (and sorry to late response). I made some comments, but overall, it seems very great PR!

disk/disk_windows.go
StefanoBalzarottiNozomi: image

as I see `err != nil`

StefanoBalzarottiNozomi: I think Go documentation is wrong, or the windows package is not implemented correctly according to the documentation

StefanoBalzarottiNozomi:
@shirou do you think we can merge before  november release ?

PR #1941, opened 2025-10-29
refactor disk windows and add tests

StefanoBalzarottiNozomi:
refactored a bit the code and added tests

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -115,8 +115,8 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
 
 	// Get drives with drive letters (including remote drives, ex: SMB shares)
 	drives, err := getLogicalDrives()
-	if logicalDrivesErr != nil {
-		return partitionStats, logicalDrivesErr
+	if err != nil {
+		return partitionStats, err
 	}
 
 	partitionStats = processLogicalDrives(drives, processedPaths, partitionStats, warnings)
@@ -132,7 +132,7 @@ func processVolumesMountedAsFolders(partitionStats []PartitionStat, warnings War
 		uintptr(unsafe.Pointer(&volNameBuf[0])),
 		uintptr(maxVolumeNameLength))
 	if windows.Handle(nextVolHandle) == windows.InvalidHandle {
-		warnings.Add(fmt.Errorf("failed to get first-volume: %w", logicalDrivesErr))
+		warnings.Add(fmt.Errorf("failed to get first-volume: %w", err))
 		return partitionStats
 	}
 	defer procFindVolumeClose.Call(nextVolHandle)


disk/disk_windows.go
```diff
@@ -102,46 +114,26 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
...
 
 	// Get volumes without drive letters (ex: mounted folders with no drive letter)
+	partitionStats = processVolumesMountedAsFolders(partitionStats, warnings, processedPaths)
+	return partitionStats, warnings.Reference()
+}
+
+func processVolumesMountedAsFolders(partitionStats []PartitionStat, warnings Warnings, processedPaths map[string]struct{}) []PartitionStat {
 	volNameBuf := make([]uint16, maxVolumeNameLength)
-	nextVolHandle, _, err := procFindFirstVolumeW.Call(
+	nextVolHandle, _, logicalDrivesErr := procFindFirstVolumeW.Call(
```
shirou: ditto

```suggestion
	nextVolHandle, _, err := procFindFirstVolumeW.Call(
```

disk/disk_windows.go
```diff
@@ -102,46 +114,26 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
...
-		return partitionStats, err // The call failed with an unexpected error
-	}
-	lpBuffer := make([]uint16, bufferLen)
-	// buffer can be longer than MAX_PATH
-	_, _, err = procGetLogicalDriveStringsW.Call(
-		uintptr(len(lpBuffer)),
-		uintptr(unsafe.Pointer(&lpBuffer[0])))
-	if !errors.Is(err, windows.ERROR_SUCCESS) {
-		return partitionStats, err // The call failed with an unexpected error
+	drives, logicalDrivesErr := getLogicalDrives()
```
shirou: I think we don't need this long name logicalDrivesErr, just err is enough.
```suggestion
	drives, err := getLogicalDrives()
```

disk/disk_windows.go
```diff
@@ -177,7 +169,46 @@ func PartitionsWithContext(_ context.Context, _ bool) ([]PartitionStat, error) {
...
+				partitionStats = append(partitionStats, partitionStat)
+			} else {
+				warnings.Add(warning)
+			}
+		}
+	}
+	return partitionStats
+}
+
+// We first call GetLogicalDriveStringsW with a buffer length of 0 to get the required buffer size.
```
shirou, line 192: The logic has been separated into functions, so the comment should follow the function comment format.
```suggestion
// getLogicalDrives retrieves all logical drives using GetLogicalDriveStringsW.
// We first call GetLogicalDriveStringsW with a buffer length of 0 to get the required buffer size.
```

shirou:
The errors are lint errors and those are fixed in #1942. So I think we can merge this PR. Thank you for you awesome contribution!

PR #1951, opened 2025-11-20
[mac] expose the memory pressure percentage value

atoulme:
This adds supports on Darwin system to expose the memory pressure value of the system.

This value is a percentage indicating the memory pressure of the system as a whole, computed by the kernel.

This is based off prior art:
https://github.com/exelban/stats/issues/1453
https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/osfmk/arm/commpage/commpage.c#L338
https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/osfmk/vm/vm_pageout.c#L1878

### DIFF
--- a/mem/mem_darwin_test.go
+++ b/mem/mem_darwin_test.go
@@ -5,7 +5,6 @@ package mem
 
 import (
 	"context"
-	"fmt"
 	"strconv"
 	"strings"
 	"testing"


anelson-splunk (approved):
Looks great

mem/mem_darwin_test.go
```diff
@@ -46,3 +48,18 @@ func TestVirtualMemoryDarwin(t *testing.T) {
...
+func TestMemoryPressureWithContext(t *testing.T) {
+	p, err := MemoryPressureWithContext(context.Background())
+	outBytes, err2 := invoke.Command("/usr/sbin/sysctl", "vm.memory_pressure")
+	require.NoError(t, err)
+	require.NoError(t, err2)
+	outString := string(outBytes)
+	outString = strings.TrimSpace(outString)
+	outParts := strings.Split(outString, " ")
+	pressure, err := strconv.ParseInt(outParts[1], 10, 64)
+	fmt.Println(pressure)
```
atoulme:
```suggestion
```

mem/mem_darwin_test.go
```diff
@@ -4,6 +4,8 @@
 package mem
 
 import (
+	"context"
+	"fmt"
```
atoulme:
```suggestion
```

shirou (changes requested):
Thank you for taking the time to implement this feature. I really appreciate your effort and contribution.

However, `gopsutil` aims to be a multi-platform library. We generally do not accept features that are specific to a single OS. At a minimum, new features need to be supported on Windows and Linux as well. And keeping strict alignment with Python's psutil API. Since this feature is macOS-specific and not part of the original psutil, adding it would deviate from our project goals.

Regrettably, I have to reject this PR. I hope you understand our position on maintaining a consistent scope for the library. Thanks again.

atoulme:
Not a problem. Great to get validation that this might be useful on Mac OS. We will directly implement this feature in the hostmetricsreceiver of the OpenTelemetry Collector for Mac OS X support only.

PR #1973, opened 2025-12-17
[sensors][darwin]: retrieve sensor information in one function call

uubulb:
The new `getSensors` function retrieves both sensor names and temperature metrics in a single call, avoiding the need to invoke `ioHIDEventSystemClientCopyServices` twice. It also performs deduplication internally, removing the need for a separate helper function.

Additionally, some code in the `process` package was refactored for readability. These changes are mostly equivalent in behavior and should not affect testing results.

### DIFF
--- a/process/types_darwin.go
+++ b/process/types_darwin.go
@@ -157,15 +157,15 @@ type Label C.struct_label
 
 type ProcTaskInfo C.struct_proc_taskinfo
 
-type VInfoStat C.struct_vinfo_stat
+type vinfoStat C.struct_vinfo_stat
 
-type Fsid C.struct_fsid
+type fsid C.struct_fsid
 
-type VnodeInfo C.struct_vnode_info
+type vnodeInfo C.struct_vnode_info
 
-type VnodeInfoPath C.struct_vnode_info_path
+type vnodeInfoPath C.struct_vnode_info_path
 
-type VnodePathInfo C.struct_proc_vnodepathinfo
+type vnodePathInfo C.struct_proc_vnodepathinfo
 
 type (
 	AuditinfoAddr C.struct_auditinfo_addr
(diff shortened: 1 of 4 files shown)


process/types_darwin.go
```diff
@@ -157,6 +157,16 @@ type Label C.struct_label
...
+type VInfoStat C.struct_vinfo_stat
+
+type Fsid C.struct_fsid
+
+type VnodeInfo C.struct_vnode_info
+
+type VnodeInfoPath C.struct_vnode_info_path
+
+type VnodePathInfo C.struct_proc_vnodepathinfo
+
```
shirou: These structs will become public. The existing structs were made public unintentionally, and the new types should remain private.

sensors/sensors_darwin_arm64.go
```diff
@@ -81,93 +72,75 @@ type temperatureArm struct {
...
 		sc := ta.cfArrayGetValueAtIndex(uintptr(matchingsrvs), i)
-		event := ioHIDServiceClientCopyEvent(uintptr(sc), common.KIOHIDEventTypeTemperature, 0, 0)
-		temp := 0.0
-
-		if event != nil {
-			temp = ioHIDEventGetFloatValue(uintptr(event), ioHIDEventFieldBase(common.KIOHIDEventTypeTemperature))
-			ta.cfRelease(uintptr(event))
+		event := ta.ioHIDServiceClientCopyEvent(uintptr(sc), common.KIOHIDEventTypeTemperature, 0, 0)
+		if event == nil {
+			continue
```
shirou, line 110: In the current implementation, when `event == nil`, the code continues with `temp = 0.0`. With this change, it will be skipped. Is this intended?

uubulb: Yes. In the current implementation, when `event` is `nil`, a placeholder sensor named `noname` is added to keep the arrays aligned. Since this is somewhat redundant, I decided to simply ignore sensors that have no values to read (i.e., sensors whose corresponding event is `nil`).

PR #1975, opened 2025-12-18
fix disk.Partition cut off after first disk

sni:
using windows.UTF16ToString() converts a []uint16 to a string but ends after the first NUL. There is already a split0() function which does the right thing in this file, so simply use that one.

before:

    getLogicalDrives() -> ['C:\']

after:

    getLogicalDrives() -> ['C:\', 'D:\', 'Z:\']

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"syscall"
 	"unicode/utf16"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


OrbintSoft (approved):

StefanoBalzarottiNozomi:
> There are tests in place and they failed on my machine before this fix.
> The problem is not the test coverage here.
Ok, I still think we have to find a determistic way  to test this  but your PR is fine for me.
Lets wait for @shirou

StefanoBalzarottiNozomi (approved):

PR #1976, opened 2025-12-19
[darwin]: wrap library functions as struct methods

uubulb:
`Library` is now a private struct, used to compose new structs representing different system libraries/frameworks: `CoreFoundationLib`, `IOKitLib`, and `SystemLib`. Each of them provides public methods that directly register and call the underlying library functions, improving readability by eliminating the need to use `GetFunc` in other packages.

A temporary cache `fnMap` is added to the `Library` struct to ensure the same lazy loading behavior as before.

~~Note: This PR contains changes from https://github.com/shirou/gopsutil/pull/1971 and https://github.com/shirou/gopsutil/pull/1973, so it has lower review priority than those. These changes are included here to make upcoming conflict fixing easier.~~

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -61,17 +61,17 @@ func Times(percpu bool) ([]TimesStat, error) {
 }
 
 func TimesWithContext(_ context.Context, percpu bool) ([]TimesStat, error) {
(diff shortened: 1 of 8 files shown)


shirou (approved):
Good refactoring. LGTM! Thank you so much.

PR #1982, opened 2025-12-23
fix: AIX memory statistics: add Available field and improve swap calculations

Dylan-M:
## Summary

Fixes and improves AIX memory statistics for the nocgo path:

### Memory (callSVMon)

- **Available field**: Populated from the real kernel-computed `available` column in `svmon -G -O unit=KB`, which includes free pages plus reclaimable cache. This value cannot be derived from raw page counts alone.
- **Precision preserved**: Primary stats (Total, Used, Free, swap) are read from raw `svmon -G` page output and multiplied by `os.Getpagesize()`, avoiding the sub-KB truncation that `-O unit=KB` introduces. A second `svmon` call with `-O unit=KB` is made only to read the `available` column.
- **Swap calculations fixed**: `swap.Used` is now explicitly set, and `swap.UsedPercent` is calculated.
- **Parsing extracted for testability**: `parseSVMonPages()` and `parseSVMonAvailable()` are separate functions with table-driven unit tests using real AIX sample output.

### Swap Devices (SwapDevicesWithContext)

- **GB/TB unit handling**: `parseLspsSize()` now handles MB, GB, and TB suffixes in `lsps -a` output.
- **Unit tests**: `parseLspsOutput()` and `parseLspsSize()` have comprehensive tests covering MB, GB, NaNQ (NFS paging spaces), multiple devices, and edge cases.

### CGO path (mem_aix_cgo.go)

- **SwapDevicesWithContext**: Implemented using `perfstat.PagingSpaceStat()`.

Only affects AIX — no cross-platform changes.

### DIFF
--- a/mem/mem_test.go
+++ b/mem/mem_test.go
@@ -50,7 +50,9 @@ func TestVirtualMemory(t *testing.T) {
 		"Total should be computable (%v): %v", totalStr, v)
 
 	assert.True(t, runtime.GOOS == "windows" || v.Free > 0)
-	assert.Truef(t, runtime.GOOS == "windows" || v.Available > v.Free,
+	// On AIX, Available is typically equal to Free
+	// On other systems, Available should be >= Free
+	assert.Truef(t, runtime.GOOS == "windows" || v.Available >= v.Free,
 		"Free should be a subset of Available: %v", v)
 
 	inDelta := assert.InDelta
--- a/mem/mem_test.go
+++ b/mem/mem_test.go
@@ -50,9 +50,7 @@ func TestVirtualMemory(t *testing.T) {
 		"Total should be computable (%v): %v", totalStr, v)
 
 	assert.True(t, runtime.GOOS == "windows" || v.Free > 0)
-	// On AIX, Available is typically equal to Free
-	// On other systems, Available should be >= Free
-	assert.Truef(t, runtime.GOOS == "windows" || v.Available >= v.Free,
+	assert.Truef(t, runtime.GOOS == "windows" || v.Available > v.Free,
 		"Free should be a subset of Available: %v", v)
 
 	inDelta := assert.InDelta
--- a/mem/mem_aix.go
+++ b/mem/mem_aix.go
@@ -5,8 +5,6 @@ package mem
 
 import (
 	"context"
-
-	"github.com/shirou/gopsutil/v4/internal/common"
 )
 
 func VirtualMemory() (*VirtualMemoryStat, error) {
@@ -18,5 +16,5 @@ func SwapMemory() (*SwapMemoryStat, error) {
 }
 
 func SwapDevices() ([]*SwapDevice, error) {
-	return nil, common.ErrNotImplementedError
+	return SwapDevicesWithContext(context.Background())
 }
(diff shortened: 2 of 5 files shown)


shirou (changes requested):
Hi @Dylan-M, I've reviewed this PR more carefully. Unlike your other PRs, this one is relatively small, has a single concern, and is focused only on AIX — which made it much easier to review thoroughly. I appreciate that.

I'm happy to merge this with the following changes:

1. Handle GB (and potentially other units) in `lsps -a` parsing: The current code assumes the size column always ends with "MB", but `lsps` can report sizes in GB on larger systems. Please handle at least MB and GB.
2. Add a unit test for `SwapDevicesWithContext` (nocgo path): The `lsps -a` parsing logic should have a test with sample output, similar to how other packages test their command parsing.

These are small fixes — once addressed, this is good to merge. Thanks!

mem/mem_test.go
```diff
@@ -50,7 +50,9 @@ func TestVirtualMemory(t *testing.T) {
 		"Total should be computable (%v): %v", totalStr, v)
 
 	assert.True(t, runtime.GOOS == "windows" || v.Free > 0)
-	assert.Truef(t, runtime.GOOS == "windows" || v.Available > v.Free,
+	// On AIX, Available is typically equal to Free
+	// On other systems, Available should be >= Free
+	assert.Truef(t, runtime.GOOS == "windows" || v.Available >= v.Free,
```
shirou: This relaxes the assertion for all non-Windows platforms. On Linux/macOS/BSD, Available == Free would indicate a real bug (missing cache/buffer accounting), and we'd no longer catch it. Could you scope it to AIX, matching the switch `runtime.GOOS` pattern already used above?

```
if runtime.GOOS == "aix" {
    assert.Truef(t, v.Available >= v.Free, ...)
} else if runtime.GOOS != "windows" {
    assert.Truef(t, v.Available > v.Free, ...)
}
```

### Is the relaxation actually needed?

If `available` = free + reclaimable cache (per the PR description), then `Available > Free` should hold on AIX too in the typical case — the equality only happens in the edge case where reclaimable cache is zero. Is that edge case actually hit in CI? If yes, the in-code comment should describe it as the edge case (e.g. "Available can equal Free when no reclaimable cache is present") rather than implying it's typical. If no, the assertion can stay `>` for AIX as well.

shirou (changes requested):
Overall looks good. I'd like to see tightened before merge.

Dylan-M:
Tested empirically on AIX 7.3.0.0 — across 30 sequential samples with a 500ms cadence, `Available > Free` held strictly with a stable gap of ~253 MB (range 252.5–254.6 MB). The kernel maintains a non-trivial reclaimable cache footprint in normal operation, so the equality case isn't hit in practice.

The relaxation was appropriate for an earlier iteration of this PR where `Available` was set equal to `Free` as a fallback, but became stale once the parsing switched to reading the kernel's real `available` value. Tightened back to `>` and dropped the AIX-specific note. Pushed.

Dylan-M:
@shirou — going to follow your guidance and close out the AIX stack to restart with the one-PR-at-a-time discipline you asked for. The other 20+ PRs and the tracking issue close today.

One carve-out question on this PR specifically: it's been through three rounds of your review, the changes you asked for in the last round are in, and the diff is strictly AIX-only (`mem/mem_aix*.go` plus a single test-assertion edit). Would you prefer I close this one too and re-submit as the first PR under the new pattern, or keep it open since the review investment is already largely there?

PR #2032, opened 2026-03-21
disk: fix UsageWithContext unit conversions on AIX nocgo

Dylan-M:
## Summary

Fixes several bugs in the AIX nocgo `UsageWithContext` implementation:

- **Used/Free not converted to bytes**: The `df -v` output reports values in 512-byte blocks, but they were stored directly without multiplying by the blocksize. Now correctly converts to bytes.
- **UsedPercent/InodesUsedPercent incorrect scale**: The `%Used` and `%Iused` columns from `df` are already whole percentages (e.g., `42` means 42%), but the code was dividing by 100, producing values like `0.42`. Removed the incorrect division.
- **Path matching broken**: The code compared the `Filesystem` column (device name like `/dev/hd4`) against the requested path (mount point like `/var`), which never matched. Now correctly matches against the `Mounted on` column and iterates all lines instead of returning the first filesystem.
- **Dash values cause parse errors**: AIX virtual filesystems (`/proc`, `/ahafs`) report `-` for numeric fields. `strconv.ParseUint` fails on `-`. Now treats `-` as zero for all numeric fields.
- **Total assigned before error check**: `ret.Total = total * blocksize` was executed before checking the `strconv.ParseUint` error. Reordered to check error first.
- **No error on missing path**: Previously returned an empty struct if no filesystem matched. Now returns `fmt.Errorf("mountpoint %s not found", path)`, consistent with the CGO variant.

Only affects AIX nocgo path — no cross-platform changes.

### DIFF
--- a/disk/disk_aix_nocgo.go
+++ b/disk/disk_aix_nocgo.go
@@ -5,6 +5,7 @@ package disk
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


shirou (approved):
LGTM. Both fixes are clearly correct:

- `Used`/`Free` need the same `* blocksize` conversion that `Total` already has
- `df -v` outputs whole percentages (e.g. `42%`), so dividing by 100 again was wrong

This PR is clear and focused on one concern — much easier to review. Thanks for splitting things up!

One minor note (not related to this PR): the pre-existing `512-blocks` case at L128-132 executes `ret.Total = total * blocksize` before checking `err`. Your new `Used`/`Free` code correctly checks `err` first — might be worth a follow-up to fix the `Total` case to match:

```go
  case `512-blocks`:
      total, err := strconv.ParseUint(fs[i], 10, 64)
      if err != nil {
          return nil, err
      }
      ret.Total = total * blocksize
```

shirou (approved):
LGTM — all the bug fixes look correct and the `Total` assignment order fix from #2059 is properly folded in.

PR #2044, opened 2026-03-21
host: collect errors in InfoWithContext instead of failing on first error

Dylan-M:
## Summary

Changes `InfoWithContext()` to collect all non-fatal errors into a slice and return them via `errors.Join()` at the end, rather than returning immediately on the first error.

**Motivation**: On AIX, some host info functions (e.g., `numProcs` in nocgo mode) may fail while others succeed. The current fail-fast behavior means a single broken function prevents the caller from getting any host info at all — even fields that would have succeeded.

**Behavior change**: Previously, `InfoWithContext` returned `(nil, err)` on the first non-`ErrNotImplementedError` failure. Now it returns `(nil, errors.Join(errs...))` containing all failures. Callers using `errors.Is()` or `errors.As()` can still match individual errors within the joined error.

This is a cross-platform behavioral change. I'd appreciate discussion on whether this approach works for the project.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -64,53 +64,58 @@ func Info() (*InfoStat, error) {
 
 func InfoWithContext(ctx context.Context) (*InfoStat, error) {
 	var err error
+	var errs []error
 	ret := &InfoStat{
 		OS: runtime.GOOS,
 	}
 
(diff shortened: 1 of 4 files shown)


Dylan-M:

PR #2048, opened 2026-03-21
process: add SignalsPendingWithContext to Process interface

Dylan-M:
Adds a `SignalsPending()` wrapper and `SignalsPendingWithContext` method to the Process type. Linux gets a real implementation reading from /proc/<pid>/status. All other platforms return `ErrNotImplementedError`. This is a cross-platform interface addition needed for AIX process support.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -465,6 +465,11 @@ func (p *Process) Terminal() (string, error) {
 	return p.TerminalWithContext(context.Background())
 }
 
+// SignalsPending returns the signals pending for the process.
+func (p *Process) SignalsPending() (SignalInfoStat, error) {
+	return p.SignalsPendingWithContext(context.Background())
+}
+
 // Nice returns a nice value (priority).
 func (p *Process) Nice() (int32, error) {
 	return p.NiceWithContext(context.Background())
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -521,3 +521,7 @@ func (p *Process) NumFDsWithContext(_ context.Context) (int32, error) {
 	numFDs := ret / sizeofProcFDInfo
 	return numFDs, nil
 }
+
+func (*Process) SignalsPendingWithContext(_ context.Context) (SignalInfoStat, error) {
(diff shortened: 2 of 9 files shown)


shirou:
Thanks for the PR, but I'd like to close this for those reasons:

1. Not in psutil's API: gopsutil follows psutil's Process API surface, and psutil does not expose a `signals_pending` equivalent — only signal-sending methods (`send_signal`, `terminate`, etc.). We generally reject new methods that don't have a psutil counterpart.

2. Insufficient platform coverage.: Our policy is that new public methods must be implemented on at least two of Linux/macOS/Windows. This PR only implements Linux, and signal pending/blocked/ignored/caught masks don't map cleanly to macOS or Windows process models.

3. AIX-specific needs shouldn't require cross-platform API additions.: If this is needed internally for AIX process support, please keep it as a private helper within the AIX files rather than adding a public method to all platforms. Note that `sigInfo` is already populated by `fillFromStatusWithContext` on Linux and consumed internally by `RlimitUsage` — the data is available without a new public API.

4. Is this API actually needed?: Signal pending/blocked/ignored/caught masks have a poor fit with gopsutil's primary use cases (observability and monitoring). They are mainly useful for low-level debugging of stuck signal handlers, not for metrics collection, health  checks, or resource monitoring. I'd like to understand the concrete use case before adding this to the public API surface.

Please feel free to reopen if the AIX work can use an internal helper instead.

Dylan-M:
You're right, and thanks for the clear explanation. I got overzealous implementing every piece of data I could pull out of AIX and forgot to check that each one had a psutil counterpart. I've removed the AIX-specific `SignalsPendingWithContext` implementation from #2049 — the internal `sigInfo` field is still populated for consistency with the Linux implementation, matching your note about how it's used internally for `RlimitUsage`.

I also did an audit of the rest of the open AIX PRs with the same lens (psutil parity + cross-platform coverage) and removed another similar case. I'll make an effort to be better about checking psutil parity first going forward.

PR #2060, opened 2026-03-21
cpu: cache prtconf output for static CPU info on AIX nocgo

Dylan-M:
> ⚠️ **Review order:** This PR depends on #2036 being merged first. See #2072 for the full review order.

prtconf returns static CPU hardware information (model, speed, count) that does not change at runtime. Every call to InfoWithContext was spawning a new subprocess to re-query this immutable data. This caches the output using sync.Once so the subprocess is only spawned on the first call.

Only affects AIX nocgo path — no cross-platform changes.


Dylan-M:

PR #2061, opened 2026-03-21
host: cache static system info on AIX to avoid repeated subprocess spawns

Dylan-M:
> ⚠️ **Review order:** This PR depends on #2040 being merged first. See #2072 for the full review order.

## Summary

AIX host info functions (HostID, PlatformInformation, KernelVersion, KernelArch) spawn subprocesses (`uname`, `oslevel`, `bootinfo`) to query system identifiers that never change at runtime. Each call was re-executing these commands unnecessarily.

This caches the results using `sync.Once` so each command is only executed on first access. Dynamic data (boot time, uptime, users, virtualization) is not cached.

> **Note:** This PR depends on #2040 being merged first (uses getInvoker pattern from that PR).

Only affects AIX — no cross-platform changes.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -64,53 +64,58 @@ func Info() (*InfoStat, error) {
 
 func InfoWithContext(ctx context.Context) (*InfoStat, error) {
 	var err error
+	var errs []error
 	ret := &InfoStat{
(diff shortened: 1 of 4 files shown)


Dylan-M:

PR #2062, opened 2026-03-22
perf(process): build snapshot map once on Windows to fix O(N²) process listing

### LINKED ISSUE #818: Combine Process .Ppid(), .Name(), NumThreads() for Windows

HarshalPatel1972:
Fixes #818

## Problem
On Windows, `ProcessesWithContext` calls `getFromSnapProcess` once per process to retrieve name, PPID, and thread count. `getFromSnapProcess` internally iterates through the entire `CreateToolhelp32Snapshot` result each time — making the overall complexity O(N²). On a machine with ~320 processes this results in ~13 seconds and 30–40% CPU usage, vs ~100ms on Linux.

## Fix
Added `buildSnapProcessMap()` which calls `CreateToolhelp32Snapshot` exactly once and builds a `map[uint32]windows.ProcessEntry32` keyed by PID. `ProcessesWithContext` now calls this once at startup and does O(1) per-process lookups for name, PPID, and thread count.

`getFromSnapProcess` is left intact for single-process query paths — no API changes.

## Benchmark
| | Before | After |
|---|---|---|
| 320 processes — Name + Ppid + NumThreads | ~13,130ms | ~68ms |
| Improvement | | **~99.5% faster** |

Tested on Windows 11 with ~320 running processes.

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -336,7 +336,9 @@ func (p *Process) NameWithContext(ctx context.Context) (string, error) {
 		return "", fmt.Errorf("could not get Name: %w", err)
 	}
 
-	return filepath.Base(exe), nil
+	name := filepath.Base(exe)
+	p.name = name
+	return name, nil
 }
 
 func (*Process) TgidWithContext(_ context.Context) (int32, error) {
@@ -925,7 +927,10 @@ func buildSnapProcessMap() (map[uint32]windows.ProcessEntry32, error) {
 	for {
 		snapMap[pe32.ProcessID] = pe32
 		if err = windows.Process32Next(snap, &pe32); err != nil {
-			break
+			if err == windows.ERROR_NO_MORE_FILES {
+				break
+			}
+			return nil, err
 		}
 	}
 	return snapMap, nil
@@ -941,7 +946,7 @@ func ProcessesWithContext(ctx context.Context) ([]*Process, error) {
 
 	snapMap, err := buildSnapProcessMap()
 	if err != nil {
-		return nil, err
+		return out, fmt.Errorf("could not build process snapshot: %w", err)
 	}
 
 	for _, pid := range pids {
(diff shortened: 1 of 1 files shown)


shirou (changes requested):
Good fix for a real performance problem. A few items to address:

1. In `NameWithContext`, the `p.name != ""` early return should be placed after the `p.Pid == 0` check, not before — otherwise PID 0 may return the snapshot's ExeFile name instead of the hardcoded "System Idle Process".
2. The `numThreads != 0` cache guard is incorrect — 0 is a valid thread count. Either remove the early return (the snapshot-populated value will flow through the existing code), or use a separate bool to track whether it's been cached.
3. Minor: Consider adding a brief comment in `ProcessesWithContext` noting that the snapshot and PID enumeration are separate calls
  and may not be perfectly consistent.

shirou (approved):
Thanks for the iterations — all the previous feedback is addressed in bf6e9497 (PID 0 ordering, removed the `numThreads != 0` guard, snapshot/PID consistency comment, ERROR_NO_MORE_FILES handling, error wrapping, and writing name back). Builds fine for GOOS=windows, and the O(N²) listing is correctly fixed.

Thanks you for your contribution!

PR #2064, opened 2026-03-23
load: add SysCalls and Interrupts fields to MiscStat

Dylan-M:
## Summary

Adds two new fields to the cross-platform MiscStat struct:

- **SysCalls** — cumulative number of system calls since boot
- **Interrupts** — cumulative number of device interrupts since boot

These follow the same pattern as the existing Ctxt (context switches) field, which is already populated on Linux from /proc/stat and on AIX from vmstat -s.

### Why

System call and interrupt rates are standard system health metrics. Without these fields, the only way to expose them was through platform-specific standalone functions that break cross-platform compilation and are inconsistent with how Ctxt is already exposed via MiscStat.

### Backward compatibility

New fields default to zero, so existing code that reads MiscStat is unaffected. JSON serialization gains two new keys (sysCalls, interrupts) with zero values on platforms that don't populate them. The test for MiscStat.String() is updated accordingly.

No platform implementations are changed in this PR — this only adds the struct fields and updates the serialization test. Populating the fields on AIX is done in a separate PR (#2039).

### DIFF
--- a/load/load.go
+++ b/load/load.go
@@ -25,7 +25,17 @@ type MiscStat struct {
 	ProcsCreated int `json:"procsCreated"`
 	ProcsRunning int `json:"procsRunning"`
 	ProcsBlocked int `json:"procsBlocked"`
-	Ctxt         int `json:"ctxt"`
+	// Ctxt is the cumulative number of context switches since boot.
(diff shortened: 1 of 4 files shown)


Dylan-M:

PR #2066, opened 2026-03-31
Fix infinite loop when failed to find path for volume on Windows

woct0rdho:
When `getVolumePaths` returns an error, we should not just continue the for loop. We should skip `processMountsForVolume`, then update `volNameBuf ` using `procFindNextVolumeW`. Otherwise it would loop forever.

I discovered this bug when running wandb, which uses gopsutil to collect system info, and there is a virtual CD program on my machine that leaves a stale volume that causes an error in `getVolumePaths`. It's discussed in https://github.com/shirou/gopsutil/issues/1911 but not really fixed yet.

### DIFF
--- a/disk/disk_windows.go
+++ b/disk/disk_windows.go
@@ -140,20 +140,6 @@ func processVolumesMountedAsFolders(ctx context.Context, partitionStats []Partit
 }
 
 func processVolumeLoop(ctx context.Context, nextVolHandle uintptr, volNameBuf []uint16, processedPaths map[string]struct{}, partitionStats []PartitionStat, warnings Warnings) []PartitionStat {
-	return processVolumeLoopWith(ctx, nextVolHandle, volNameBuf, processedPaths, partitionStats, warnings, getVolumePaths, processMountsForVolume, findNextVolume)
-}
-
-func processVolumeLoopWith(
-	ctx context.Context,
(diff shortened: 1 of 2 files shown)


StefanoBalzarottiNozomi:
thanks for the fix, but are you able to add a test to simulate the problem ?

woct0rdho:
I've refactored it to allow inserting custom functions `getVolumePathsFn, processMountsForVolumeFn, findNextVolumeFn`, then test how many times these functions are called in `processVolumeLoopWith`.

I'm not very familiar with Go. In Python people would monkey-patch `getVolumePaths` to test it, but I'm not sure if there is a way to do so in Go.

StefanoBalzarottiNozomi (approved):

shirou (changes requested):
Thank you for your PR. The fix for the infinite loop is a nice catch!

I appreciate you adding the test based on the review feedback. However, the test only verifies the if/else control flow, which is fairly straightforward. I think changing the function signature to 9 arguments just for this test is a bit much.

Sorry for the extra work, but could you revert to your original minimal fix? That's all we need here. Thanks!

woct0rdho:
Ok I've reverted the test and kept it a minimal change.

shirou (approved):
Thank you so much!

PR #2071, opened 2026-04-11
[net][linux]: guard against malformed /proc/net/dev lines to avoid panic

### LINKED ISSUE #2069: Panic in net.IOCounters on malformed /proc/net/dev lines (missing bounds check)
## Problem

`IOCountersByFileWithContext` in `net/net_linux.go` (lines 71-112) parses `/proc/net/dev` and accesses `fields[0]` through `fields[12]` without first checking that `len(fields) >= 13`. This causes an index-out-of-range panic when any interface line has fewer stats columns than expected.

**Affected file:** `net/net_linux.go:71-112`

The same codebase already includes the correct pattern in `disk/disk_linux.go:463-464`:

```go
if len(fields) < 14 {
    // malformed line in /proc/diskstats, avoid panic by ignoring.
    continue
}
```

The net parser is missing the equivalent guard.

## Reproduction

1. Set `HOST_PROC` to a directory containing a `/proc/net/dev` with a truncated interface line:
   ```
   Inter-|   Receive                                                |  Transmit
    face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
   eth0: 0 0 0
   ```
2. Call `net.IOCounters(true)`
3. **Result:** `panic: runtime error: index out of range` (or similar)

This can also occur in production with:
- Container environments where `/proc/net/dev` exposes partial kernel state
- Virtual network drivers (WireGuard, macvlan, IPVLAN) that report fewer stats
- Custom `HOST_PROC` paths used for container process inspection

## Impact

Process crash — callers cannot recover from a panic in a library function without explicit `recover()` wrapping. Affects containerized monitoring agents and custom proc-filesystem deployments.

## Suggested Fix

Add a bounds check before field access, consistent with the existing `disk/disk_linux.go` pattern:

```go
fields := strings.Fields(strings.TrimSpace(statsPart))
if len(fields) < 13 {
    // malformed line in /proc/net/dev, avoid panic by ignoring
    continue
}
bytesRecv, err := strconv.ParseUint(fields[0], 10, 64)
// ... rest of parsing
```

Found via automated codebase analysis.

shirou:
`IOCountersByFileWithContext `accessed fields[0]..fields[12] without checking the slice length, causing an index-out-of-range panic on truncated interface lines (custom HOST_PROC, partial reads, etc.). Skip such lines instead, matching the existing pattern in `disk/disk_linux.go`.

Fixes #2069

### DIFF
--- a/net/net_linux.go
+++ b/net/net_linux.go
@@ -69,6 +69,10 @@ func IOCountersByFileWithContext(_ context.Context, pernic bool, filename string
 		}
 
 		fields := strings.Fields(strings.TrimSpace(statsPart))
+		if len(fields) < 13 {
+			// malformed line in /proc/net/dev, avoid panic by ignoring.
+			continue
+		}
 		bytesRecv, err := strconv.ParseUint(fields[0], 10, 64)
 		if err != nil {
 			return ret, err


shirou:
dup #2075

PR #2073, opened 2026-04-12
net: add parseEntstat and parseNetstatI unit tests with AIX fixtures

Dylan-M:
Follow-up to #2034 as requested in review — adds unit tests for `parseEntstat` and `parseNetstatI` with captured output from a real AIX 7.3 system, guarding against future format drift.

The `entstat` command and its two-column Transmit/Receive format (including the `Bytes:` line) have been stable across AIX 4.3 through 7.3, spanning over 25 years of releases. The `netstat -idn` format has similarly been stable across AIX versions.

### Tests added

**parseEntstat:**
- `TestParseEntstat` — verifies byte extraction from real entstat en0 output
- `TestParseEntstatEmpty` — empty input returns zeros
- `TestParseEntstatNoBytesLine` — output without "Bytes:" line returns zeros
- `TestParseEntstatSingleColumn` — transmit-only output correctly returns zero for receive

**parseNetstatI:**
- `TestParseNetstatI` — verifies packet/error parsing from real netstat -idn output, including deduplication (en0 appears twice, lo0 three times, both correctly deduplicated to one entry each)
- `TestParseNetstatIInvalidHeader` — non-netstat output returns error
- `TestParseNetstatIEmpty` — header-only output returns empty result

Also adds a comment to `parseEntstat` noting the long-term format stability.

Only affects AIX nocgo test path — no behavioral changes.

### DIFF
--- a/net/net_aix_nocgo.go
+++ b/net/net_aix_nocgo.go
@@ -79,6 +79,8 @@ func parseNetstatI(output string) ([]IOCountersStat, error) {
 }
 
 // parseEntstat extracts BytesSent and BytesRecv from entstat output.
+// The entstat two-column Transmit/Receive format (including the "Bytes:"
+// line) has been stable across AIX 4.3 through 7.3 (over 25 years).
 // The output has a two-column layout with Transmit on the left and Receive
 // on the right, e.g.:
 //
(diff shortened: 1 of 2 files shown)


shirou (approved):
OK. Thank you for the follow up!

PR #2076, opened 2026-04-13
fix: add bounds check for /proc/[pid]/stat fields in fillFromTIDStat

### LINKED ISSUE #2070: Panic in process stat parsing on short /proc/[pid]/stat (missing bounds check)
## Problem

`fillFromTIDStatWithContext` in `process/process_linux.go` (lines 1048-1096) parses `/proc/[pid]/stat` via `splitProcStat` and accesses indices `fields[4]`, `fields[7]`, `fields[14]`, `fields[15]`, `fields[18]`, and `fields[22]` without any bounds check. This causes an index-out-of-range panic when the stat file has fewer than 23 fields.

**Affected file:** `process/process_linux.go:1048-1096`

Notably, the same function **does** guard `fields[42]` at line 1073:

```go
if len(fields) > 42 {
    iotime, err = strconv.ParseFloat(fields[42], 64)
} else {
    iotime = 0 // e.g. SmartOS containers
}
```

The comment explicitly acknowledges that non-standard environments can return short stat files, but the same protection is not applied to the lower-indexed fields.

## Reproduction

1. Set `HOST_PROC` to a directory containing a `/proc/[pid]/stat` with fewer than 23 fields:
   ```
   1234 (myproc) S 0 1234 1234
   ```
2. Call `process.NewProcess(1234)` followed by `Times()`, `Ppid()`, or `CreateTime()`
3. **Result:** `panic: runtime error: index out of range [22] with length 6`

Also reproducible in:
- Container environments where `/proc/[pid]/stat` exposes truncated kernel state
- Processes in zombie or incomplete kernel states that emit partial stat lines
- Custom `HOST_PROC` paths used for container process inspection

## Impact

Process crash — callers cannot recover from a panic in a library function without explicit `recover()` wrapping. Affects container monitoring agents using `HOST_PROC` and cross-container process inspection.

## Suggested Fix

Add an upfront bounds guard before accessing any field:

```go
fields := splitProcStat(contents)
if len(fields) < 23 {
    return 0, 0, nil, 0, 0, 0, nil, fmt.Errorf("process %d: stat has %d fields, need at least 23", pid, len(fields))
}
```

The minimum unconditional access is `fields[4]`; the maximum is `fields[22]`. A single check `len(fields) < 23` covers all unguarded accesses. The existing `len(fields) > 42` guard for `fields[42]` can remain as-is.

Found via automated codebase analysis.

Yanhu007:
Fixes #2070

## Problem

`fillFromTIDStatWithContext` accesses `fields[4]`, `fields[7]`, `fields[14]`, `fields[15]`, `fields[18]`, and `fields[22]` without bounds checking. This causes an index-out-of-range panic when `/proc/[pid]/stat` has fewer than 23 fields.

The function already guards `fields[42]` at line 1073:
```go
if len(fields) > 42 {
    iotime, err = strconv.ParseFloat(fields[42], 64)
} else {
    iotime = 0 // e.g. SmartOS containers
}
```

But the same protection is missing for the lower-indexed fields.

## Fix

Add an early bounds check:
```go
if len(fields) < 23 {
    return 0, 0, nil, 0, 0, 0, nil, fmt.Errorf("malformed stat file: expected at least 23 fields, got %d", len(fields))
}
```

This complements the fix in #1995 which added a similar check in a different code path.

### DIFF
--- a/process/process_linux.go
+++ b/process/process_linux.go
@@ -1046,6 +1046,9 @@ func (p *Process) fillFromTIDStatWithContext(ctx context.Context, tid int32) (ui
 	}
 	// Indexing from one, as described in `man proc` about the file /proc/[pid]/stat
 	fields := splitProcStat(contents)
+	if len(fields) < 23 {
+		return 0, 0, nil, 0, 0, 0, nil, fmt.Errorf("malformed stat file: expected at least 23 fields, got %d", len(fields))
+	}
 
 	terminal, err := strconv.ParseUint(fields[7], 10, 64)
 	if err != nil {


shirou (approved):
LGTM. Minimal and correct fix — the < 23 threshold matches the highest low-index access (fields[22]), and the early return prevents the panic cleanly.

Thanks for the fix!

PR #2077, opened 2026-04-14
Fix NetBSD mem / net stats

fraggerfox:
This pull request adds NetBSD support to the `net` package and makes a small fix to the `mem` package for NetBSD.

### NetBSD network support

* Implement basic `netstat` based information extraction for NetBSD in `net_netbsd.go`.
  * The implementation copies the basic file structure from the OpenBSD implementation, due to the closeness of how `netstat` works in both of these systems.
* Updated the fallback build constraint in `net_fallback.go` to exclude NetBSD, ensuring the new NetBSD-specific implementation is used on that platform.

### NetBSD memory support

* Fixed the `Sscanf` format string in `mem_netbsd.go` to match the actual output from `swapctl` on NetBSD, changing from 1K-blocks to KBytes.

### Example output

This is the output generated by a sample golang program running in my NetBSD vm.

```
=== Virtual Memory ===
  Total:           7916.9 MB
  Used:            6002.8 MB (75.8%)
  Available:       1914.1 MB
  Free:              21.7 MB

=== Swap Memory ===
  Total:           4096.0 MB
  Used:             292.3 MB (7.1%)
  Free:            3803.6 MB

=== CPU Usage ===
  Total:       5.0%
  CPU0:        2.0%
  CPU1:        4.0%
  CPU2:        5.9%
  CPU3:        7.9%

=== Network IO ===
  vioif         recv:   952756.7 KB  sent:  1289601.2 KB  pkts-in: 6029800  pkts-out: 10704817  errin: 0  errout: 0  dropin: 0  dropout: 0
  lo0           recv:    25617.4 KB  sent:    25617.4 KB  pkts-in: 524644  pkts-out: 524644  errin: 0  errout: 0  dropin: 0  dropout: 0
```

### References
- https://man.netbsd.org/netstat.1
- https://man.netbsd.org/swapctl.8

> P.S: Some of the code has been written using AI Agent assistance.


shirou (changes requested):
Thanks for the PR! Overall this looks good, but a couple of requests:

1. Unexport ParseNetstat

This function is an internal parsing helper and doesn't need to be part of the public API. Please rename it to `parseNetstat`. (The OpenBSD version happens to be exported, but that's historical — no need to repeat it.)

2. Add tests

Since the function takes a string and is exported, it's easy to test with table-driven cases using captured netstat output. I'd suggest placing fixture files under `net/testdata/netbsd/` (following the convention used in `cpu/testdata/freebsd/`), for example:

net/testdata/netbsd/netstat_inb.txt
net/testdata/netbsd/netstat_ind.txt

Ideally the fixtures should cover both cases the code handles:
- interfaces with an Address column (e.g. vioif0)
- interfaces without it (e.g. lo0)

fraggerfox:
> Thanks for the PR! Overall this looks good, but a couple of requests:
> 
>     1. Unexport ParseNetstat
> 
> 
> This function is an internal parsing helper and doesn't need to be part of the public API. Please rename it to `parseNetstat`. (The OpenBSD version happens to be exported, but that's historical — no need to repeat it.)
> 
>     2. Add tests
> 
> 
> Since the function takes a string and is exported, it's easy to test with table-driven cases using captured netstat output. I'd suggest placing fixture files under `net/testdata/netbsd/` (following the convention used in `cpu/testdata/freebsd/`), for example:
> 
> net/testdata/netbsd/netstat_inb.txt net/testdata/netbsd/netstat_ind.txt
> 
> Ideally the fixtures should cover both cases the code handles:
> 
>     * interfaces with an Address column (e.g. vioif0)
> 
>     * interfaces without it (e.g. lo0)

Yeah I copied over the OpenBSD implementation, let me go through some example tests and make the changes you requested.

fraggerfox:
> > Thanks for the PR! Overall this looks good, but a couple of requests:
> 

These should be done in the latest commit.

net/net_netbsd_test.go
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: BSD-3-Clause
+package net
```
shirou, line 4: Could you add a `//go:build netbsd` build tag here? The file name contains `_netbsd`, but adding it explicitly would be safer.

net/net_netbsd_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseNetstat(t *testing.T) {
+	if runtime.GOOS != "netbsd" {
+		t.SkipNow()
+	}
+
```
shirou: After the build tag inserted, this condition could be removed.

shirou (commented):
Thank you for adding the tests. They look good overall, but could you make a small tweak?

net/net_netbsd_test.go
fraggerfox: Makes sense, the latest commit should address this.

net/net_netbsd_test.go
fraggerfox: Should be done in the latest commit.

shirou (approved):
Thank you for your great contribution!

PR #2079, opened 2026-04-20
Use getconf instead of bootinfo on AIX to get kernel architecture

pgimalac:
`bootinfo` is a deprecated AIX-specific tool, while `getconf` is a POSIX-standard utility that provides the same information.

Additionally, `bootinfo` requires special permissions to run in some AIX environments (`bin` group), whereas `getconf` is executable by all users out of the box (NB: at least on my host).

This PR updates `KernelArch()` to try `getconf KERNEL_BITMODE` first, and falls back to `bootinfo -y` if `getconf` fails, preserving backward compatibility.

https://www.ibm.com/support/pages/use-getconf-instead-bootinfo

### DIFF
--- a/host/host_aix.go
+++ b/host/host_aix.go
@@ -116,9 +116,12 @@ func KernelVersionWithContext(ctx context.Context) (version string, err error) {
 }
 
 func KernelArch() (arch string, err error) {
-	out, err := getInvoker().Command("bootinfo", "-y")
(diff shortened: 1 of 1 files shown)


shirou (approved):
LGTM. Sensible migration per IBM's own guidance, and the `bootinfo` fallback keeps backward compatibility. Thank you!

Issue #887, opened 2020-05-31
In windows with multiple processor groups, cpu.Percent(percpu=true) returns cpus info of any one processor group only. 


msays2000:
**Describe the bug**
In a windows instance having 36 cpu i.e having 36 * 2 logical cpus, windows groups CPU's into batch of 64 logical cpu in a processor group. In this example, it will have
- 64 logical cpus in processor group A
-   8 logical cpus in processor group B

gopsutil's cpu.Percent(percpu=true) randomly picks a processor group only and surfaces its result. In above example, if you run once it will log per cpu results showing 64 cpu's. It you stop and run again it will show the other processor group of 8 cpus. 

**To Reproduce**
```go
package main

import (
    "fmt"

    "github.com/shirou/gopsutil/cpu"
)

const (
	CPUPercentDuration = time.Millisecond * 200
)

func main() {
    cpuPercents, err := cpu.Percent(CPUPercentDuration, true)

    fmt.Printf("logical cpus %v \n", cpuPercents)

}
```

**Expected behavior**
Expected is it should return total cpu info of 72 logical cores consistently.
Actual is it returns either 64 or 8 cpus. 

**Environment (please complete the following information):**
 - [X ] Windows: [Windows 2016 Server Standard]

**Additional context**
Noticed the bug is calling an API that is probably not CPU processor group away. 
https://github.com/shirou/gopsutil/blob/master/cpu/cpu_windows.go#L188-L193
```
       // Invoke windows api proc.
	// The returned err from the windows dll proc will always be non-nil even when successful.
	// See https://godoc.org/golang.org/x/sys/windows#LazyProc.Call for more information
	retCode, _, err := common.ProcNtQuerySystemInformation.Call(
		win32_SystemProcessorPerformanceInformationClass, // System Information Class -> SystemProcessorPerformanceInformation
		uintptr(unsafe.Pointer(&resultBuffer[0])),        // pointer to first element in result buffer
		bufferSize,                        // size of the buffer in memory
		uintptr(unsafe.Pointer(&retSize)), // pointer to the size of the returned results the windows proc will set this
	)

```
In https://github.com/shirou/gopsutil/blob/master/internal/common/common_windows.go#L64
ProcNtQuerySystemInformation = ModNt.NewProc("NtQuerySystemInformation")

Probably we should make use of NtQuerySystemInformationEx
This function gets a wide range of system properties but allows for refining the query by specifying such things as a **processor group**.

msays2000:
[...]

PR #2081, opened 2026-04-24
[cpu][windows]: fix percpu stats on Windows hosts with multiple processor groups

shirou:
## Symptom

On Windows hosts with more than 64 logical CPUs (e.g. 36 cores × 2 HT = 72 logical CPUs, multi-socket servers, AWS EC2 instances with 96 vCPUs, etc.), `cpu.Percent(interval, percpu=true)` **returns the CPUs of only one Processor Group** instead of all CPUs.

- On a 72-CPU host, each call returns either 64 or 8 entries
- Which group is returned depends on which Processor Group the calling thread happens to be running on, so the result is non-deterministic
- Meanwhile `cpu.Counts(logical=true)` correctly returns 72 — so the array length of `Percent` does not match `Counts`, making aggregation by callers impossible

This has been reproduced by the reporter of #887 (72-CPU Server 2016), and by other users on a 92-vCPU Server 2022 and a 96-CPU EC2 instance.

## Cause

Windows splits logical CPUs into [**Processor Groups**](https://learn.microsoft.com/en-us/windows/win32/procthread/processor-groups) (up to 64 CPUs per group) on systems with more than 64 logical CPUs. Many legacy Win32 APIs only see "the group the calling thread belongs to."

`perfInfo()` uses [`NtQuerySystemInformation(SystemProcessorPerformanceInformation)`](https://docs.microsoft.com/en-us/windows/desktop/api/winternl/nf-winternl-ntquerysysteminformation#system_processor_performance_information) (class 8), which is exactly one of those non-group-aware APIs: it returns data only for the CPUs in the calling thread's group. Because the Go runtime may schedule the goroutine on any group, the result is non-deterministic.

Note that `cpu.Counts(logical=true)` is already Processor Group-aware via [`GetActiveProcessorCount(ALL_PROCESSOR_GROUPS)`](https://learn.microsoft.com/en-us/windows/win32/api/processtopologyapi/nf-processtopologyapi-getactiveprocessorcount) in `CountsWithContext` (`cpu/cpu_windows.go`). psutil handles this the same way ([psutil reference](https://github.com/giampaolo/psutil/blob/d01a9eaa35a8aadf6c519839e987a49d8be2d891/psutil/_psutil_windows.c#L97)). Only `perfInfo()` had been left behind.

## Fix

Use [`NtQuerySystemInformationEx`](https://www.geoffchappell.com/studies/windows/km/ntoskrnl/api/ex/sysinfo/queryex.htm),  which takes a group number (`USHORT`) as `InputBuffer` and returns data for only that group's CPUs. Iterate over all active groups and concatenate the results.

- Get the number of groups via [`GetActiveProcessorGroupCount`](https://learn.microsoft.com/en-us/windows/win32/api/processtopologyapi/nf-processtopologyapi-getactiveprocessorgroupcount)
- For each group `g`, size the buffer using `GetActiveProcessorCount(g)` and call `NtQuerySystemInformationEx`
- When `NtQuerySystemInformationEx` cannot be resolved (older OS without the Ex entry point), fall back to the original `NtQuerySystemInformation` path

C++ reference implementation: https://github.com/r3p3r/WindowsInternals/blob/05638d4d28cfb1bdfe7d7d1d1331d5be541f91a3/CpuSet/CpuSet.cpp#L180-L182

`NtQuerySystemInformationEx` is available on Windows 7 / Server 2008 R2 and later, which matches gopsutil's supported OS range.

## Scope

- `Times(percpu=true)` — now returns entries for all CPUs
- `Percent(percpu=true)` — fixed automatically since it's built on `Times()`
- `Times(percpu=false)` — still uses `GetSystemTimes`, which is also not Processor Group-aware; out of scope for this PR (better tracked as a separate issue)

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -25,6 +25,7 @@ var (
 	procGetLogicalProcessorInformationEx = common.Modkernel32.NewProc("GetLogicalProcessorInformationEx")
 	procGetSystemFirmwareTable           = common.Modkernel32.NewProc("GetSystemFirmwareTable")
 	procCallNtPowerInformation           = common.ModPowrProf.NewProc("CallNtPowerInformation")
+	procGetActiveProcessorGroupCount     = common.Modkernel32.NewProc("GetActiveProcessorGroupCount")
 )
 
 type win32_Processor struct { //nolint:revive //FIXME
@@ -263,6 +264,21 @@ func perCPUTimes() ([]TimesStat, error) {
 
 // makes call to Windows API function to retrieve performance information for each core
 func perfInfo() ([]win32_SystemProcessorPerformanceInformation, error) {
+	// On hosts with more than 64 logical CPUs Windows splits CPUs into Processor Groups
+	// (up to 64 logical CPUs per group). The non-Ex NtQuerySystemInformation only returns
+	// data for the calling thread's group, so whenever the Ex variant is available we
+	// iterate every active group and concatenate the results. See issue #887.
+	if common.ProcNtQuerySystemInformationEx.Find() == nil {
+		return perfInfoAllGroups()
+	}
+	return perfInfoSingleGroup()
+}
+
+// perfInfoSingleGroup queries SystemProcessorPerformanceInformation via the non-Ex
+// NtQuerySystemInformation call. This is the legacy fallback for environments where
+// NtQuerySystemInformationEx cannot be resolved; it only returns data for the calling
+// thread's processor group.
+func perfInfoSingleGroup() ([]win32_SystemProcessorPerformanceInformation, error) {
 	// Make maxResults large for safety.
 	// We can't invoke the api call with a results array that's too small.
 	// If we have more than 2056 cores on a single host, then it's probably the future.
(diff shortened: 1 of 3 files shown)


woodpeqr:
I was in the middle of trying to write a fix too, but you beat me to it :D thank you for your effort
I'm available to help with the verification, should you need help

shirou:
@woodpeqr No worries about the timing, and thanks for the offer! I was thinking we might need to spin up a large instance with our sponsor funds to verify this — so if you can test it on your 96-CPU machine, that would be a huge help. Much appreciated!

woodpeqr:
@shirou sure thing, I'll test it out first thing on Monday if it isn't too late

woodpeqr:
I've ran this code on a 96-core windows server 2025: 
```
package main

import (
	"time"

	"github.com/shirou/gopsutil/v4/cpu"
)

func main() {
	times, err := cpu.Times(true)
	if err != nil {
		panic(err)
	}
	for i, t := range times {
		println("CPU", i, "User Time:", t.User)
		println("CPU", i, "System Time:", t.System)
		println("CPU", i, "Idle Time:", t.Idle)
		println()
	}

	times, err = cpu.Times(false)
	if err != nil {
		panic(err)
	}
	println("Total User Time:", times[0].User)
	println("Total System Time:", times[0].System)
	println("Total Idle Time:", times[0].Idle)

	percent, err := cpu.Percent(1 * time.Second, true)
	if err != nil {
		panic(err)
	}
	for i, p := range percent {
		println("CPU", i, "Usage Percent:", p)
	}
}
```
with the old code code in master branch and with the new code in this PR: 
[old.txt](https://github.com/user-attachments/files/27118435/old.txt)
[new.txt](https://github.com/user-attachments/files/27118436/new.txt)

Everything is looking good! The new code correctly identifies all CPUs and the utilization when `percpu==true`, otherwise the values are comparable between old and new when `percpu==false` as is expected

shirou:
@woodpeqr 
Great! Thank you very much for confirming this on a 64-CPU instance — that’s very helpful.

PR #2083, opened 2026-04-28
Implement net ProtoCounters for AIX

pgimalac:
Implement `net.ProtoCounters()` on AIX, by parsing `netstat -s` output, similar to other functions / platforms.

Tested on AIX 7.2 TL2 and AIX 7.3 TL2.

NB: I think this implementation of `ProtoCounters` parsing `netstat -s` could be reused for other platforms, but I would need to get hosts to validate so for now just AIX is a good start (and what I'm really interested in)

### DIFF
--- a/net/net_aix.go
+++ b/net/net_aix.go
@@ -31,8 +31,240 @@ func ConntrackStatsWithContext(_ context.Context, _ bool) ([]ConntrackStat, erro
 	return nil, common.ErrNotImplementedError
 }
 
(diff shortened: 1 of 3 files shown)


shirou (approved):
LGTM. Nice work — the per-protocol key functions keep the mapping readable, the switch ordering correctly prioritizes the specific cases (e.g. "fragments dropped after timeout" before "fragments dropped"), and the fixture covers the accumulation cases well.

Issue #788, opened 2019-11-13
Expose super block mount options in disk.PartitionStat


danielnelson:
**Is your feature request related to a problem? Please describe.**

The effective read/write mode of a disk mount point is write only if both the mount options and the super block options are both `rw`.  Since we currently only expose the mount options, it is not possible to determine if a partition is writable.

From `man 2 mount`:

> Since Linux 2.6.16, MS_RDONLY can be set or cleared on a per-mount-
> point basis as well as on the underlying filesystem superblock. The
> mounted filesystem will be writable only if neither the filesystem nor
> the mountpoint are flagged as read-only.

**Describe the solution you'd like**

Unsure, perhaps we should expose the super block options in the `disk.PartitionStat` type?  Or, perhaps a more OS agnostic option specifically for `ro` vs `rw`?

**Describe alternatives you've considered**

None

**Additional context**

Here is an example of a partition mounted with mixed mount and super block options:
```
$ cat /proc/self/mountinfo
117 25 0:41 / /backup rw,relatime shared:65 - btrfs /dev/mapper/vg1-backup ro,compress=lzo,space_cache,subvolid=5,subvol=/
```

This is related to https://github.com/influxdata/telegraf/issues/6633

PR #2085, opened 2026-04-30
disk: report effective mount mode on Linux

paulojmdias:
## Summary

Fix Linux `disk.PartitionStat.Opts` mode detection when mountpoint options and superblock options disagree.
Since Linux 2.6.16, `MS_RDONLY` can be set or cleared independently on the mount point and on the underlying filesystem superblock. A mounted filesystem is writable only when both option sets are read-write.
gopsutil currently parses only the mountpoint options from `/proc/*/mountinfo`, so it can report `rw` even when the superblock is `ro`.

## Changes
- Parse Linux `mountinfo` superblock options.
- Replace the `ro`/`rw` option in `PartitionStat.Opts` with the effective mode.
- Treat the mount as `ro` if either the mountpoint options or superblock options contain `ro`.
- Treat the mount as `rw` only when both option sets contain `rw`.
- Add a regression test covering mixed mountpoint/superblock option cases.

## Reproduction
This can be reproduced without corrupting a filesystem by creating a loopback ext4 mount where the mountpoint/VFS options are `rw`, but the filesystem superblock options are `ro`.

```sh
sudo dd if=/dev/zero of=/tmp/gopsutil-disk-test.img bs=1M count=128
sudo mkfs.ext4 -F /tmp/gopsutil-disk-test.img
sudo mkdir -p /mnt/gopsutil-disk-test
sudo mount -o loop,rw /tmp/gopsutil-disk-test.img /mnt/gopsutil-disk-test

target=/mnt/gopsutil-disk-test

sudo mount -o remount,rw "$target"
sudo mount -o remount,ro "$target"
sudo mount -o remount,bind,rw "$target"

findmnt -no TARGET,SOURCE,FSTYPE,VFS-OPTIONS,FS-OPTIONS "$target"
grep " $target " /proc/self/mountinfo
grep " $target " /proc/mounts
```

Expected output shows the mismatch:

```sh
/mnt/gopsutil-disk-test /dev/loop0 ext4 rw,relatime ro
... /mnt/gopsutil-disk-test rw,relatime ... - ext4 /dev/loop0 ro
/dev/loop0 /mnt/gopsutil-disk-test ext4 ro,relatime 0 0
```

Writes fail because the effective mount state is read-only:

```sh
touch "$target/write-should-fail"
touch: cannot touch '/mnt/gopsutil-disk-test/write-should-fail': Read-only file system
```

## Impact

Consumers are relying on `PartitionStat.Opts` to determine read/write mode can currently make the wrong decision for mounts where the mountpoint is `rw` but the superblock is `ro`.

One example is Telegraf's disk input, which emits a mode tag from `PartitionStat.Opts`. In this state, Telegraf can emit `mode="rw"` while `/proc/mounts` and actual write attempts show the filesystem is read-only.

## Notes

This change keeps the existing PartitionStat API and avoids appending duplicate conflicting options, such as both rw and ro.

Relates: #788

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -359,6 +359,14 @@ func parseFieldsOnMountinfo(ctx context.Context, lines []string, all bool, filen
 		}
 		fsType := fields[0]
 		mntSrc := fields[1]
+		superOpts := []string{}
+		if len(fields) >= 3 {
+			superOpts = strings.Split(fields[2], ",")
+		}
+		// Since Linux 2.6.16, read-only can be set independently on the
+		// mount point and the filesystem superblock. The mount is writable
+		// only when both option sets are read-write.
+		mountOpts = effectiveMountOpts(mountOpts, superOpts)
 		isBind := false
 		// Per fstab(5), the device can be any string for non-storage-backed filesystems.
 		if !all && !strings.HasPrefix(mntSrc, "/") {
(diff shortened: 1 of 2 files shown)


disk/disk_linux.go
```diff
@@ -421,6 +429,49 @@ func parseFieldsOnMountinfo(ctx context.Context, lines []string, all bool, filen
...
+
+func containsMountOpt(opts []string, opt string) bool {
+	for _, o := range opts {
+		if o == opt {
+			return true
+		}
+	}
+	return false
+}
+
```
shirou: Could you use `slices.Contains` instead of `containsMountOpt`? It was introduced in Go 1.21.

shirou (changes requested):
LGTM. I left a nit comment. I'll merge this PR after the update. Thank you for your contribution!

disk/disk_linux.go
paulojmdias: Totally agree, thanks for the heads up 🙏 

Already pushed the simplification, PTAL

shirou (approved):
Thank you for the update!

PR #2092, opened 2026-05-16
[process]: fix envp leak into Cmdline on darwin

### LINKED ISSUE #2082: env leak in darwin Cmdline()
**Describe the bug**
return program env when calling Cmdline() in darwin
since `func (p *Process) cmdlineSlice() ([]string, error)` is flawed

Image

**To Reproduce**
```go
package main

import (
	"encoding/hex"
	"fmt"

	"github.com/shirou/gopsutil/v4/process"
	"golang.org/x/sys/unix"
)

func main() {
	pid := int32(xxxx)
	p, err := process.NewProcess(pid)
	if err != nil {
		panic(err)
	}
	fmt.Println(p.Cmdline())

	procargs, _ := unix.SysctlRaw("kern.procargs2", int(pid))
	fmt.Println(hex.Dump(procargs))
}
```

**Expected behavior**
Should not print env in Cmdline() call

ruokeqx@ruokeqx-mini ~/s/d/gopsutil> sw_vers
ProductName:		macOS
ProductVersion:		26.4.1
BuildVersion:		25E253
ruokeqx@ruokeqx-mini ~/s/d/gopsutil> uname -a
Darwin ruokeqx-mini.local 25.4.0 Darwin Kernel Version 25.4.0: Thu Mar 19 19:31:09 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T8132 arm64

**Additional context**

kerlenton:
On darwin, `Process.Cmdline()` could leak the first environment
variable of the target process when its argv contained a legitimate
empty element.

`cmdlineSlice` skipped empty entries from `bytes.Split` without
decrementing `nargs` — needed for padding NULs between `exec_path` and
`argv[0]`, but it also skipped real empty argv elements, causing the
loop to consume one chunk past `argv[nargs-1]` (i.e. `envp[0]`).

Extracted parsing into a pure `parseCmdline` helper that skips padding
only until `argv[0]`, then consumes exactly `nargs` chunks including
empties. Added table-driven unit tests covering the leak scenario,
padding variations and bogus `nargs` values.

Fixes #2082

### DIFF
--- a/process/process_darwin.go
+++ b/process/process_darwin.go
@@ -375,29 +375,38 @@ func (p *Process) cmdlineSlice() ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	// The first bytes hold the nargs int, skip it.
-	args := bytes.Split((pargs)[unsafe.Sizeof(int(0)):], []byte{0})
-	var argStr string
-	// The first element is the actual binary/command path.
-	// command := args[0]
-	var argSlice []string
-	// var envSlice []string
-	// All other, non-zero elements are arguments. The first "nargs" elements
-	// are the arguments. Everything else in the slice is then the environment
-	// of the process.
-	for _, arg := range args[1:] {
-		argStr = string(arg)
-		if argStr != "" {
-			if nargs > 0 {
-				argSlice = append(argSlice, argStr)
-				nargs--
-				continue
-			}
-			break
-			// envSlice = append(envSlice, argStr)
-		}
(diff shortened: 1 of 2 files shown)


shirou (approved):
LGTM !

Nice fix. This correctly addresses the root cause: the old loop skipped empty argv elements without decrementing nargs, so each empty arg consumed one extra envp entry and leaked it into the output. Extracting `parseCmdline` and consuming exactly nargs chunks (empties included) is the right approach.

PR #2093, opened 2026-05-19
refactor: replace Split in loops with more efficient SplitSeq

cuoguojida:
strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.

More info: https://github.com/golang/go/issues/61901

### DIFF
--- a/cpu/cpu_darwin.go
+++ b/cpu/cpu_darwin.go
@@ -92,19 +92,19 @@ func InfoWithContext(_ context.Context) ([]InfoStat, error) {
 	c.Stepping = int32(stepping)
 	features, err := unix.Sysctl("machdep.cpu.features")
 	if err == nil {
-		for _, v := range strings.Fields(features) {
+		for v := range strings.FieldsSeq(features) {
 			c.Flags = append(c.Flags, strings.ToLower(v))
 		}
 	}
 	leaf7Features, err := unix.Sysctl("machdep.cpu.leaf7_features")
 	if err == nil {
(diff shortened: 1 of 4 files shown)


shirou:
Thanks for the PR and for the clear explanation.

After some thought, I'm going to close this one. A couple of reasons:

- The affected paths here all parse short strings from `sysctl` / external command output, so they aren't hot paths. The allocation savings are real but quite small in practice, so the benefit is limited.
- More importantly, `strings.SplitSeq` / `strings.FieldsSeq` were added in Go 1.24, and we don't currently use any of these `*Seq` APIs anywhere in the codebase. While our `go.mod` does say go 1.24.0, that's a constraint inherited from `golang.org/x/sys` rather than a deliberate choice on our side — we'd actually like to keep gopsutil buildable on older Go versions as much as we reasonably can. Introducing a new, direct dependency on the latest Go minor's APIs cuts against that, so I'd prefer to avoid it unless there's a stronger reason.

That said, I'm not opposed to the direction in general. If we do adopt it, I'd rather do it consistently across all platforms (linux, windows, freebsd, etc.) in one pass rather than just darwin — otherwise the codebase ends up with a mix of styles. Once we're comfortable relying on these APIs, a sweep across all platforms would be very welcome.

Thanks again for taking the time to put this together!

PR #2097, opened 2026-05-29
fix(mem): map correct PageFile fields for Windows VirtualMemory calcu…

### LINKED ISSUE #1875: `VirtualMemory` on Windows reports physical instead of virtual
**Describe the bug**
When getting `VirtualMemory` on Linux, `gopsutil` uses `/proc/meminfo`'s `MemTotal` field. This is the total userspace-accessible virtual memory. However, calling the same function on Windows reports `ullTotalPhys` from [`struct MEMORYSTATUSEX`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-memorystatusex) instead of `ullTotalVirtual`. This means the value is semantically different across systems. Why was `ullTotalPhys` used? Should it be switched? Considering the Go struct is called `VirtualMemory` it is strange for physical memory to be used on one platform unlike the rest.

**To Reproduce**
Read https://github.com/shirou/gopsutil/blob/d0dd5e027c0ff389b0c9029cdf4dca77588be177/mem/mem_windows.go#L51

**Expected behavior**
The `VirtualMemory` struct should report virtual memory on all platforms.

Any version of Linux or Windows will report that semantics described.

HarshalPatel1972:
Fixes #1875

## Problem
On Windows systems, `VirtualMemoryWithContext` incorrectly populates the `Total`, `Available`, and `Free` metrics using physical RAM bytes (`ullTotalPhys` / `ullAvailPhys`) instead of reporting the actual system virtual memory space.

## Fix
Updated `mem_windows.go` to correctly map fields from the `GlobalMemoryStatusEx` Win32 API structure:
- Swapped `ullTotalPhys` out for `ullTotalPageFile`.
- Swapped `ullAvailPhys` out for `ullAvailPageFile`.
- Adjusted the `UsedPercent` calculation to dynamically evaluate based on total virtual allocations instead of reading the physical hardware load scalar.

## Tested on
Windows 11 (Cross-compiled and verified via `go test ./mem/... -v`).

### DIFF
--- a/mem/mem_windows.go
+++ b/mem/mem_windows.go
@@ -48,13 +48,15 @@ func VirtualMemoryWithContext(_ context.Context) (*VirtualMemoryStat, error) {
 	}
 
 	ret := &VirtualMemoryStat{
-		Total:       memInfo.ullTotalPhys,
-		Available:   memInfo.ullAvailPhys,
-		Free:        memInfo.ullAvailPhys,
-		UsedPercent: float64(memInfo.dwMemoryLoad),
+		Total:     memInfo.ullTotalPageFile,
+		Available: memInfo.ullAvailPageFile,
+		Free:      memInfo.ullAvailPageFile,
 	}
 
 	ret.Used = ret.Total - ret.Available
+	if ret.Total > 0 {
+		ret.UsedPercent = float64(ret.Used) / float64(ret.Total) * 100.0
+	}
 	return ret, nil
 }
 


shirou:
Thanks for the PR, but I'll decline this one — the current behavior is correct and this would be a regression.

`VirtualMemory` is meant to report physical RAM on all platforms (see the field docs in `mem/mem.go`: "Total amount of RAM on this system"). Linux/Darwin/BSD all do this, and Windows' current `ullTotalPhys` matches.

psutil — which gopsutil origin — reports the same thing (physical RAM). Its Windows `virtual_memory()` reports physical RAM:
https://github.com/giampaolo/psutil/blob/master/psutil/_pswindows.py#L125-L136

`ullTotalPageFile` is not page-file size or virtual address space — it's the system commit limit (physical + page file). In psutil that belongs to `swap_memory()`, not `virtual_memory()`:
https://github.com/giampaolo/psutil/blob/master/psutil/_pswindows.py#L139-L164

gopsutil's existing `SwapMemory` already implements that same commitLimit - physical logic, so this info is already exposed.

Also: #1875 asked for `ullTotalVirtual,` which this PR doesn't use (and which would also be wrong — it's per-process address space, ~128 TB). And switching `Total` would be a silent breaking change for Windows users relying on it being physical RAM.

I'll close this and add a note to #1875. Thanks for digging into it.

HarshalPatel1972:
Thank you for the detailed breakdown, @shirou! That makes total sense. I didn't realize VirtualMemory was intentionally locked to physical RAM metrics across all platforms to maintain cross-package consistency with python's psutil. I completely agree that changing this would introduce an unwanted breaking change for Windows users, and that SwapMemory is the proper place for the commit limit. Thanks for taking the time to clarify the architecture!

Issue #2005, opened 2026-02-18
Various disk mountinfo parsing bugs on Linux introduced by #1931


pgimalac:
**Describe the bug**
https://github.com/shirou/gopsutil/pull/1931 introduced a lot of changes in how the `mountinfo` file is parsed, which results in at least 3 bugs (on top of the already fixed https://github.com/shirou/gopsutil/pull/2002), and possibly a 4th one, see below.

1. For a same underlying device, each item will have its `Device` field set to the previous entry's mountpoint. This is because `seenDevIDs` is updated for each line in the file instead of only for the first line for a given device I believe.  
Eg. in my output below, item 7 has `Device=/dev/nvme0n1p3` and `Mountpoint=/tmp` (which makes sense), but then item 8 has `Device=/tmp` and `Mountpoint=/etc/datadog-agent`, item 12 has `Device=/etc/datadog-agent` and `Mountpoint=/etc/passwd`, and finally item 13 has `Device=/etc/passwd` and `Mountpoint=/host/root`.
I fail to see how this cascading `Device` field value could be a desired output, I think they should all have the same device.

2. Some `bind` opts are missing, I think it's only missing for the first occurrences of a given device

3. Maybe a side effect of 1.: some entries now have `bind` when they previously didn't, I think it's another bug due of `seenDevIDs` which mixes up the devices

4. This might be a desired change but I'd rather check (as I disagree with the change): as mentioned above, items of the same device have the `Device` field containing a mount point (except the very first), while before they would all contain a path to the device under `/dev`. I think the previous behavior made more sense, and would argue this should be reverted.

Considering the number of bugs introduced by #1931, I think it should just be reverted, and later re-introduced carefully, piecemeal, and with tests.
Otherwise I can open PRs to fix the issues mentioned above as I think I understand where they come from.

The main question is whether 4. is really on purpose or not.

**To Reproduce**
<details>
<summary>Real `mountinfo` file taken from a node on my infra</summary>

```
1184 1047 0:70 / / ro,relatime - overlay overlay rw,lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.nydus/snapshots/1084/mnt,upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.nydus/snapshots/1499/fs,workdir=/var/lib/containerd/io.containerd.snapshotter.v1.nydus/snapshots/1499/work,xino=off,nouserxattr
[...]

PR #2101, opened 2026-06-07
[disk][linux]: fix missiing sub volume partitions on zfs and btrfs

### LINKED ISSUE #2100: disk_linux: no longer showing zfs or btrfs filesystem stats unless all is set
`PartitionsWithContext` no longer includes any zfs or btrfs fileystems. This is because #1931 overshot it's intend quite a bit.

**Bug 1**:
https://github.com/shirou/gopsutil/blob/d781152c7fcf353e319f19bba54147e58de8990b/disk/disk_linux.go#L373-L375

This filters out all mounts where the mntSrc does not start with a '/'  unless all is set. This is true for virtual many virtual filesystems but importantly it's also true for zfs where mount entries look like this. This likely needs a special case just for zfs.

```
nanachi01/home on /home type zfs (rw,relatime,xattr,posixacl,casesensitive)
```

**Bug 2**
https://github.com/shirou/gopsutil/blob/d781152c7fcf353e319f19bba54147e58de8990b/disk/disk_linux.go#L386-L394

The bind mount detection logic is wrong. For btrfs filesystems field 4 in mountinfo will actually be the subvolume so this logic is filtering out all btrfs subvolumes.

shirou:
Fixes #2100.

Since #1931, `Partitions(false)` on Linux dropped zfs filesystems and btrfs subvolumes:

- **zfs** was skipped because its source isn't a `/`-prefixed path (it's a dataset name like `pool/dataset`).
- **btrfs** subvolumes (`rootDir != "/"`) were misclassified as bind mounts.

## Changes

- Filter by the existing `getFileSystems()` list instead of the source-path heuristic, matching the `/proc/mounts` path.
- Detect btrfs subvolumes vs binds by comparing `rootDir` to the `subvol=` option (subvolume: equal; bind: deeper), so binds are still filtered (#2005).

Added a regression test; verified against a real btrfs subvolume on Linux.

### DIFF
--- a/disk/disk_linux.go
+++ b/disk/disk_linux.go
@@ -293,7 +293,7 @@ func PartitionsWithContext(ctx context.Context, all bool) ([]PartitionStat, erro
 	}
 
 	// use mountinfo
-	ret, err = parseFieldsOnMountinfo(ctx, lines, all, filename)
+	ret, err = parseFieldsOnMountinfo(ctx, lines, all, fs, filename)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing mountinfo file %s: %w", filename, err)
 	}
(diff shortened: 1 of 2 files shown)


buengese:
Sry I forgot reporting back here. I'd been running this version locally and it looks good.

PR #2108, opened 2026-06-30
fix(net): pad GetExtendedTcpTable buffer to prevent GC thrashing on Windows

HarshalPatel1972:
### Problem
Under heavy network load, querying the Windows TCP/UDP tables can trigger a near-infinite `ERROR_INSUFFICIENT_BUFFER` spin-loop. Because the Go wrapper naively allocates the exact required size returned by the Win32 API, any new connections spawned by the OS before the next iteration cause the subsequent API call to fail again. This rapid, tight reallocation cycle completely overwhelms the Go Garbage Collector, mimicking a severe memory leak and causing massive CPU spikes.

### Fix
* Added a 4KB (4096 byte) padding to the requested size before reallocating the buffer in both `getTCPConnections` and `getUDPConnections`. 
* This padding safely absorbs concurrent connection growth between the failed API call and the next iteration, allowing the retry loop to exit cleanly and preventing GC thrashing.

### Verification
* Validated via `go test -v ./net/...`. Struct alignment and parsing remain intact.

### DIFF
--- a/net/net_windows.go
+++ b/net/net_windows.go
@@ -365,7 +365,7 @@ func getTCPConnections(family uint32) ([]ConnectionStat, error) {
 		return nil, errors.New("faimly must be required")
 	}
 
-	for {
+	for retry := 0; retry < 10; retry++ {
 		switch family {
 		case kindTCP4.family:
 			if len(buf) > 0 {
@@ -395,13 +395,12 @@ func getTCPConnections(family uint32) ([]ConnectionStat, error) {
 		if !errors.Is(err, windows.ERROR_INSUFFICIENT_BUFFER) {
 			return nil, err
 		}
+		if retry == 9 {
+			return nil, errors.New("GetExtendedTcpTable: failed to allocate sufficient buffer after 10 retries")
+		}
 
 		// Add 4KB padding to account for concurrent connection growth between calls.
-		const padding uint32 = 4 * 1024
-		if size > ^uint32(0)-padding {
-			return nil, fmt.Errorf("GetExtendedTcpTable required buffer size too large: %d", size)
-		}
-		size += padding
+		size += 4096
 		buf = make([]byte, size)
 	}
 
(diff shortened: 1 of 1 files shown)


shirou:
Thanks for the fix! The approach makes sense. One thing:

Could you make the TCP and UDP changes consistent? Copilot's comment about the `uint32` overflow applies to both paths, but the guard was only added on the TCP side. Personally I'd lean toward dropping the guard in both places — a table size near 4 GiB can't realistically happen, and on 32-bit builds make would panic above `MaxInt32` before the guard helps anyway. But if you prefer keeping it, please add the same guard (with the shared constant) to the UDP side too.

Optionally, a small retry limit on the loop might be worth considering, since very heavy churn could in theory still keep it spinning.

HarshalPatel1972:
> Thanks for the fix! The approach makes sense. One thing:
> 
> Could you make the TCP and UDP changes consistent? Copilot's comment about the `uint32` overflow applies to both paths, but the guard was only added on the TCP side. Personally I'd lean toward dropping the guard in both places — a table size near 4 GiB can't realistically happen, and on 32-bit builds make would panic above `MaxInt32` before the guard helps anyway. But if you prefer keeping it, please add the same guard (with the shared constant) to the UDP side too.
> 
> Optionally, a small retry limit on the loop might be worth considering, since very heavy churn could in theory still keep it spinning.

@shirou Thanks for the feedback!

You are completely right about the uint32 overflow guard being overkill. I've just pushed a new commit that rips it out entirely, making the TCP and UDP logic perfectly consistent.

I also took your advice and wrapped both buffer allocation loops with a hard 10-retry limit (for retry := 0; retry < 10; retry++). If the OS manages to churn the TCP/UDP tables so violently that we can't allocate a sufficient buffer after 10 tries, it now cleanly bails out with an error rather than spinning infinitely.

Let me know if this looks good to merge.

shirou (approved):
Thanks for the update. LGTM! I appreciate your contribution!

HarshalPatel1972:
> Thanks for the update. LGTM! I appreciate your contribution!

Thanks for the guidance throughout.

PR #2111, opened 2026-07-07
[cpu][windows]: compute cpu-total times from integer ticks

### LINKED ISSUE #2110: cpu: Times(percpu=false) on Windows can report a decreasing System counter (float64 conversion of GetSystemTimes)
**Describe the bug**

On Windows, `cpu.Times(false)` converts the `GetSystemTimes` counters to `float64` before doing any arithmetic:

https://github.com/shirou/gopsutil/blob/25a6867c93f7fb55b735d27fb6b8c6674035f735/cpu/cpu_windows.go#L125-L130

```go
LOT := float64(0.0000001)
HIT := (LOT * 4294967296.0)
idle := ((HIT * float64(lpIdleTime.DwHighDateTime)) + (LOT * float64(lpIdleTime.DwLowDateTime)))
...
system := (kernel - idle)
```

`idle` and `kernel` are cumulative 100 ns tick counts summed across all logical CPUs, so after a few days of uptime each conversion carries a small rounding error that changes from call to call, and `system = kernel - idle` exposes the difference of two such errors. Even though the underlying Windows counters are monotonic, the `System` (and `Idle`/`User`) values returned by `Times(false)` are not: two consecutive calls can return a `System` value that went backwards even when the true tick delta is exactly 0.

Downstream consumers that compute usage percentages from deltas then produce small negative values. I found this while investigating influxdata/telegraf#18141, where a 20-CPU Windows host reports `usage_system` values like `-4.656612873077393e-10`.

Only the `percpu=false` path is affected. The per-CPU path is already correct because it subtracts the raw integers before converting:

https://github.com/shirou/gopsutil/blob/25a6867c93f7fb55b735d27fb6b8c6674035f735/cpu/cpu_windows.go#L256

```go
System: float64(v.KernelTime-v.IdleTime) / ClocksPerSec,
```

Unlike reports such as #1978, where out-of-range percentages accurately reflect what the kernel reports, here the raw values from the Windows API are monotonic and the non-monotonicity is introduced by the conversion itself, so it cannot be handled by clamping on the caller side.

**To Reproduce**

The following simulates the exact conversion done by `TimesWithContext(percpu=false)` on a mostly idle 20-CPU host, advancing the idle and kernel counters by identical amounts so the true System delta is exactly 0 ticks in every interval:

```go
package main

import "fmt"

// toSec mirrors TimesWithContext(percpu=false) in cpu/cpu_windows.go:
// each FILETIME half is converted to float64 separately, then combined.
func toSec(ticks uint64) float64 {
	lot := float64(0.0000001)
	hit := lot * 4294967296.0
	return (hit * float64(uint32(ticks>>32))) + (lot * float64(uint32(ticks)))
}

func main() {
	const ticksPerSec = 10_000_000 // GetSystemTimes counts 100 ns ticks

	// A mostly idle machine with 20 logical CPUs, up for 3 days.
	// GetSystemTimes returns counters summed across all CPUs.
	idle := uint64(3) * 86400 * 20 * ticksPerSec
	kernel := idle + 720*ticksPerSec // kernel time includes idle time

	// Advance both counters by identical amounts: the true System
	// (= kernel - idle) delta is exactly 0 ticks in every interval.
	delta := uint64(1998 * ticksPerSec / 10) // ~10 s interval x 20 CPUs

	prev := toSec(kernel) - toSec(idle) // System as returned by Times(false)
	decreases, shown := 0, 0
	for i := 0; i < 100000; i++ {
		idle += delta
		kernel += delta
		cur := toSec(kernel) - toSec(idle)
		if cur < prev {
			decreases++
			if shown < 3 {
				fmt.Printf("System went backwards by %.6e s\n", prev-cur)
				shown++
			}
		}
		prev = cur
	}
	fmt.Printf("%d of 100000 intervals report a decreasing System counter\n", decreases)
}
```

Output (amd64):

```text
System went backwards by 9.313226e-10 s
System went backwards by 9.313226e-10 s
System went backwards by 9.313226e-10 s
27674 of 100000 intervals report a decreasing System counter
```

**Expected behavior**

The values returned by `Times(false)` should be monotonically non-decreasing, like the underlying Windows counters. Doing the arithmetic on integers and converting once, mirroring the per-CPU path, guarantees this:

```go
idleT := uint64(lpIdleTime.DwHighDateTime)<<32 | uint64(lpIdleTime.DwLowDateTime)
userT := uint64(lpUserTime.DwHighDateTime)<<32 | uint64(lpUserTime.DwLowDateTime)
kernT := uint64(lpKernelTime.DwHighDateTime)<<32 | uint64(lpKernelTime.DwLowDateTime)

ret = append(ret, TimesStat{
	CPU:    "cpu-total",
	Idle:   float64(idleT) / ClocksPerSec,
	User:   float64(userT) / ClocksPerSec,
	System: float64(kernT-idleT) / ClocksPerSec,
})
```

Happy to send a PR.

- [x] Windows: reported on Windows Server (20 logical CPUs) via influxdata/telegraf#18141; the arithmetic reproduces on any platform

**Additional context**

Affects all released versions including v4.26.5; the conversion has been unchanged since it was introduced.

skartikey:
Fixes #2110.

`Times(percpu=false)` converts each FILETIME half of the `GetSystemTimes` counters to float64 separately and then computes `system = kernel - idle` in floating point. On machines with a few days of uptime the cumulative counters are large enough that the conversion carries a rounding error of around 2^-30 s, and the error changes from call to call. As a result the returned `System`, `Idle` and `User` values are not monotonic even though the underlying Windows counters are, and consumers computing usage deltas occasionally see small negative values (see influxdata/telegraf#18141).

This change reassembles the full 64-bit tick counts from the FILETIME halves, does the subtraction on integers, and converts to seconds once, the same way `perCPUTimes` already does it. Since integer-to-float conversion and division by a constant are monotone, the reported values can no longer appear to decrease.

Running the simulation from #2110 against the new conversion shows 0 phantom decreases in 100000 intervals, versus 27674 with the current code.

### DIFF
--- a/cpu/cpu_windows.go
+++ b/cpu/cpu_windows.go
@@ -122,18 +122,19 @@ func TimesWithContext(_ context.Context, percpu bool) ([]TimesStat, error) {
 		return nil, err
 	}
 
-	LOT := float64(0.0000001)
-	HIT := (LOT * 4294967296.0)
-	idle := ((HIT * float64(lpIdleTime.DwHighDateTime)) + (LOT * float64(lpIdleTime.DwLowDateTime)))
-	user := ((HIT * float64(lpUserTime.DwHighDateTime)) + (LOT * float64(lpUserTime.DwLowDateTime)))
-	kernel := ((HIT * float64(lpKernelTime.DwHighDateTime)) + (LOT * float64(lpKernelTime.DwLowDateTime)))
-	system := (kernel - idle)
+	// Do all arithmetic on the integer tick counts and convert to float64
+	// only once, mirroring perCPUTimes. Converting each FILETIME half to
+	// float64 first loses precision on large counters and can make the
+	// returned values non-monotonic. See issue #2110.
+	idle := uint64(lpIdleTime.DwHighDateTime)<<32 | uint64(lpIdleTime.DwLowDateTime)
+	user := uint64(lpUserTime.DwHighDateTime)<<32 | uint64(lpUserTime.DwLowDateTime)
+	kernel := uint64(lpKernelTime.DwHighDateTime)<<32 | uint64(lpKernelTime.DwLowDateTime)
 
 	ret = append(ret, TimesStat{
 		CPU:    "cpu-total",
-		Idle:   float64(idle),
-		User:   float64(user),
-		System: float64(system),
+		Idle:   float64(idle) / ClocksPerSec,
+		User:   float64(user) / ClocksPerSec,
+		System: float64(kernel-idle) / ClocksPerSec, // kernel time includes idle time
 	})
 	return ret, nil
 }


shirou (approved):
Thanks for the detailed issue report and the reproducer in #2110.

The fix looks correct to me:

- Reassembling the FILETIME halves and doing the subtraction on integers is exactly what `perCPUTimes` already does, so both paths are now consistent.
- Monotonicity holds by construction: integer-to-float64 conversion and division by a positive constant are both monotone non-decreasing, so the reported values can no longer appear to decrease while the underlying counters don't.
- `kernel - idle` is now exact, so an interval whose true delta is 0 reports exactly 0 instead of accumulated rounding noise.

`kernel >= idle` is guaranteed by the `GetSystemTimes` contract (kernel time includes idletime) described in MSDN, so the subtraction is safe.

LGTM. Thanks!

PR #2116, opened 2026-07-13
fix(net): handle short Darwin netstat output

### LINKED ISSUE #2115: [Darwin] IOCountersWithContext panics on short netstat output
**Describe the bug**

  `IOCountersWithContext` panics on macOS when `netstat` returns a line with fewer than three fields.

  **To Reproduce**

  ```go
  package main

  import (
        "context"
        "fmt"

        "github.com/shirou/gopsutil/v4/net"
  )

  func main() {
        counters, err := net.IOCountersWithContext(context.Background(), false)
        fmt.Printf("counters=%v err=%v\n", counters, err)
  }
  ```

  **Expected behavior**

  Malformed or unexpected `netstat` output should return an error instead of panicking.

  - [ ] Windows:
  - [ ] Linux:
  - [x] Mac OS:
    ```text
    ProductName:        macOS
    ProductVersion:     26.5.2
    BuildVersion:       25F84

    Darwin MacBookPro.fritz.box 25.5.0 Darwin Kernel Version 25.5.0: Tue Jun 9 22:26:23 PDT 2026; root:xnu-12377.121.10~1/RELEASE_X86_64 x86_64
    ```
  - [ ] FreeBSD:
  - [ ] OpenBSD:

  **Additional context**

  gopsutil version: `v4.26.4`

  Go version: `go1.26.5 darwin/amd64`

  `GOOS=darwin`, `GOARCH=amd64`, `CGO_ENABLED=1`

  Not cross-compiling.

  The panic was first observed through OpenTelemetry host instrumentation, which calls `net.IOCountersWithContext`.

  The failing code accesses `columns[2]` before checking `len(columns)`.

alexfalkowski:
Fixes #2115.

## Summary

On macOS, `netstat -ibdnW` can succeed while emitting sandbox diagnostics to stderr. gopsutil combines stdout and stderr, so a standalone `netstat:` line was passed to `parseNetstatLine`, which indexed `columns[2]` before checking the field count and panicked.

This change:

- validates short records before accessing parsed fields;
- locates the actual `Name ...` table header so prefixed diagnostics are ignored;
- returns parsing errors for malformed records instead of panicking; and
- adds regression coverage for short lines and sandbox diagnostics.

The behavior and output format for valid netstat data are unchanged. The focused Darwin parser tests pass, and the Darwin target cross-compiles successfully. A full local network test run cannot produce per-interface counters because this sandbox denies the underlying netstat interface-data query; that is the environment that motivated the issue.

## Reviewer notes

The header search is intentional: `Invoke` combines stdout and stderr, and macOS may write permission diagnostics before the valid table even when netstat exits with status 0. The parser still rejects malformed records after the table header, preserving error visibility without allowing malformed input to cause a panic.

### DIFF
--- a/net/net_darwin.go
+++ b/net/net_darwin.go
@@ -24,12 +24,23 @@ const endOfLine = "\n"
 
 func parseNetstatLine(line string) (stat *IOCountersStat, linkID *uint, err error) {
 	var (
-		numericValue uint64
-		columns      = strings.Fields(line)
+		numericValue  uint64
+		columns       = strings.Fields(line)
+		numberColumns = len(columns)
 	)
 
-	if columns[0] == "Name" {
+	if numberColumns > 0 && columns[0] == "Name" {
 		return nil, nil, errNetstatHeader
+	}
+
+	if numberColumns < 11 || numberColumns > 13 {
+		return nil, nil, fmt.Errorf("line %q has an invalid number of columns: %d", line, numberColumns)
+	}
+
+	base := 1
+	// sometimes Address is omitted
+	if numberColumns < 12 {
+		base = 0
 	}
 
 	// try to extract the numeric value from <Link#123>
@@ -40,16 +51,6 @@ func parseNetstatLine(line string) (stat *IOCountersStat, linkID *uint, err erro
 		}
 		linkIDUint := uint(numericValue)
 		linkID = &linkIDUint
-	}
-
-	base := 1
-	numberColumns := len(columns)
-	// sometimes Address is omitted
-	if numberColumns < 12 {
-		base = 0
-	}
-	if numberColumns < 11 || numberColumns > 13 {
-		return nil, nil, fmt.Errorf("line %q do have an invalid number of columns %d", line, numberColumns)
 	}
 
 	parsed := make([]uint64, 0, 7)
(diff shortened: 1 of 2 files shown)


shirou (changes requested):
Good catch! I agree the problem exists and the approach in this PR is correct. Could you fix a few things before we merge?
1. This PR only handles the case where the stderr output comes *before* the stdout output. Since both streams are written into the same buffer, stderr can also appear *after* it. So I think we should simply skip malformed lines instead of relying on their position.
2. The `len(columns) < 3` check looks redundant because we already have the `numberColumns < 11 || numberColumns > 13` check. Could you move that existing check above the `columns[2]` access instead? Note that a guard for an empty line (`len(columns) == 0`) is still needed for the `columns[0] == "Name"` check.
3. The test doesn't reflect the actual output reported in #2115. Could you use those actual lines in the test?

alexfalkowski:
@shirou - Apologies for the delay. I did not see any notifications. I have applied the changes.

shirou (approved):
Thank you for your update — CI is green, and dropping the `error` return made the parser much cleaner. LGTM!

For the record (no change needed): the PR description still describes the earlier approach — the `Name ...` header search, and "returns parsing errors for malformed records". What actually landed is: the column count is validated before any field is touched, and any line that fails to parse is skipped wherever it appears. Note that this also silently drops rows with a valid shape but a bad number (`strconv.ParseUint` failure); that's an acceptable trade-off since gopsutil has no logger.

alexfalkowski:
Thanks for merging this. Appreciate the feedback.

Issue #1832, opened 2025-04-03
Occational crash when using sensors on MacOS arm64


PierreF:
**Describe the bug**

I got some random crash when calling SensorsTemperatures() on MacOS (24.3.0 on arm64).

The error seems to be some kind of race-condition, since it only occur when multiple threads are calling SensorsTemperatures concurrently.

On real usage, I *think* my true issue is with concurrent call between SensorsTemperatures and another usage of ioKit and/or corefoundation. But I was not able to have reproducible code sample that only code one call per system (sensors, disk, cpu, mem...)

**To Reproduce**
```go
package main

import (
        "log/slog"
	"sync"

	"github.com/shirou/gopsutil/v4/sensors"
)

func main() {
	var wg sync.WaitGroup

	for range 30 { // The higher is this number, the more likely issue will occur. Empirically 30 seems a good value
		wg.Add(1)

		go func() {
			defer wg.Done()
			r, err := sensors.SensorsTemperatures()

			if false {
				// The log itself isn't required to produce the bug, but without
				// assigning SensorsTemperatures result to variable the bug don't seems to
				// occure, maybe due to compiler optimization ?
				slog.Info("sensors", slog.Any("r", r), slog.Any("err", err))
			}
		}()
	}

	wg.Wait()
}

```

Run the program (possibly multiple time, the race condition seems rather unlikely):
```
go build sensors_bug.go

while ./sensors_bug ; do echo "Sucess"; done 2>&1 | tee large_error_message.log
```

It result in error like:
```
unexpected fault address 0x100921808
fatal error: fault
[signal SIGBUS: bus error code=0x1 addr=0x100921808 pc=0x10092181c]

goroutine 39 gp=0x14000106c40 m=28 mp=0x140000ee008 [running]:
runtime.throw({0x100923b7e?, 0x0?})
	/opt/homebrew/Cellar/go/1.24.2/libexec/src/runtime/panic.go:1101 +0x38 fp=0x14000297a70 sp=0x14000297a40 pc=0x1008d0fe8
runtime.sigpanic()
	/opt/homebrew/Cellar/go/1.24.2/libexec/src/runtime/signal_unix.go:922 +0x170 fp=0x14000297ad0 sp=0x14000297a70 pc=0x1008d2800
github.com/shirou/gopsutil/v4/internal/common.NewLibrary({0x0, 0x0})
[...]

PR #2117, opened 2026-07-16
process: implement Darwin IOCounters via proc_pid_rusage

DavRack:
## Summary
  - Implement `process.IOCounters()` on Darwin using `proc_pid_rusage(RUSAGE_INFO_V2)` from
  `libSystem`
  - Add `SystemLib` bindings for `proc_pid_rusage` and thread-local errno via `__error()`
  - Return cumulative disk I/O bytes in both generic (`ReadBytes`/`WriteBytes`) and
  disk-specific (`DiskReadBytes`/`DiskWriteBytes`) fields
  - Map `EPERM` to `ErrorNotPermitted` and `ESRCH` to `ErrorProcessNotRunning`
  - Update `IOCountersStat` docs and the README platform matrix for macOS `io_counters`
  
 ### Notes:
  - Darwin only exposes disk byte totals; `ReadCount`/`WriteCount` remain `0`
  - Protected system processes may still return `ErrorNotPermitted` without elevated privileges

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ Some code is ported from Ohai. Many thanks.
 |uids                |x      |x        |x        |x      |         |
 |gids                |x      |x        |x        |x      |         |
 |terminal            |x      |x        |x        |       |         |
-|io\_counters        |x      |x        |x        |       |x        |
+|io\_counters        |x      |x        |x        |x      |x        |
 |nice                |x      |x        |x        |x      |x        |
 |num\_fds            |x      |         |         |x      |x        |
 |num\_ctx\_switches  |x      |         |         |       |         |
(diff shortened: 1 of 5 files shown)


shirou (approved):
Thank you for your contribution! Layout, flavor constant and the `unix.EPERM/unix.ESRCH` mapping on error handling all look right — merging as-is.

Heads-up that I'll follow up before the next release:

- `ReadBytes`/`WriteBytes` will go back to 0 on Darwin. On Linux they mean "all I/O incl. cache" (`rchar`), unlike `DiskReadBytes` (`read_bytes`), so filling both from the same figure changes the field's meaning per platform. Your `DiskReadBytes`/`DiskWriteBytes` stay as-is.
- `errno` handling and the per-call `Dlopen`/`Dlclose` — both inherited from `CwdWithContext`, not introduced by this PR. The latter is the same pattern we already fixed for IOKit/CoreFoundation in #1832; libSystem never got it.

PR #2120, opened 2026-07-19
net: fix MIB_IF_ROW2 layout on Windows 386

### LINKED ISSUE #1865: Incorrect Network Metrics on Windows 386 Due to Zeroed pad0for64_4for32 in mibIfRow2
**Describe the bug**
When running on Windows 386 (32-bit) systems, network performance metrics retrieved via gopsutil appear incorrect. The root cause is that the mibIfRow2 struct’s field pad0for64_4for32 always returns zero on 32-bit systems, which leads to miscalculated values such as bandwidth or throughput.

**To Reproduce**
```go
package main

import (
    "fmt"
    "github.com/shirou/gopsutil/v3/net"
)

func main() {
    counters, err := net.IOCounters(true)
    if err != nil {
        panic(err)
    }

    for _, counter := range counters {
        fmt.Printf("%+v\n", counter)
    }
}
```

**Expected behavior**
Network interface statistics (e.g., bytes sent/received) should be correctly reported, including on 32-bit Windows systems

 - Windows: Microsoft Windows [Version 10.0.19045.4291] (32-bit build)

**Additional context**
This issue occurs specifically on Windows 386 builds. Upon inspecting the MibIfRow2 structure in syscall usage, the field pad0for64_4for32 seems to always be zero on 32-bit, which suggests an architecture-specific issue with the struct alignment or access.
Cross-compilation was done using:

GOOS=windows GOARCH=386 go build -o netstat.exe main.go

zhangli091011:
Fixes #1865.

Windows aligns the `ULONG64` fields in `MIB_IF_ROW2` to an 8-byte boundary. Go only gives `uint64` 4-byte alignment on 386, which caused `TransmitLinkSpeed` and every following network counter to be read four bytes early.

This adds four bytes of explicit padding for Windows 386 while retaining the existing layout on other Windows architectures. A deterministic 386 regression test verifies the `TransmitLinkSpeed` offset and total structure size.

## Verification

- Windows amd64 and 386 `go test ./net -count=1` with CI behavior enabled
- Windows amd64 and 386 focused network tests
- Windows amd64 and 386 build and vet
- `git diff --check`

## AI disclosure

OpenCode using `myself/gpt-5.6-sol` assisted with issue investigation, implementation, and test preparation. I reviewed the ABI rationale, source changes, and verification results.

### DIFF
--- a/net/net_windows.go
+++ b/net/net_windows.go
@@ -90,7 +90,6 @@ type guid struct {
 const (
 	maxStringSize        = 256
 	maxPhysAddressLength = 32
-	pad0for64_4for32     = 0
 )
 
 type mibIfRow2 struct {
--- a/net/net_windows_386.go
+++ b/net/net_windows_386.go
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//go:build windows && 386
+
+package net
+
+const pad0for64_4for32 = 4
--- a/net/net_windows_386_test.go
+++ b/net/net_windows_386_test.go
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//go:build windows && 386
+
+package net
+
+import (
+	"testing"
+	"unsafe"
+)
+
+func TestMibIfRow2Layout(t *testing.T) {
+	var row mibIfRow2
+	if offset := unsafe.Offsetof(row.TransmitLinkSpeed); offset != 1192 {
+		t.Errorf("TransmitLinkSpeed offset = %d, want 1192", offset)
+	}
+	if size := unsafe.Sizeof(row); size != 1352 {
+		t.Errorf("mibIfRow2 size = %d, want 1352", size)
+	}
+}
--- a/net/net_windows_not386.go
+++ b/net/net_windows_not386.go
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: BSD-3-Clause
+//go:build windows && !386
+
+package net
+
+const pad0for64_4for32 = 0


zhangli091011:
Would you have a chance to review this PR when convenient? CI is green. Thank you!

shirou (commented):
Thanks for tracking this down — the analysis is correct, and I verified that `TransmitLinkSpeed` lands at offset 1192 with a total size of 1352 on 386 with this padding, matching amd64.

One suggestion: since `unsafe.Alignof` is a constant expression and `net_windows.go` already imports `unsafe`, the padding can be derived directly, without the two build-tagged files:

```go
const (
      maxStringSize        = 256
      maxPhysAddressLength = 32
      // Windows aligns ULONG64 to 8 bytes, but Go only gives uint64 4-byte
      // alignment on 32-bit architectures, so pad ConnectionType up to the
      // next 8-byte boundary there.
      // https://learn.microsoft.com/en-us/windows/win32/api/netioapi/ns-netioapi-mib_if_row2
      pad0for64_4for32 = (8 - unsafe.Alignof(uint64(0))) % 8
)
```

This yields 4 on windows/386 and 0 on windows/amd64 and windows/arm64 (same as today), and it also stays correct for any other 32-bit Windows port, which `!386` would miss.

Would you mind giving it a try?

zhangli091011:
Friendly bump: all CI checks are green and the branch is clean. Could a maintainer review when convenient?

shirou:
@zhangli091011 
Thank you for the ping, and thank you for working on this fix.

The CI status is not what is holding this PR. I left a review on Jul 30 with one change request. I think it may have been missed — the branch is still at `6c1a621` from Jul 19.

Here is the request again:

- Please remove `net/net_windows_386.go` and `net/net_windows_not386.go`.
- In `net/net_windows.go`, please derive the padding from `unsafe.Alignof`:

```go
const (
        maxStringSize        = 256
        maxPhysAddressLength = 32
        // Windows aligns ULONG64 to 8 bytes. Go gives uint64 only 4-byte
        // alignment on 32-bit, so pad ConnectionType up to the next 8-byte
        // boundary there.
        // https://learn.microsoft.com/en-us/windows/win32/api/netioapi/ns-netioapi-mib_if_row2
        pad0for64_4for32 = (8 - unsafe.Alignof(uint64(0))) % 8
)
```

`unsafe` is already imported. I checked the value: 4 on 386, and 0 on amd64 and arm64. The layout is the same as your patch, with one file instead of three.

Please keep `net/net_windows_386_test.go`. It is useful either way.

Could you try this change? If it does not work for some reason, please let me know what you see. And if you would rather not change the PR, that is also fine — please just say so, and I will prepare the change myself.

Thanks again for tracking down this bug.

zhangli091011:
Implemented the requested simplification in 3f0f019:

- Removed `net/net_windows_386.go` and `net/net_windows_not386.go`.
- Derived `pad0for64_4for32` from `unsafe.Alignof(uint64(0))` directly in `net/net_windows.go`.
- Kept `net/net_windows_386_test.go`.

Validated with `go test ./net` on the host architecture and with `GOARCH=386 go test ./net`; both pass. The branch is rebased on current master.

shirou (approved):
Thank you so much. This is exactly what I asked for. LGTM!

PR #2127, opened 2026-07-29
Closes #2126 Implement NtQueryInformationProcess to get commandline Windows 8.1+ 

### LINKED ISSUE #2126: windows proccess.Cmdline() fails on Protected processes ex: (anti cheat protected) due to memory read but NtQueryInformationProcess works.
**Describe the bug**
I was working on a League of Legends LCU client which reads the credentials from the Game Client 
but once in game with the game client closed if you try to get the cmdline from the game it will result an error 
because the PEB memory read is blocked by Vanguard. meanwhile Process hacker/system informer had the command line data with no issues and it wasn't even in admin mode or anything. checked [their github 
](https://github.com/winsiderss/systeminformer/blob/master/phlib/nativeprocess.c#L1820) and found out about `NtQueryInformationProcess`  + `ProcessCommandLineInformation`. this is however a windows 8.1+ undocumented class. 
```
Riot Vanguard is a kernel-level anti-cheat developed by Riot Games.
```

**To Reproduce**
```go
	if err != nil {
		return nil, err
	}
	for _, p := range procs {
		name, _ := p.Name()
		if name != "LeagueClient.exe" {
			continue
		}
		cmdline, err := p.Cmdline() // Fails to read memory because LeagueClient.exe is protected
		if err != nil {
			log.Printf("Err reading League Process: %s", err)
			continue
		}
		return parseClientCmdline(cmdline, p.Pid)
	}
	return nil, errors.New("league client is not running")
```
result: 
```
2026/07/30 01:22:51 Err reading League Process: could not get CommandLine: cannot read process PEB
```
**Expected behavior**
On windows 8.1+ we can use the `NtQueryInformationProcess`  `ProcessCommandLineInformation`
with  only `PROCESS_QUERY_LIMITED_INFORMATION`
to obtain the command line without having to touch the memory. 
 it's also implemented in [ Python Psutil](https://github.com/giampaolo/psutil/pull/1398/) and [System informer](https://github.com/winsiderss/systeminformer/blob/master/phlib/nativeprocess.c#L1820) 

 - [x] Windows: Windows [Version 10.0.19045.7548] (this should work on 8.1+ tho

**Additional context**
I've already implemented and tested a fix for this and I'm willing to open a PR. 
I'm really not so good with opening issues/PRs so please if anything is missing or needs addtional details/fixes/tests please let me know and I'll be more than happy to do so. 

there is also a possible optimization to my code since my current implementation does 2 system calls (to obtain the size first) which may be possible to accomplish in one call in the same manner the systeminformer phlib does. 

Drafted A PR #2127

Ahm3dRN:
Adds `getCmdlineProtected`, which uses  `ProcessCommandLineInformation` ->
`NtQueryInformationProcess` before falling back to the existing PEB memory read
approach in `getProcessCommandLine`. 

If it fails (pre-8.1 version of windows or such) it should fall back to the current implementation of PEB. 
Tested against `LeagueClient.exe` a Vanguard-protected process on Windows 10
this Closes #2126

### DIFF
--- a/process/process_windows.go
+++ b/process/process_windows.go
@@ -6,6 +6,7 @@ package process
 import (
 	"bufio"
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
@@ -48,7 +49,10 @@ var (
 	processorArchitecture uint
 )
 
-const processQueryInformation = windows.PROCESS_QUERY_LIMITED_INFORMATION
+const (
+	processQueryInformation             = windows.PROCESS_QUERY_LIMITED_INFORMATION
+	processCommandLineInformation int32 = 60 // Windows 8.1+ ProcessCommandLineInformation  ntpsapi.h
+)
 
 type systemProcessorInformation struct {
 	ProcessorArchitecture uint16
@@ -1054,48 +1058,6 @@ func getUserProcessParams32(handle windows.Handle) (rtlUserProcessParameters32, 
 	return *(*rtlUserProcessParameters32)(unsafe.Pointer(&buf[0])), nil
 }
 
-func getCmdlineProtected(handle windows.Handle, pid int32) (string, error) {
-	const processCommandLineInformation int32 = 60 // System Informer ntpsapi.h
-	var returnLength uint32
-	// first call is to get the actual returnLength
-	err := windows.NtQueryInformationProcess(
-		handle,
-		processCommandLineInformation,
-		nil,
-		0,
-		&returnLength,
-	)
-	if err != nil && !errors.Is(err, windows.STATUS_INFO_LENGTH_MISMATCH) {
-		return "", fmt.Errorf("querying command line size for pid %d: %w", pid, err)
-	}
-	if returnLength == 0 {
-		return "", fmt.Errorf("process %d returned zero-length command line info", pid)
-	}
-
-	// second call: actually fetch it into a correctly-sized buffer
-	buf := make([]byte, returnLength)
-	err = windows.NtQueryInformationProcess(
-		handle,
-		processCommandLineInformation,
-		unsafe.Pointer(&buf[0]),
-		returnLength,
-		&returnLength,
-	)
-	if err != nil {
-		return "", fmt.Errorf("reading command line for pid %d: %w", pid, err)
-	}
-
-	us := (*windows.NTUnicodeString)(unsafe.Pointer(&buf[0]))
-	if us.Length == 0 {
-		return "", nil
-	}
-	// Compute the offset manually rather than dereferencing us.Buffer
-	// directly — see https://github.com/golang/go/issues/73460
-	strOffset := int(unsafe.Sizeof(windows.NTUnicodeString{}))
-	strEnd := strOffset + int(us.Length)
-	return convertUTF16ToString(buf[strOffset:strEnd]), nil
-}
-
 func getUserProcessParams64(handle windows.Handle) (rtlUserProcessParameters64, error) {
 	pebAddress, err := queryPebAddress(syscall.Handle(handle), false)
 	if err != nil {
@@ -1246,20 +1208,82 @@ func (p *processReader) Read(buf []byte) (int, error) {
 
 func getProcessCommandLine(pid int32) (string, error) {
 	h, err := windows.OpenProcess(processQueryInformation|windows.PROCESS_VM_READ, false, uint32(pid))
-	if errors.Is(err, windows.ERROR_ACCESS_DENIED) || errors.Is(err, windows.ERROR_INVALID_PARAMETER) {
+	if err == nil {
+		defer syscall.CloseHandle(syscall.Handle(h))
+
+		if cmdLine, err := getProcessCommandLinePEB(h); err == nil {
+			return cmdLine, nil
+		}
+
+		// PEB read failed even though the handle opened. VM_READ may
+		// have been granted but ineffective against this process's
+		// protection driver. Fall back to the native query on the same handle.
+		if cmdLine, err := getProcessCommandLineNative(h, pid); err == nil {
+			return cmdLine, nil
+		}
+
 		return "", nil
 	}
-	if err != nil {
+
+	if errors.Is(err, windows.ERROR_INVALID_PARAMETER) {
+		return "", nil
+	}
+	if !errors.Is(err, windows.ERROR_ACCESS_DENIED) {
 		return "", err
 	}
-	defer syscall.CloseHandle(syscall.Handle(h))
 
-	// Try the native command-line query first succeeds on protected
-	// processes where the PEB-read fallback below gets ACCESS_DENIED.
-	if cmdLine, err := getCmdlineProtected(h, pid); err == nil {
-		return cmdLine, nil
+	// fallback in case it's a genuine PPL process, where
+	// PROCESS_VM_READ itself is denied at OpenProcess time. Retry with
+	// just PROCESS_QUERY_LIMITED_INFORMATION, which PPL still grants.
+	lh, lerr := windows.OpenProcess(processQueryInformation, false, uint32(pid))
+	if lerr != nil {
+		return "", nil
+	}
+	defer syscall.CloseHandle(syscall.Handle(lh))
+
+	return getProcessCommandLineNative(lh, pid)
+}
+
+func getProcessCommandLineNative(handle windows.Handle, pid int32) (string, error) {
+	var returnLength uint32
+	// first call is to get the actual returnLength
+	err := windows.NtQueryInformationProcess(
+		handle,
+		processCommandLineInformation,
+		nil,
+		0,
+		&returnLength,
+	)
+	if err != nil &&
+		!errors.Is(err, windows.STATUS_INFO_LENGTH_MISMATCH) &&
+		!errors.Is(err, windows.STATUS_BUFFER_OVERFLOW) &&
+		!errors.Is(err, windows.STATUS_BUFFER_TOO_SMALL) {
+		return "", fmt.Errorf("querying command line size for pid %d: %w", pid, err)
+	}
+	if returnLength == 0 {
+		return "", fmt.Errorf("process %d returned zero-length command line info", pid)
 	}
 
+	// second call: actually fetch it into a correctly-sized buffer
+	buf := make([]byte, returnLength)
+	err = windows.NtQueryInformationProcess(
+		handle,
+		processCommandLineInformation,
+		unsafe.Pointer(&buf[0]),
+		returnLength,
+		&returnLength,
+	)
+	if err != nil {
+		return "", fmt.Errorf("reading command line for pid %d: %w", pid, err)
+	}
+	cmdLine, err := parseCommandLineInformation(buf)
+	if err != nil {
+		return "", fmt.Errorf("parsing command line for pid %d: %w", pid, err)
+	}
+	return cmdLine, nil
+}
+
+func getProcessCommandLinePEB(h windows.Handle) (string, error) {
 	procIs32Bits := is32BitProcess(h)
 
 	if procIs32Bits {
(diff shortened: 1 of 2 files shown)


shirou (changes requested):
Thanks for the thorough issue and for digging up the SystemInformer / psutil references — the direction is right and I'd like to take this. Four things before merging.

**1. Make the native query a fallback, not the primary path.**
Right now it runs first for *every* Windows process, so everyone's cmdline source becomes an undocumented information class. psutil does the opposite — PEB first, `ProcessCommandLineInformation` only on permission errors ([`_pswindows.py`](https://github.com/giampaolo/psutil/blob/master/psutil/_pswindows.py), see `use_peb`) — because a process can be started suspended with its PEB command line patched, and then the PEB value is the one the process actually sees. #2126 only needs the case where the PEB read fails, so a fallback fixes it with zero impact on every other process.

**2. Real protected processes aren't reached yet — please cover them here.**
`getProcessCommandLine` opens with `PROCESS_QUERY_LIMITED_INFORMATION | PROCESS_VM_READ` and returns `("", nil)` on `ACCESS_DENIED`, before `getCmdlineProtected` is ever called. For PPL processes (`services.exe`, `csrss.exe`, `MsMpEng.exe`, …) that `OpenProcess` is exactly what fails, so the PR as it stands only helps the "handle opens, but `ReadProcessMemory` is blocked" case (yours). SystemInformer's doc says `PROCESS_QUERY_LIMITED_INFORMATION` alone suffices on 8.1+, and psutil opens a separate handle with just that. Since the PR is titled "protected processes", I'd like it to actually reach them:

```go
h, err := windows.OpenProcess(processQueryInformation|windows.PROCESS_VM_READ, false, uint32(pid))
if errors.Is(err, windows.ERROR_ACCESS_DENIED) {
    // PPL: VM_READ is refused, but a query-only handle may still open
    if lh, lerr := windows.OpenProcess(processQueryInformation, false, uint32(pid)); lerr == nil {
        defer windows.CloseHandle(lh)
        return getProcessCommandLineNative(lh)
    }
    return "", nil
}
```

**3. Accept the other "buffer too small" statuses.** psutil treats `STATUS_BUFFER_OVERFLOW` and `STATUS_BUFFER_TOO_SMALL` as success alongside `STATUS_INFO_LENGTH_MISMATCH`; all three are in `golang.org/x/sys/windows`. On a machine returning one of the other two, this silently does nothing.

**4. Bounds-check the buffer.** `returnLength == 0` is guarded but `1..15` isn't (`sizeof(NTUnicodeString)` is 16 on amd64, 8 on 386), and `strEnd` isn't checked against `len(buf)` — `buf[strOffset:strEnd]` can panic, which callers can't recover from. Unlikely in practice, but we're trusting an undocumented API's return value. Reading the length via `binary.LittleEndian.Uint16(buf[0:2])` instead of casting to `NTUnicodeString` also sidesteps golang/go#73460 (the issue your comment links to) entirely.

Smaller points:

- Once it's a fallback, CI never exercises this path. Please split the buffer parsing into a pure `func([]byte) (string, error)` and table-test it (valid / zero length / truncated) — that covers 4 too.
- As you noted in the issue, SystemInformer allocates a reasonable buffer up front and only retries on `STATUS_INFO_LENGTH_MISMATCH`, halving the syscalls. Lower priority once this is a fallback.
- `getCmdlineProtected` splits the `getUserProcessParams32`/`64` pair — please move it next to `getProcessCommandLine`. It's a general native query rather than protected-process-specific, so `getProcessCommandLineNative` reads better.
- Move `processCommandLineInformation = 60` next to `processQueryInformation` with a note that it's `ProcessCommandLineInformation` (ntpsapi.h, Windows 8.1+).
- `// Try the native command-line query first succeeds on protected` is missing a clause.

1–4 are what I'd like fixed before merging; the smaller points can follow up if you prefer. Thanks again.

Ahm3dRN:
Thank you so much for the thorough review and for giving my PR time. 

for Points 1 and 2, I've created a new func `getProcessCommandLinePEB` which holds the logic for PEB 
in `getProcessCommandLine` I first try to open a handle with `processQueryInformation|PROCESS_VM_READ` 
If no errors we call the PEB function, If  `ERROR_INVALID_PARAMETER` we return as usual if any other error except for `ERROR_ACCESS_DENIED` we return the error as usual 

then I open a handle with only `processQueryInformation`  or `PROCESS_QUERY_LIMITED_INFORMATION` and call `getProcessCommandLineNative` 

my current approach that considers both cases where a handle is granted but memory read is blocked and a genuine PPL process 

```go
func getProcessCommandLine(pid int32) (string, error) {
	h, err := windows.OpenProcess(processQueryInformation|windows.PROCESS_VM_READ, false, uint32(pid))
	if err == nil {
		defer syscall.CloseHandle(syscall.Handle(h))

		if cmdLine, err := getProcessCommandLinePEB(h); err == nil {
			return cmdLine, nil
		}

		// PEB read failed even though the handle opened. VM_READ may
		// have been granted but ineffective against this process's
		// protection driver. Fall back to the native query on the same handle. 
		if cmdLine, err := getProcessCommandLineNative(h, pid); err == nil {
			return cmdLine, nil
		}

		return "", nil
	}

	if errors.Is(err, windows.ERROR_INVALID_PARAMETER) {
		return "", nil
	}
	if !errors.Is(err, windows.ERROR_ACCESS_DENIED) {
		return "", err
	}

	// fallback in case it's a genuine PPL process, where
	// PROCESS_VM_READ itself is denied at OpenProcess time. Retry with
	// just PROCESS_QUERY_LIMITED_INFORMATION, which PPL still grants.
	lh, lerr := windows.OpenProcess(processQueryInformation, false, uint32(pid))
	if lerr != nil {
		return "", nil
	}
	defer syscall.CloseHandle(syscall.Handle(lh))

	return getProcessCommandLineNative(lh, pid)
}
```
Point 3 Done 
Point 4 
I've added more checks and error return to avoid panic situations and used `binary.LittleEndian.Uint16` as well 
would you like to have the tests in `process_test.go` or a new `process_windows_test.go`? 
```go

func parseCommandLineInformation(buf []byte) (string, error) {
	if len(buf) < 2 {
		return "", errors.New("command line buffer too small to read length")
	}
	length := binary.LittleEndian.Uint16(buf[0:2])
	if length == 0 {
		return "", nil
	}
	strOffset := int(unsafe.Sizeof(windows.NTUnicodeString{}))
	strEnd := strOffset + int(length)
	if strEnd > len(buf) {
		return "", errors.New("command line length exceeds buffer size")
	}
	return convertUTF16ToString(buf[strOffset:strEnd]), nil
}

```
aside from the second smaller point "reasonable buffer allocation"  everything else is ready just waiting your confirmation.

Edit: I've pushed the changes just to save your time but I'm open to any feedback or modifications.

shirou (commented):
Sorry for the slow reply — and thanks for rebasing in the meantime; that red CI was never yours. 1–4 are all addressed, and the split into `getProcessCommandLinePEB` / `getProcessCommandLineNative` / `parseCommandLineInformation` reads well. Three things left.

1. Keep the error behaviour neutral. Two cases changed vs master, in opposite directions:

- Handle opens but PEB fails: master returned the error, now it's `("", nil)`. The message you reported — `could not get CommandLine: cannot read process PEB` — becomes an unexplained empty string whenever the native query doesn't rescue it.
- `ACCESS_DENIED`: master returned `("", nil)`, now the native error propagates. That fires whenever the query-limited handle opens but class 60 has nothing to give back — pid 4, Registry, Memory Compression — so `Cmdline()` starts failing on system PIDs that used to be quietly empty. (Same on pre-8.1 Windows, where the class doesn't exist at all.)

I'd like this PR to only ever turn a failure into a success, never the reverse: keep the PEB error in a variable and `return "", pebErr` when both paths fail, and `return "", nil` when the native query fails on the query-limited handle.

2. `windows.ProcessCommandLineInformation` already exists in `x/sys` — I verified it's 60 and passes to the `int32` parameter as-is. My earlier "move the constant next to `processQueryInformation`" was a miss on my part; please just drop it.

3. The 32-bit offset is the one thing I can't verify. `unsafe.Sizeof(windows.NTUnicodeString{})` is *our* pointer width — 16 on amd64, 8 on 386. If a 386 build gets the 64-bit layout back, it reads the `Buffer` pointer as the start of the string, and the bounds checks can't catch it (`8+Length <= len(buf)` always holds when the buffer is `16+Length`). `UTF16ToString` stops at the pointer's zero high bytes, so we'd return garbage *as a success*. psutil dereferences `Buffer` instead and ships win32 wheels without complaints, so the thunk probably converts — but CI only tests amd64, and this repo handles WOW64 explicitly elsewhere. Could you try a `GOARCH=386` build against `csrss.exe` on 64-bit Windows?

On the tests: `buildBuf` derives `headerSize` from the same `unsafe.Sizeof(windows.NTUnicodeString{})` as the code under test, so it can't catch a wrong offset — both sides move together. It also always sizes the buffer to exactly `header+Length`, so `buf[strOffset:]` (ignoring Length) would pass every case, and real kernel buffers are padded. Now that the native path is a fallback, nothing exercises it end-to-end either — a test asserting `getProcessCommandLineNative` equals `getProcessCommandLinePEB` for the current process would cover the info class, the size negotiation and the offset in one go. Also missing the `// SPDX-License-Identifier: BSD-3-Clause` header, and `process_test.go` uses testify throughout (`testifylint` is enabled).

1 and 2 before merge; 3 is a question. Thanks for sticking with this.

Ahm3dRN:
Hi thanks for taking the time to re-review my code.
for Point 1 and 2 it's clear I added I've returned pebErr once the native pass fails as well for when an open handle fails part. 

```go
cmdLine, pebErr := getProcessCommandLinePEB(h)
		if pebErr == nil {
			return cmdLine, nil
		}
		// PEB read failed even though the handle opened. VM_READ may
		// have been granted but ineffective against this process's
		// protection driver. Fall back to the native query on the same handle.
		if cmdLine, err := getProcessCommandLineNative(h, pid); err == nil {
			return cmdLine, nil
		}
		return "", pebErr
```

ignored the native error and returned 
End of getProcessCommandLine

```go
	cmdLine, _ := getProcessCommandLineNative(lh, pid)
	return cmdLine, nil
```

as for point 3
I've tried a simple test with `GOARCH=386` on my 64-bit windows machine `csrss.exe` returned an empty string with no error 
(System informer also doesn't have a commandline for this process)
so i tried with notepad.exe and explorer.exe, both the 386 and amd64 builds return identical 
command lines. 
```bash
AMD64
go run ./testfeature/feature_example.go
2026/08/23 21:19:38 found process
2026/08/23 21:19:38 found process
2026/08/23 21:19:38 "C:\Windows\system32\notepad.exe"

386
./test386.exe
2026/08/23 21:19:48 found process
2026/08/23 21:19:48 found process
2026/08/23 21:19:48 "C:\Windows\system32\notepad.exe"
```

Ahm3dRN:
as for the second point I've dropped the variable and directed called the class from x/sys 

and regarding the tests I changed `ntUnicodeStringHeaderSize ` a little bit to have a different path to the same value. 
```go
ntUnicodeStringHeaderSize := int(unsafe.Offsetof(windows.NTUnicodeString{}.Buffer)) +
		int(unsafe.Sizeof((*uint16)(nil)))
```
and added tests for `getProcessCommandLinePEB` vs `getProcessCommandLineNative`

shirou (approved):
Merging this. Thanks for the back and forth — four revisions in, the final shape is a good bit better than where it started, and I'm sorry the review rounds were so far apart on my side.

The 32-bit layout question I raised last time is settled: I checked it on my own machine and your offset is right, so there's nothing left for you to do there.

Two small things are left, neither worth another round from you — I'll fold them into a follow-up. A test buffer with trailing padding, since the kernel sizes the buffer by `returnLength` and that isn't necessarily `header + Length`. And making the discarded error in the PPL path explicit rather than a bare `_`, which currently reads the same as an accidental swallow. I'd also like a real windows/386 test job in CI.

**One favour, purely for the record.** Your PR description says you tested against `LeagueClient.exe`, but that was the original native-first implementation, and I can't reproduce it here — I'm not elevated, so even `csrss.exe` wouldn't open with `PROCESS_QUERY_LIMITED_INFORMATION` for me. What I don't know is which branch LeagueClient takes now. If its handle opens and only the PEB read fails, the native query runs on that same `PROCESS_VM_READ` handle, and that's the same call your first version made. If `OpenProcess` is denied outright, it goes down the new path that reopens with `PROCESS_QUERY_LIMITED_INFORMATION` only — a weaker handle, and nothing has confirmed that class 60 still answers on it.

So when you get a chance, could you build master after this merge and post the `Cmdline()` result for `LeagueClient.exe` on #2126? The merge closes that issue automatically, but a comment there still reaches me and I'll reopen if it doesn't hold. No rush, and nothing breaks either way: the worst this path can do is return `("", nil)`, which is exactly what master already returned for these processes, so it can only add command lines and never take one away.

Thanks so much!.

PR #2128, opened 2026-07-31
[cpu][windows]: harden the cpu-total computation added in #2125

shirou:
Follow-up to #2125, which stopped deriving cpu-total from GetSystemTimes and accumulates the per-processor counters instead.

- Fall back to GetSystemTimes for cpu-total when perfInfo() fails. perfInfo() relies on the undocumented NtQuerySystemInformation, so Times(false) would now fail outright where that call is unavailable, even though the public API would still work. The processor-group problem #2125 fixes does not apply to the fallback, which is only taken when the group-aware query is unusable in the first place.
- Return an error instead of an all-zero cpu-total when perfInfo() reports no processors.
- Drop the accumulation of DpcTime and InterruptCount: TimesStat has no field for either, so the sums were never used. The struct fields stay, they are needed for the buffer layout passed to the Windows API.
- Restore the note from #2110 explaining why the tick counts are summed as integers and converted to float64 only once.
- Add TestTimesTotalMatchesPerCPUSum, asserting that cpu-total is the field-wise sum of the per-CPU stats.
- Fix a stale comment in the psutil comparison test: Windows no longer derives cpu-total from GetSystemTimes.

### DIFF
--- a/cpu/cpu_psutil_test.go
+++ b/cpu/cpu_psutil_test.go
@@ -82,8 +82,10 @@ func TestTimes_Against_Psutil(t *testing.T) {
 	require.NoError(t, err)
 
 	// user/system/idle exist on all supported platforms and both sides
-	// derive them identically (linux: /proc/stat; windows: GetSystemTimes
-	// with system = kernel - idle; darwin: mach host statistics).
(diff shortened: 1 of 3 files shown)


srebhan (approved):
LGTM.

Issue #562, opened 2018-07-27
UsageStats.UsedPercent is different than df command output Use%


tomekkolo:
https://github.com/shirou/gopsutil/blob/d9d9688f3d20a6b7942c5222607c29b32b69d15e/disk/disk_unix.go#L53

Current way how UsageStats.UsedPercent is calculated differs from what df command does. UsedPercent in gopsutil is calculated as (used/total)*100, while in df command it is (used/(used+available))*100. 

UsageStats.Free is set as Bavail, and to be consistent UsageStats.UsedPercent should be calculated for available memory.

Lomanic:
We don't care especially about `df` but it's the same formula in python's psutil, so we will calculate this way to be consistent with the original. Thanks for the report @tomekkolo :+1:

tomekkolo:
I've created pull request to address this: https://github.com/shirou/gopsutil/pull/563

shirou:
#563 was merged. Thank you for your report!

PR #2136, opened 2026-08-19
fix(disk/aix): use statfs for UsageWithContext on AIX

pgimalac:
Update `UsageWithContext` on AIX to use the same implementation as other Unix OSes, based on the `statfs` syscall.

- Aligns implementation with other OSes
- Avoids running commands (and parsing the output) in the `CGO_ENABLED=0` case
- Makes `UsageWithContext` work with arbitrary paths. The old implementations (both CGO and no-CGO) only work when `path` is exactly a mountpoint (fails with `mountpoint ... not found` otherwise), while `statfs` works properly no matter what

I confirmed getting the exact same output for real mountpoints, and it now works for non-mountpoints.

### DIFF
--- a/disk/disk_aix.go
+++ b/disk/disk_aix.go
@@ -8,11 +8,26 @@ import (
 	"errors"
 	"strings"
 
+	"golang.org/x/sys/unix"
+
 	"github.com/shirou/gopsutil/v4/internal/common"
 )
 
 func LabelWithContext(_ context.Context, _ string) (string, error) {
 	return "", common.ErrNotImplementedError
+}
+
+// FSType maps AIX filesystem type numbers (as reported by statfs and perfstat)
+// to their string name. Shared by the cgo and nocgo build variants.
+var FSType = map[int]string{
+	0: "jfs2", 1: "namefs", 2: "nfs", 3: "jfs", 5: "cdrom", 6: "proc",
+	16: "special-fs", 17: "cache-fs", 18: "nfs3", 19: "automount-fs", 20: "pool-fs", 32: "vxfs",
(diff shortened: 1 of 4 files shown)


shirou (approved):
Merging this — thank you so much! Nice cleanup, and it's also a return to what we had: until May 2024 `disk_unix.go` was tagged `... || (aix && !cgo)` and AIX nocgo used this same `statfs` path; #1651 swapped in the `df -v` parser with no stated reason.

One thing I'd like on the record here, if you don't mind answering after the fact: AIX's `statfs` has both `f_bsize` and `f_fsize`, and the shared code scales the block counts by `Bsize`. On JFS2 the two are identical, but classic JFS has fragments — does `Blocks * Bsize` still give the right total there? I couldn't reach IBM's docs to check either way. "JFS2 is what matters in practice" is a perfectly good answer — I just want it written down, so whoever hits a wrong `Total` on AIX later has a place to start. Whichever build and filesystem types you checked against would be useful in the same breath.

Two things I'll take care of separately, nothing for you to do:

- The removed cgo `UsageWithContext` fell back to `"unknown"` for unmapped vfstypes and `PartitionsWithContext` still does, so a cgo build now reports `"unknown"` from `Partitions()` and `""` from `Usage()`. `""` matches the other platforms, so I'll leave `getFsType` as is and align `Partitions()` on its own.
- On nocgo the values shift slightly: the old parser read `%Used`/`%Iused` with `ParseInt` (always whole numbers) and took `Free` from df's column rather than `Bavail`. The new ones are better — `Used/(Used+Free)` is the project-wide convention from #562 — I'll note the change in the release notes.

No tests needed; we have no AIX runner. For context, our lint matrix also only builds AIX with `CGO_ENABLED=0` (the cgo row is commented out with a `# FIXME`), so CI doesn't cover the side this PR actually changes. That's our gap, not yours.

PR #2142, opened 2026-08-29
Return an error on empty OpenBSD KinfoProc

### LINKED ISSUE #2130: Panic in Process.Times() on OpenBSD: missing bounds check in callKernProcSyscall
**Describe the bug**

`Process.Times()` panics on OpenBSD with `index out of range [0] with length 0`.
It reproduced on two separate CI runs; the second time two goroutines panicked
at once.

```
panic: runtime error: index out of range [0] with length 0

goroutine 75 [running]:
github.com/shirou/gopsutil/v4/process.callKernProcSyscall(0x1, 0x69a6)
	process/process_openbsd.go:406 +0x26d
github.com/shirou/gopsutil/v4/process.(*Process).getKProc(...)
	process/process_openbsd.go:364 +0x2d
github.com/shirou/gopsutil/v4/process.(*Process).TimesWithContext(...)
	process/process_openbsd.go:275 +0x1b
github.com/shirou/gopsutil/v4/process.(*Process).Times(...)
	process/process.go:518
github.com/shirou/gopsutil/v4/process.TestConcurrent.func1()
	process/process_test.go:768 +0xd6
```

**To Reproduce**

```go
// on OpenBSD, from several goroutines at once (this is what TestConcurrent does)
p, _ := process.NewProcess(int32(os.Getpid()))
_, err := p.Times()
```

`go test ./process/` on OpenBSD. It does not fail on every run.

**Expected behavior**

An error, not a panic. `Process.Times()` is a public API and a caller cannot
guard against this.

**Where it is**

`callKernProcSyscall` sizes the buffer from a first sysctl and then
dereferences it unconditionally:

```go
count := int32(length / uint64(sizeOfKinfoProc))
mib = []int32{CTLKern, KernProc, op, arg, sizeOfKinfoProc, count}
// get proc info itself
buf := make([]byte, length)
_, _, err = unix.Syscall6(
	unix.SYS___SYSCTL,
	uintptr(mibptr),
	uintptr(miblen),
	uintptr(unsafe.Pointer(&buf[0])),   // <- process_openbsd.go:406
```

If the first sysctl returns `length == 0` without an error, `buf` is empty and
`&buf[0]` panics. I have not worked out *why* the length comes back zero, only
that it does under concurrent use.

The guard added in #1694 for this same area lives in `getKProc`:

```go
if length != sizeOfKinfoProc {
	return nil, errors.New("unexpected size of KinfoProc")
}
```

but that runs *after* `callKernProcSyscall` returns, so it cannot prevent the
dereference inside it. `process_freebsd.go` and `process_netbsd.go` do not
have this pattern.

 - [X] OpenBSD: OpenBSD 7.9 amd64, go1.26.2, gopsutil master

**Additional context**

Found while running the test suite on OpenBSD for the first time -- #2129 adds
a FreeBSD job, and OpenBSD is left out of it precisely because of this. The two
runs:

https://github.com/neilpang/gopsutil/actions/runs/30750671511/job/91504049494
https://github.com/neilpang/gopsutil/actions/runs/30751019667/job/91504958976

Same shape as the bounds checks in #2069 / #2070 / #2071 and #2109 -- happy to
send a PR if you would like it fixed the same way.

AshSgDe29071999:
Fixes #2130

`callKernProcSyscall` queries sysctl for the `KinfoProc` length, allocates a buffer of that size, then passes `&buf[0]` into the second sysctl. When the first query returns `length == 0` (observed under concurrent `Process.Times()` on OpenBSD), `buf` is empty and that dereference panics:

```
panic: runtime error: index out of range [0] with length 0
```

The size check added in #1694 runs only after this function returns, so it cannot prevent the panic. Return an error instead of indexing an empty buffer. Callers already handle errors from `Times()`.

### DIFF
--- a/process/process_openbsd.go
+++ b/process/process_openbsd.go
@@ -392,6 +392,9 @@ func callKernProcSyscall(op, arg int32) ([]byte, uint64, error) {
 	if err != 0 {
 		return nil, length, err
 	}
+	if length == 0 {
+		return nil, 0, errors.New("empty KinfoProc")
+	}
 
 	count := int32(length / uint64(sizeOfKinfoProc))
 	mib = []int32{CTLKern, KernProc, op, arg, sizeOfKinfoProc, count}


shirou (approved):
Thank you!. This looks good for me.

The guard is in the right place. `getKProc` has the `length != sizeOfKinfoProc` check from #1694, but `ProcessesWithContext` does not call `getKProc`. So the check must be inside `callKernProcSyscall`, before `make([]byte, length)`. That is what this patch does.

Issue #1061, opened 2021-04-19
Docker package won't pull stats for ecs EC2s and custom cgroups


jayar95:
**Describe the bug**
EC2 host machines for amazon ECS store docker stats about containers in a way that deviates from the norm. 

For instance, ecs container stats are stored in `/sys/fs/cgroup/memory/system.slice/var-lib-docker-containers-<containerid>-mounts-shm.mount/` as apposed to `/sys/fs/cgroup/memory/docker/<containerid>/`

**To Reproduce**
```go
memory, err := docker.CgroupMemDocker(container.ContainerID)
// err: will tell you it couldn't find the memory.stat file in `fs/cgroup/memory/system.slice/docker-<containerID>.scope" (because this dir doesn't exist)
```

**Expected behavior/suggested implementation**
Should produce mem/cpu stats when invoked

Consider this method in docker/docker_linux.go:

```go
// getCgroupFilePath constructs file path to get targeted stats file.
func getCgroupFilePath(containerID, base, target, file string) string {
	if len(base) == 0 {
		base = common.HostSys(fmt.Sprintf("fs/cgroup/%s/docker", target))
	}
	statfile := path.Join(base, containerID, file)

	if _, err := os.Stat(statfile); os.IsNotExist(err) {
		statfile = path.Join(
			common.HostSys(fmt.Sprintf("fs/cgroup/%s/system.slice", target)), "docker-"+containerID+".scope", file)
	}

	return statfile
}
```

instead of `statfile := path.Join(base, containerID, file)` perhaps allowing the user to provide just a base path ( and no container id ) would be easier for the user. A work around I am implementing for now is passing an empty string like so:

```go
base := "/sys/fs/cgroup/memory/system.slice/var-lib-docker-containers-"+container.ContainerID+"-mounts-shm.mount"

memory, _ = docker.CgroupMem("", base)
```
as `path.Join` will ignore empty strings, this works ok. But it took a good amount of digging through the source code to figure this out as a potential solution

**Environment (please complete the following information):**
 - [x] Linux: [paste contents of `/etc/os-release` and the result of `uname -a`]

 ```
 [root@ip-XXX ~]# cat /etc/os-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.9 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.9"
[...]

PR #2143, opened 2026-09-05
[docker][linux]: support cgroup v2 in Cgroup* functions

### LINKED ISSUE #1416: cgroup2 support?
**Describe the bug**
I'm trying to get the total memory allocated to my docker container, from inside the container. gopsutil tries to read `/sys/fs/cgroup/memory/`, which does not exist on my system.

**To Reproduce**
```go
    hostname, _ := os.Hostname()

    cgmem, err := docker.CgroupMemDocker(hostname)
    if err != nil {
        fmt.Println(err)
        os.Exit(1)
    }

    fmt.Printf("CgroupMemStat: %+v\n", cgmem)
```

output:
```bash
(ins) ~/erigon-testing/erigon $ docker run --memory 256m -it image /bin/bash
root@71173d45655b:/# /usr/local/bin/erigon
open /sys/fs/cgroup/memory/system.slice/docker-71173d45655b.scope/memory.stat: no such file or directory
```
BUT... when i manually look around the filesystem, I can get the memory max here: 
```bash
root@71173d45655b:/# cat /sys/fs/cgroup/memory.max
268435456
```
It looks like this is a difference between cgroup and cgroup2. cgroup2 has started to get adopted in distros around 2021: https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md. So gopsutil is going to need to support cgroups2 sooner or later

**Expected behavior**
I would expect to see the value 268435456 returned, which represents 256 megabytes of memory allocated to the container.

Host is Debian 11 (linux 5.10.0-20-amd64) x86_64; running docker version 23.0.0; docker guest is ubuntu:20.04.

**Additional context**

hsdfat:
Fixes #1416

**Problem**: every `docker.Cgroup*` function only knows the cgroup v1 layout (`/sys/fs/cgroup/<controller>/docker/<id>` or `.../system.slice/docker-<id>.scope`, reading `cpuacct.stat`, `cpuacct.usage`, `memory.stat`, `memory.usage_in_bytes`, `memory.limit_in_bytes`, `memory.max_usage_in_bytes`, `memory.failcnt`). On a host that mounts the cgroup v2 unified hierarchy (Docker on Debian 11+/Ubuntu 22.04+, every current Kubernetes node) none of those files exist, so every call fails with the error from the issue, and the package's own `TestCgroupCPU`/`TestCgroupMem` fail on such a host.

**Fix**: detect the unified hierarchy through `/sys/fs/cgroup/cgroup.controllers` (the hybrid layout has no such file at the root and keeps the v1 path) and, on v2, resolve the same two directory layouts without the controller segment and read the v2 files into the existing structs:

| cgroup v2 | `CgroupCPUStat` / `CgroupMemStat` |
|---|---|
| `cpu.stat` usage_usec / user_usec / system_usec | `Usage` / `User` / `System` in seconds, same unit as v1 |
| `memory.current` | `MemUsageInBytes` |
| `memory.peak` (Linux 5.19+, 0 when absent) | `MemMaxUsageInBytes` |
| `memory.max` (`max` → `math.MaxUint64`) | `MemLimitInBytes` |
| `memory.events` max | `MemFailCnt` |
| `memory.stat` anon / file / anon_thp / file_mapped / pgfault / pgmajfault / inactive_anon / active_anon / inactive_file / active_file / unevictable | `RSS` / `Cache` / `RSSHuge` / `MappedFile` / `Pgfault` / `Pgmajfault` / same-named fields; the `Total*` fields carry the same values since v2 counters are hierarchical |

`Pgpgin`, `Pgpgout` and `HierarchicalMemoryLimit` have no v2 equivalent and stay 0; the mapping is documented on the reader. No new exported API, no new package, no change to the `cpu`/`mem` host packages, go.mod untouched. The v1 code path is unchanged; `CgroupCPUDocker*`/`CgroupMemDocker*` now pass an empty base so the hierarchy-aware default applies (on v1 that resolves to exactly the path they hard-coded before). The container-ID traversal guard is kept.

**Testing**: fixture trees under `docker/testdata/linux/{cgroup1,cgroup2}` (taken from a Linux 6.8 host, with unmapped keys left in to show they are ignored) driven through `HOST_SYS`, covering both hierarchies with the cgroupfs and systemd driver layouts, the `max` limit, a missing `memory.peak`, an explicit `base`, and invalid container IDs. On a pure cgroup v2 host (Linux 6.8, systemd driver) `CgroupCPUDocker`/`CgroupMemDocker` now return correct values for all 25 running containers, including a real `--memory` limit; on master all 25 fail. `go vet`, cross-GOOS vet and golangci-lint are clean; the file stays `//go:build linux`.

Not covered: the ECS layout from #1061, and a process reading its own cgroup from inside a container with a private cgroup namespace (no `docker/<id>` subdirectory), which the v1 code did not handle either.


shirou (approved):
Thanks for the PR.

I tested on a real cgroup v2 host: Linux 6.6, Docker 29.5.2, systemd cgroup driver, 7 running containers. On master, `TestCgroupCPU` and `TestCgroupMem` fail on this host. With this PR they pass.

The values match the raw cgroup files and `docker stats`:

| field | this PR | cgroup file |
|---|---|---|
| `Usage` | 965.509977 | `cpu.stat` `usage_usec 965510252` |
| `MemUsageInBytes` | 53370880 | `memory.current` 53370880 |
| `MemLimitInBytes` | 268435456 | `memory.max` 268435456 (`docker run --memory 256m`) |
| `RSS` / `Cache` | 18993152 / 11882496 | `memory.stat` `anon` / `file` |

The cgroup v1 path is unchanged. I confirmed that with the same probe on master and on this PR. Your `testdata/linux/cgroup1` tree is also the first cgroup v1 test in this package. Our CI only runs on cgroup v2 hosts, so these fixtures are now our only cgroup v1 coverage.

Thank you for your great contribution!

PR #2148, opened 2026-09-10
Bump perfstat to fix memory leaks

pgimalac:
Bump perfstat to benefit from https://github.com/power-devops/perfstat/commit/88456608a4f64370a1b43fd5d347e0a5f2d1b18a (last commit on main), which fixes many memory leaks in the library.

I was seeing a significant memory leak over time (~1GiB per day when I reproduced it, more than that for some users), tracked it down to `github.com/power-devops/perfstat`, and noticed the last commit fixed a lot of memory leaks.
The leak was particularly visible for logic getting info about filesystem / disks, but I believe I also had smaller leaks from other functions.


shirou (approved):
Thank you!

PR #2149, opened 2026-09-10
[cpu][linux]: expose raw tick counters through ExLinux

### LINKED ISSUE #2147: Add ExLinux to cpu package to expose raw tick counters from /proc/stat

TimesWithContext on Linux reads `/proc/stat` and immediately converts the integer tick counters to float64 seconds by dividing by ClocksPerSec. This means callers that want to compute utilization ratios from integer deltas have no way to do so, the original uint64 values are gone before the function returns.

This matters when computing delta / total for CPU utilization: if the subtraction happens on the already-scaled float64 values, the result varies slightly depending on the absolute counter position even when the tick delta is identical. Computing the ratio from integer deltas (uint64 - uint64) avoids that position-dependent noise.

This would be very helpful for an upstream consumer (open-telemetry/opentelemetry-collector-contrib#47972), as we ended up shipping our own `/proc/stat` parser inside the receiver because there is no raw-ticks API here.

Add cpu/ex_linux.go following the same pattern as mem/ex_linux.go and sensors/ex_linux.go:

```go
  type TimesStatEx struct {
      CPU       string `json:"cpu"`
      User      uint64 `json:"user"`
      // ... same fields as TimesStat but uint64
  }

  type ExLinux struct{}

  func NewExLinux() *ExLinux { return &ExLinux{} }

  func (*ExLinux) TimesWithContext(ctx context.Context, percpu bool) ([]TimesStatEx, error) {
      // reads /proc/stat with strconv.ParseUint, no ClocksPerSec division
  }
```

ClocksPerSec is already exported so consumers can do their own unit conversion when needed.

**Additional context**

See open-telemetry/opentelemetry-collector-contrib#47972

fzlzjerry:
Fixes #2147.

Add the Linux-only `cpu.NewExLinux()` extension with `Times`/`TimesWithContext` and `TimesStatEx`. The counters are parsed directly as `uint64` USER_HZ ticks, retaining adjacent values above the float64 precision boundary and `math.MaxUint64`. The existing floating-point API and its implementation are unchanged.

The new reader uses the existing context/`HOST_PROC` configuration and CPU naming convention. It supports aggregate and per-CPU reads, optional columns, and files without a final newline. It stops at non-CPU metadata without consuming a potentially large interrupt-count line. File and parsing errors are returned rather than silently dropping selected CPUs. No unit conversion, normalization, or utilization calculation is added; the README calls out Linux counter semantics such as decreasing iowait.

### Validation

- CPU tests and race checks pass on Go 1.26.8 and 1.27.1 with the CI-pinned psutil 7.2.2 comparisons enabled.
- Linux/386 CPU tests execute successfully; CPU test binaries also compile for Linux arm, arm64, loong64, riscv64, s390x, and mips, plus Darwin amd64 and Windows amd64/386.
- The new API tests fail to compile against the unmodified base because the API does not exist; they pass with this change. A public-API fixture probe verifies exact integer deltas and unchanged legacy output.
- Repository-pinned golangci-lint 2.13.1 checks pass for the CPU package. Whole-project lint reproduces a pre-existing `reflect.Ptr` govet finding in unchanged `internal/common/common.go`, also present on the base.
- Full-suite comparison in identical isolated Linux containers has the same four failures on base and head: no physical partitions, no utmp, no swap devices, and no temperature sensors. All other tested packages pass. These environment-dependent failures are not reported as a green full-suite run.


shirou (changes requested):
Thank you for your contribution! I agree with the `Ex` approach, and I want this in. The tests are good.

I checked the values against a real `/proc/stat` on a 16-CPU Linux box. For every field and every CPU, `legacy == float64(raw) / ClocksPerSec`.

Could you look at three things before we merge?

### 1. Renaming `TimesStatEx` to `ExTimesStat`

The `NewEx<OS>` API family always puts `Ex` first:

- `mem.ExVirtualMemory`, from `mem.NewExLinux` and `mem.NewExWindows`
- `sensors.ExTemperature`, from `sensors.NewExLinux`

`TimesStatEx` would be the first one in this family with the suffix.

There is one older public type with `Ex` in the middle, `process.MemoryInfoExStat`. That name copies psutil's `memory_info_ex()`, and it is not part of the `NewEx<OS>` family, so I do not think it should decide this.

This is a public API name, so we cannot change it later. The method names (`ExLinux.Times`, `TimesWithContext`) are fine as they are.

### 2. Documenting that `User` includes `Guest`

The kernel adds guest time to both `CPUTIME_GUEST` and `CPUTIME_USER` (or `CPUTIME_NICE`). So `User` already contains `Guest`, and `Nice` already contains `GuestNice`. gopsutil itself subtracts them in `cpu.go`:

```go
func getAllBusy(t TimesStat) (float64, float64) {
	tot := t.Total()
	if runtime.GOOS == "linux" {
		tot -= t.Guest     // Linux 2.6.24+
		tot -= t.GuestNice // Linux 3.2.0+
	}
```

This API exists so callers can compute utilization themselves. If they sum all fields, guest time is counted twice. Please say so in the type doc, and also mention `ClocksPerSec` for unit conversion:

```go
// User includes Guest, and Nice includes GuestNice, because the kernel
// accounts guest time into both. Subtract Guest and GuestNice when summing
// all fields. Divide a counter by ClocksPerSec to get seconds.
```

### 3. Documenting that a delta can underflow

The counters are `uint64`, and the doc already says `Iowait` can go down. When it does, `now.Iowait - before.Iowait` wraps to a huge number instead of going negative. The old `float64` API does not have this problem.

Computing deltas is the whole point of this API, so please warn about it in the type doc:

```go
// A counter can decrease, in particular Iowait. Check that the new value is
// not smaller than the old one before you subtract.
```

The README says "This allows integer deltas to be computed before unit conversion." Could you add the same caution there? Right now the two sentences are several lines apart, and the reader may not connect them.

### suggestion, not blocking: reuse one `/proc/stat` parser

`parseStatLine` and `parseStatLineEx` now read the same file with different code, so they can drift apart. They already disagree: `parseStatLine` needs 7 counters, `parseStatLineEx` needs 4 (see the first nit below).

Routing the old one through the new one works. The old function is 78 lines, the new one is 22, and all existing tests still pass. I tried this locally:

```go
func parseStatLine(line string) (*TimesStat, error) {
	if len(strings.Fields(line)) < 8 {
		return nil, errors.New("stat does not contain cpu info")
	}
	ex, err := parseStatLineEx(line)
	if err != nil {
		return nil, err
	}
	return &TimesStat{
		CPU:       ex.CPU,
		User:      float64(ex.User) / ClocksPerSec,
		Nice:      float64(ex.Nice) / ClocksPerSec,
		System:    float64(ex.System) / ClocksPerSec,
		Idle:      float64(ex.Idle) / ClocksPerSec,
		Iowait:    float64(ex.Iowait) / ClocksPerSec,
		Irq:       float64(ex.Irq) / ClocksPerSec,
		Softirq:   float64(ex.Softirq) / ClocksPerSec,
		Steal:     float64(ex.Steal) / ClocksPerSec,
		Guest:     float64(ex.Guest) / ClocksPerSec,
		GuestNice: float64(ex.GuestNice) / ClocksPerSec,
	}, nil
}
```

Behaviour changes only for malformed lines (`cpu 1.5 ...`, `cpu -1 ...`), which real kernels never produce. Please keep the reader (`bufio` + `Peek`) separate for now, so the old error handling stays as it is.

This is a cleanup, not a bug, so it can also be a follow-up PR. Your call.

### nits

- The doc says "CPU names and selection match the package-level Times function". Not exactly: `parseStatLine` needs 7 counters, `parseStatLineEx` needs 4. A line with 4 to 6 counters is returned by `Ex` but dropped by `Times`. Either say "names match" only, or use the same threshold.
- The README block is a bit long for a README. Splitting `cpu.NewExLinux().TimesWithContext(ctx, percpu)` and `cpu.NewExLinux().Times(percpu)` across lines is hard to read. The `sensor` -> `sensors` fix is correct.

### optional

A `ToTimesStat() TimesStat` method would let callers move between the two types and would keep the two APIs aligned by definition.

rogercoll:
@fzlzjerry Are you planning still planning to address this PR? Thank you for the initial work

fzlzjerry:
@rogercoll yes, sorry for the delay. The review is addressed in 729ee51.

@shirou thanks for checking this on real hardware. Point by point:
- Rename: `TimesStatEx` is now `ExTimesStat` everywhere; the method names are unchanged.
- Guest: the type doc now says `User` includes `Guest` and `Nice` includes `GuestNice` (subtract them when summing all fields), and to divide by `ClocksPerSec` for seconds.
- Underflow: the type doc warns that a counter (notably `Iowait`) can decrease, so check new >= old before subtracting. The README has the same caution right after the integer-delta sentence.
- One parser: done in this PR. `parseStatLine` is now the single uint64 parser; legacy `Times` uses it plus `ToTimesStat()`, and its `ReadLines`/skip-on-error handling is unchanged. `float64(n)` and `ParseFloat` round integers identically, so legacy values stay bit-identical; only malformed lines such as `cpu 1.5 ...` or `cpu -1 ...` are now skipped. Legacy `Times` output is byte-identical to the previous head on a copy of a real `/proc/stat` and on the testdata fixtures.
- Threshold: the shared parser keeps the legacy minimum of 7 counters, so both APIs select the same rows. `Ex` returns an error for a short row where `Times` skips it.
- README: shortened to one paragraph that only mentions `cpu.NewExLinux().Times(percpu)`; the `sensors` fix is kept.
- `ToTimesStat()`: added, and tested for exact equality with the legacy values, including 2^53+1 and MaxUint64.

